PR #1, opened 2025-08-04
Add InteractionsEvaluator for multi-agent and more complicated sequence-based evaluations.

nhungbi:
*Issue #, if available:*

*Description of changes:*
The main change is creating InteractionsEvaluator for evaluating multi-agent systems and lengthy sequence-based evaluations. This evaluator is similar to trajectory evaluations, the only difference is that TrajectoryEvaluator evaluates the whole sequence in one call while InteractionsEvaluator evaluates each interaction one at a time (multiple calls). Other changes include: new additions and refinements of examples, case and dataset are updated to handle interactions, adding helper functions to extract information, and updating tests for the new changes.

Some specific feedback I wanted to get is for InteractionsEvaluator, to evaluate the interactions given expected interactions. I'm passing in a window of relevant interactions (3 max) since the actual interactions can be less or more than the expected interactions so we cannot assume that we can evaluate actual and expected interactions one by one. Eg. we cannot evaluate actual_interactions[i] with expected_interactions[i] since the length might differ. One concern is if the expected interactions are too long, this might cause the evaluation to fail. However, we do need to somehow pass in relevant expected interactions for each interaction evaluated. My current bet is that expected interactions are shorter because they are reference, however, there is no rules that require expected interactions to be short. Let me know what you think.

Additionally, do you think we want to create a TaskOutput class to structure the output type for task or is leaving it as a str or dictionary with "output" and/or "trajectory" and/or "interactions" sufficient? As for helper functions, do you think it's needed to create an Interaction class or is a dictionary with "node_name", "dependencies", and "message" enough?

### DIFF
--- a/src/strands_evaluation/evaluators/interactions_evaluator.py
+++ b/src/strands_evaluation/evaluators/interactions_evaluator.py
@@ -80,7 +80,7 @@ class InteractionsEvaluator(Evaluator[InputT, OutputT]):
         dependencies = interaction.get("dependencies", None)
         messages = interaction.get("messages", None)
         if node_name is None or dependencies is None or messages is None:
-            raise Exception("Please make sure the task function return a dictionary with the key 'interactions' that contains 'node_name', 'dependencies', and 'messages'.")
+            raise Exception("Please make sure the task function returns a dictionary with the key 'interactions' that contains 'node_name', 'dependencies', and 'messages'.")
         
         # Create Prompt
         if is_last:
@@ -130,7 +130,7 @@ class InteractionsEvaluator(Evaluator[InputT, OutputT]):
             The results of the evaluation as EvaluationOutput.
         """
         if not evaluation_case.actual_interactions:
-            raise Exception("Please make sure the task function return a dictionary with the key 'interactions'.")
+            raise Exception("Please make sure the task function returns a dictionary with the key 'interactions'.")
         num_interactions = len(evaluation_case.actual_interactions)
 
         # keep all of the context
@@ -148,11 +148,8 @@ class InteractionsEvaluator(Evaluator[InputT, OutputT]):
             evaluation_prompt = self._compose_prompt(evaluation_case, i, is_last)
 
             ## Evaluate ##
-            if is_last:
-                result = evaluator_agent.structured_output(EvaluationOutput, evaluation_prompt)
-            else:
-                evaluator_agent.structured_output(EvaluationOutput, evaluation_prompt)
-
+            result = evaluator_agent.structured_output(EvaluationOutput, evaluation_prompt)
+            
         return result
 
     async def evaluate_async(self, evaluation_case: EvaluationData[InputT, OutputT]) -> EvaluationOutput:
@@ -166,7 +163,7 @@ class InteractionsEvaluator(Evaluator[InputT, OutputT]):
             The results of the evaluation as EvaluationOutput.
         """
         if not evaluation_case.actual_interactions:
-            raise Exception("Please make sure the task function return a dictionary with the key 'interactions'.")
+            raise Exception("Please make sure the task function returns a dictionary with the key 'interactions'.")
         num_interactions = len(evaluation_case.actual_interactions)
 
         # keep all of the context
@@ -185,10 +182,7 @@ class InteractionsEvaluator(Evaluator[InputT, OutputT]):
             evaluation_prompt = self._compose_prompt(evaluation_case, i, is_last)
 
             ## Evaluate ##
-            if is_last:
-                result = await evaluator_agent.structured_output_async(EvaluationOutput, evaluation_prompt)
-            else:
-                await evaluator_agent.structured_output_async(EvaluationOutput, evaluation_prompt)
+            result = await evaluator_agent.structured_output_async(EvaluationOutput, evaluation_prompt)
 
         return result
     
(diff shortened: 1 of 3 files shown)


src/strands_evaluation/evaluators/interactions_evaluator.py
```diff
@@ -0,0 +1,196 @@
+from typing_extensions import TypeVar
+
+from .evaluator import Evaluator
+from ..types.evaluation import EvaluationData, EvaluationOutput
+from strands import Agent
```
poshinchen, line 5: These ordering will be a follow-up fix with linter and formatter.

src/strands_evaluation/evaluators/utils/helper_funcs.py
```diff
@@ -0,0 +1,180 @@
+from strands.multiagent import SwarmResult
+from strands.multiagent import GraphResult
+from strands import Agent
+
+### Extracting Information from SWARM ###
```
poshinchen: this is not needed

src/strands_evaluation/evaluators/utils/helper_funcs.py
```diff
@@ -0,0 +1,180 @@
...
+        swarm_result: Result object from swarm execution
+    
+    Returns:
+        list: Interactions with node names, messages, and dependencies
+        [{node_name: str, messages: list[str], dependencies: list[str]}, ...]
+    """
+    handoff_info = extract_swarm_handoffs(swarm_result)
+    return extract_swarm_interactions_from_handoffs(handoff_info)
+
+### Extracting Information from GRAPH ###
```
poshinchen: this is not needed

src/strands_evaluation/evaluators/utils/helper_funcs.py
```diff
@@ -0,0 +1,180 @@
...
+        node_messages = [m["text"] for m in node.result.result.message["content"]]
+        dependencies = [n.node_id for n in node.dependencies]
+        message_info.append({
+            "node_name": node_name,
+            "dependencies": dependencies,
+            "messages": node_messages
+        })
+    return message_info
+
+### Extract Information from Agent result ###
```
poshinchen: this is not needed

src/strands_evaluation/evaluators/utils/helper_funcs.py
```diff
@@ -0,0 +1,180 @@
...
+        tools_used.append({
+            "name": tool_name,
+            "input": tool_info.tool.get("input"),
+            "call_count": tool_info.call_count,
+            "success_count": tool_info.success_count,
+            "total_time": tool_info.total_time,
+        })
+    return tools_used
+
+### Extract Information from Agent ###
```
poshinchen: this is not needed

src/strands_evaluation/evaluators/interactions_evaluator.py
```diff
@@ -0,0 +1,196 @@
...
+        Evaluate the performance of the task on the given test cases.
+
+        Args:
+            evaluation_case: The test case with all of the neccessary context to be evaluated.
+        
+        Returns:
+            The results of the evaluation as EvaluationOutput.
+        """
+        if not evaluation_case.actual_interactions:
+            raise Exception("Please make sure the task function return a dictionary with the key 'interactions'.")
```
poshinchen: nit: returns

src/strands_evaluation/evaluators/interactions_evaluator.py
```diff
@@ -0,0 +1,196 @@
...
+        for i in range(num_interactions): # evaluate one interaction at a time
+            if i == num_interactions - 1:
+                is_last = True
+            evaluation_prompt = self._compose_prompt(evaluation_case, i, is_last)
+
+            ## Evaluate ##
+            if is_last:
+                result = evaluator_agent.structured_output(EvaluationOutput, evaluation_prompt)
+            else:
+                evaluator_agent.structured_output(EvaluationOutput, evaluation_prompt)
```
poshinchen: Using `result = evaluator_agent.structured_output(EvaluationOutput, evaluation_prompt)` should be fine right?

Only the last one will be returned

src/strands_evaluation/evaluators/utils/helper_funcs.py
```diff
@@ -0,0 +1,180 @@
...
+        added = False
+        for tool_name, tool_info in node_info.result.metrics.tool_metrics.items():
+            if tool_name == "handoff_to_agent":
+                hand_off_info.append({
+                    "from": node_name,
+                    "to": tool_info.tool["input"]["agent_name"],
+                    "messages": messages
+                })
+                added = True
+        if not added:
```
poshinchen, line 28: you can just check if `hand_off_info` is empty. Then `added` is not needed

src/strands_evaluation/evaluators/utils/helper_funcs.py
nhungbi: Just to confirm, this is the comment at #5 right? not the helper function?

src/strands_evaluation/evaluators/utils/helper_funcs.py
nhungbi: That won't work because I'm checking for each node, the boolean is for each node, not each hand_off. So, hand_off_info may have elements from previous nodes but the current node message may not have added yet if it is not handing off to any agent (eg. the last sub agent).

src/strands_evaluation/evaluators/utils/helper_funcs.py
poshinchen: yes, since you have the description after the method, there's no need for `###` :)

PR #4, opened 2025-08-08
Add support for dataset generation with strands agents

nhungbi:
*Issue #, if available:*

*Description of changes:*
Adding support for generating Dataset from nothing, from a reference dataset, and adding to a dataset. Currently, the DatasetGenerator only support OutputEvaluator, TrajectoryEvaluator, and InteractionEvaluator for rubric generation. For InteractionEvaluator, the generator can only generate one rubric for all of the interactions. Generating separate rubric for each node/interaction is not supported yet. Other changes include: adding an Interaction TypedDict, updating the Dataset class with more methods, and add some minor fixes for a cleaner interface. 

Through testing, it looks like generating multiple test cases at a time is not very deterministic in quality and quantity, so only 1 case is generated per call right now. With structured_output, creating a new class (_Case) with only fields that we want in a test case is better so that the agent doesn't generate rubbish information in other optional fields.

Future changes include: filtering and cleaning /examples.

### DIFF
--- a/src/strands_evaluation/generators/dataset_generator.py
+++ b/src/strands_evaluation/generators/dataset_generator.py
@@ -40,7 +40,7 @@ class DatasetGenerator(Generic[InputT, OutputT]):
         include_expected_output: bool = True,
         include_expected_trajectory: bool = False,
         include_expected_interactions: bool = False,
-        include_meta_data: bool = False,
+        include_metadata: bool = False,
         model: str | None = None,
         max_parallel_num_cases: int = 10,
         rubric_system_prompt: str = RUBRIC_SYSTEM_PROMPT,
@@ -56,7 +56,7 @@ class DatasetGenerator(Generic[InputT, OutputT]):
             include_expected_output: Whether to include expected outputs in test cases
             include_expected_trajectory: Whether to include expected tool/action trajectories
             include_expected_interactions: Whether to include expected interaction sequences
-            include_meta_data: Whether to include metadata fields in test cases
+            include_metadata: Whether to include metadata fields in test cases
             model: Model identifier for the generation agent, defaults to strands' default model.
             max_parallel_num_cases: Maximum number of test cases to generate in parallel asynchronously
             rubric_system_prompt: System prompt for rubric generation, defaults to one of the available templates.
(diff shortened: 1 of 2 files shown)


src/strands_evaluation/dataset.py
```diff
@@ -18,7 +17,7 @@
 OutputT = TypeVar("OutputT")
 
 
-class Dataset(BaseModel, Generic[InputT, OutputT]):
+class Dataset(Generic[InputT, OutputT]):
```
poshinchen, line 20: So you don't see any issue after removing BaseModel?

src/strands_evaluation/dataset.py
```diff
@@ -47,8 +46,52 @@ class Dataset(BaseModel, Generic[InputT, OutputT]):
         )
     """
 
-    cases: list[Case[InputT, OutputT]]
-    evaluator: Evaluator[InputT, OutputT]
+    def __init__(self, cases: list[Case[InputT, OutputT]] = None, evaluator: Evaluator[InputT, OutputT] = Evaluator()):
```
poshinchen, line 49: does it make sense to be list[Case[InputT, OutputT]] | None = None
and evaluator: Evaluator[InputT, OutputT] | None and set the Evaluator() at line 51?

src/strands_evaluation/dataset.py
nhungbi: BaseModel allow us to have a nicer print of Dataset. eg. now when you print it, you get: <strands_evaluation.dataset.Dataset object at 0x103a34320>. But it's not a problem since there's a to_dict() method that's probably better anyways since removing the basemodel here allows us to remove @dataclass in the Evaluator class.

src/strands_evaluation/dataset.py
nhungbi: I don't have strong feelings either way, so I'll fix it to be like this. I'm not sure if there's any benefit in one way.

src/strands_evaluation/evaluators/utils/case_prompt_template.py
```diff
@@ -38,7 +38,7 @@ def compose_test_prompt(
         if evaluation_case.actual_output:
             evaluation_prompt += f"<Output>{evaluation_case.actual_output}</Output>\n"
     else:
-        if not evaluation_case.actual_output:
+        if evaluation_case.actual_output is None:
```
poshinchen, line 41: nit: what's the intention of changing it to None instead of not?

nhungbi: None is the default so we never want to include it, but if the user for some reason want an empty list or array, 'not' would not include it.

src/strands_evaluation/dataset.py
nhungbi: I will do this fix in the refactor PR.

PR #15, opened 2025-10-13
fix(workflow): fix workflow failures and type checks

poshinchen:
## Description
Fixed most of the workflow failures (e.g. type checks).

1. Skipped `src/examples` at this moment as they would be moved to another directory
2. Skipped the line-length check `E501` for prompt templates as they are docstring templates.
3. Commented out `hatch test tests_integ` first as there's no integ tests running as of now. **Although the changes can only be reflected after merging the pr**

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -63,11 +63,3 @@ jobs:
       - name: Install dependencies
         run: |
           pip install --no-cache-dir hatch
-      - name: Run integration tests
-        env:
-          AWS_REGION: us-east-1
-          AWS_REGION_NAME: us-east-1 # Needed for LiteLLM
-          STRANDS_TEST_API_KEYS_SECRET_NAME: ${{ secrets.STRANDS_TEST_API_KEYS_SECRET_NAME }}
-        id: tests
-        run: |
-          hatch test tests_integ
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,9 +26,10 @@ packages = ["src/strands_evals"]
 
 [project.optional-dependencies]
 test = [
-    "pytest>=7.0",
-    "pytest-asyncio>=0.26.0",
-    "pytest-cov>=4.0",
+    "pytest>=8.0.0,<9.0.0",
+    "pytest-cov>=7.0.0,<8.0.0",
+    "pytest-asyncio>=1.0.0,<1.3.0",
+    "pytest-xdist>=3.0.0,<4.0.0",
 ]
 
 dev = [
@@ -41,6 +42,17 @@ dev = [
 [tool.ruff]
 line-length = 120
 include = ["src/**/*.py", "tests/**/*.py"]
+
+[tool.hatch.envs.hatch-test]
+installer = "uv"
+extra-args = ["-n", "auto", "-vv"]
+dependencies = [
+    "pytest>=8.0.0,<9.0.0",
+    "pytest-cov>=7.0.0,<8.0.0",
+    "pytest-asyncio>=1.0.0,<1.3.0",
+    "pytest-xdist>=3.0.0,<4.0.0",
+    "moto>=5.1.0,<6.0.0",
+]
 
 [tool.hatch.envs.default.scripts]
 list = [
@@ -58,6 +70,7 @@ lint = [
 test-lint = [
     "hatch fmt --linter --check"
 ]
+
 prepare = [
     "hatch fmt --linter",
     "hatch fmt --formatter",
@@ -87,7 +100,26 @@ select = [
   "F", # pyflakes
   "I", # isort
   "B", # flake8-bugbear
+  "T20", # flake8-print (disallow print statements)
 ]
+
+[tool.ruff.lint.per-file-ignores]
+"src/strands_evals/evaluators/prompt_templates/*" = ["E501"] # line-length
+"src/strands_evals/generators/prompt_template/*" = ["E501"] # line-length
+"src/examples/*" = ["E501", "T201"] # line-length, print
+
+[tool.mypy]
+exclude = [
+    "src/examples/",
+]
+# Disable strict checks that cause false positives with Generic classes
+disable_error_code = [
+    "no-redef",  # Allows property setters without "already defined" errors
+    "attr-defined",  # Allows property.setter pattern in Generic classes
+    "import-untyped",  # Allows imports from modules without type stubs
+]
+# Allow untyped decorators (helps with @property in Generic classes)
+disallow_untyped_decorators = false
 
 [tool.hatch.version]
 path = "src/strands_evals/__init__.py"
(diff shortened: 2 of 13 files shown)


pyproject.toml
```diff
@@ -41,6 +42,22 @@ dev = [
 [tool.ruff]
 line-length = 120
 include = ["src/**/*.py", "tests/**/*.py"]
+exclude = [
+    "src/strands_evals/evaluators/prompt_templates/*",
+    "src/strands_evals/generators/prompt_template/*",
+    "src/examples/*",
+]
```
pgrayy: Is there a need to exclude these directories?

pyproject.toml
```diff
@@ -89,6 +106,19 @@ select = [
   "B", # flake8-bugbear
 ]
 
+[tool.mypy]
+exclude = [
+    "src/examples/",
+]
```
pgrayy, line 114: Same here. Do we need to exclude the examples?

src/strands_evals/dataset.py
```diff
@@ -42,12 +42,15 @@ class Dataset(Generic[InputT, OutputT]):
                         expected_trajectory=["calculator],
                         metadata={"category": "math"})
             ],
-            evaluator=OutputEvaluator(rubric = "The output is relevant and complete. 0 if the output is incorrect or irrelevant.")
+            evaluator=OutputEvaluator(rubric = "The output is relevant and complete. 0 if the output
+                is incorrect or irrelevant.")
```
pgrayy: It would be more conventional to write this line as:
```Python
evaluator=OutputEvaluator(rubric="The output is relevant and complete. 0 if the output is incorrect or irrelevant.")
```
It is only 116 characters in length (120 should be the max).

src/strands_evals/extractors/swarm_extractor.py
```diff
@@ -14,9 +14,15 @@ def extract_swarm_handoffs(swarm_result: SwarmResult) -> list[dict]:
...
+        print("==========================================", node_name, node_info)
+        print(node_name, node_info)
+        print("==========================================")
+        messages = [m["text"] for m in node_info.result.message["content"]]  # type: ignore[union-attr]
+        print("==========================================messages")
+        print(messages)
+        print("==========================================")
         added = False
-        for tool_name, tool_info in node_info.result.metrics.tool_metrics.items():
+        for tool_name, tool_info in node_info.result.metrics.tool_metrics.items():  # type: ignore[union-attr]
```
pgrayy: Nit: I would recommend trying to avoid the `# type` ignore primitive. If there is a utility in `typing` that can be used to address the error, we should use it.

src/strands_evals/generators/dataset_generator.py
```diff
@@ -76,18 +80,19 @@ def __init__(
...
         if self.include_expected_trajectory:
-            fields["expected_trajectory"] = (list[trajectory_type], ...) if trajectory_type else (list[Any], ...)
+            # Use Any for trajectory type since we can't use runtime variables as types
+            fields["expected_trajectory"] = (list[Any], ...)
         if self.include_expected_interactions:
             fields["expected_interactions"] = (list[Interaction], ...)
         if self.include_metadata:
             fields["metadata"] = (dict[str, Any], ...)
-        self._Case = create_model("_Case", **fields)
+        self._Case = create_model("_Case", **fields)  # type: ignore[call-overload]
```
pgrayy: We should try to fix the typing overload rather than ignore it. This can actually help us statically detect bugs.

src/strands_evals/dataset.py
poshinchen: it'll become 128 including the tab indices

pyproject.toml
poshinchen: I think we'll move the examples away to another directory. Therefore I don't want to fix the types and linting at this point.

pyproject.toml
poshinchen: discussed offline, will only skipped line-length check.

.github/workflows/integration-test.yml
```diff
@@ -63,11 +63,11 @@ jobs:
...
-        run: |
-          hatch test tests_integ
\ No newline at end of file
+      # - name: Run integration tests
+      #   env:
+      #     AWS_REGION: us-east-1
+      #     AWS_REGION_NAME: us-east-1
+      #     STRANDS_TEST_API_KEYS_SECRET_NAME: ${{ secrets.STRANDS_TEST_API_KEYS_SECRET_NAME }}
+      #   id: tests
+      #   run: |
```
Unshure: Should we just remove these?

src/strands_evals/evaluators/interactions_evaluator.py
```diff
@@ -86,29 +86,45 @@ def _compose_prompt(
...
+                "Evaluate this interaction. THE SCORE MUST BE A DECIMAL BETWEEN 0.0 AND 1.0 "
+                "(NOT 0 to 10 OR 0 to 100). \n"
+            )
 
         if self.include_inputs:
-            if isinstance(evaluation_case.input, list) and len(evaluation_case.input) == len(
-                evaluation_case.actual_interactions
+            if (
+                isinstance(evaluation_case.input, list)
+                and isinstance(evaluation_case.actual_interactions, list)
```
Unshure, line 103: nit: You check if this is a list, then on line 112 you check if it is none. This seems redundant

src/strands_evals/evaluators/trajectory_evaluator.py
```diff
@@ -38,7 +40,11 @@ def __init__(
         self.trajectory_description = trajectory_description
         self.model = model
         self.include_inputs = include_inputs
-        self._tools = [exact_match_scorer, in_order_match_scorer, any_order_match_scorer]
+        self._tools: Optional[list[Union[str, dict[str, str], Any]]] = [
```
Unshure: nit: Try to avoid using the `typing` import and use pythons built in typing support

.github/workflows/integration-test.yml
poshinchen: we can, just commented it out because we might need it in the future.

src/strands_evals/dataset.py
```diff
@@ -42,12 +42,15 @@ class Dataset(Generic[InputT, OutputT]):
                         expected_trajectory=["calculator],
                         metadata={"category": "math"})
             ],
-            evaluator=OutputEvaluator(rubric = "The output is relevant and complete. 0 if the output is incorrect or irrelevant.")
+            evaluator=OutputEvaluator(rubric="The output is relevant and complete. 0 if the output is
+                incorrect or irrelevant.")
```
pgrayy, line 46: Nit: I would recommend wrapping as:
```Python
evaluator=OutputEvaluator(
    rubric="The output is relevant and complete. 0 if the output is incorrect or irrelevant."
)

# or

evaluator=OutputEvaluator(
    rubric=(
        "The output is relevant and complete. "
        "0 if the output is incorrect or irrelevant."
    )
)
```

PR #18, opened 2025-10-16
Add multi-step dataset generation with topic planning.

jjbuck:
## Description

Add optional topic planning to dataset generation for improved test case diversity. When num_topics is specified, the system uses a two-phase process: first planning diverse topics, then distributing test cases across them.

Key changes
- Add TopicPlanner class with Topic and TopicPlan Pydantic models
- Add num_topics parameter to generate_cases_async() and from_context_async()
- Refactor generation into _prepare_generation_prompts() and _generate_batch()
- Implement prompt expansion strategy: topics → multiple prompts → unified generation

Implementation approach
- Treat topic planning as prompt expansion, not a separate generation mode
- Reuse existing _case_worker pattern for all generation (with or without topics)
- Same code path handles num_topics=None (single prompt) and num_topics=N (multiple prompts)
- The current behavior of the dataset generator is maintained when num_topics=None is used.
- Exact case distribution with remainder handling ensures precise num_cases

Benefits
- Improved test case diversity through strategic topic planning
- Seamless integration with all existing generation methods
- No breaking changes - num_topics=None preserves original behavior
- Minimal code addition (~100 lines) by leveraging existing primitives

### Example usage:
**Original behavior (fast, single LLM call)**
  dataset = await generator.from_context_async(
      context="Available tools: book_flight, cancel_booking",
      task_description="Travel booking assistant",
      num_cases=30,
      evaluator=OutputEvaluator
  )

**Multi-step generation (slower, more diverse)**

  dataset = await generator.from_context_async(
      context="Available tools: book_flight, cancel_booking",
      task_description="Travel booking assistant",
      num_cases=30,
      num_topics=6,  # Plans 6 topics, ~5 cases per topic
      evaluator=OutputEvaluator
  )

## Related Issues

N/A

## Documentation PR

N/A 

## Type of Change

New feature

## Testing

## Checklist

### DIFF
--- a/src/strands_evals/generators/dataset_generator.py
+++ b/src/strands_evals/generators/dataset_generator.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import math
+from textwrap import dedent
 
 from pydantic import create_model
 from strands import Agent
@@ -158,7 +159,7 @@ class DatasetGenerator(Generic[InputT, OutputT]):
             )
             generated_cases.extend(cases)
 
-        return generated_cases[:num_cases]
+        return generated_cases
 
     async def _prepare_generation_prompts(
         self, base_prompt: str, num_cases: int, num_topics: int | None = None
@@ -186,18 +187,22 @@ class DatasetGenerator(Generic[InputT, OutputT]):
         cases_per_topic = math.ceil(num_cases / len(topic_plan.topics))
         prompt_specs: list[tuple[str, int]] = []
 
+        num_generated_cases = 0
         for topic in topic_plan.topics:
-            remaining = num_cases - sum(count for _, count in prompt_specs)
+            remaining = num_cases - num_generated_cases
             if remaining <= 0:
                 break
 
             topic_cases = min(cases_per_topic, remaining)
-            topic_prompt = f"""{base_prompt}
-            Focus on this topic:
-            - {topic.title}: {topic.description}
-            - Key aspects: {', '.join(topic.key_aspects)}"""
+            topic_prompt = dedent(f"""
+                {base_prompt}
+                Focus on this topic:
+                - {topic.title}: {topic.description}
+                - Key aspects: {", ".join(topic.key_aspects)}
+            """)
 
             prompt_specs.append((topic_prompt, topic_cases))
+            num_generated_cases += topic_cases
 
         return prompt_specs
 
--- a/src/strands_evals/generators/topic_planner.py
+++ b/src/strands_evals/generators/topic_planner.py
@@ -1,4 +1,5 @@
 import math
+from textwrap import dedent
 
 from pydantic import BaseModel, Field
 from strands import Agent
@@ -35,20 +36,21 @@ class TopicPlanner:
 
         planning_agent = Agent(model=self.model, system_prompt=self.planning_prompt, callback_handler=None)
 
-        prompt = f"""Generate {num_topics} diverse topics for creating {num_cases} test cases.
+        prompt = dedent(f"""
+                        Generate {num_topics} diverse topics for creating {num_cases} test cases.
 
-Agent Context:
-{context}
+                        Agent Context:
+                        {context}
 
-Task Description:
-{task_description}
+                        Task Description:
+                        {task_description}
 
-Requirements:
-- Create exactly {num_topics} distinct topics
-- Each topic will generate approximately {cases_per_topic} test cases
-- Include 2-5 key aspects per topic that test cases should explore
-- Ensure topics span different complexity levels and use cases
-- Make topics diverse and non-overlapping"""
+                        Requirements:
+                        - Create exactly {num_topics} distinct topics
+                        - Each topic will generate approximately {cases_per_topic} test cases
+                        - Include 2-5 key aspects per topic that test cases should explore
+                        - Ensure topics span different complexity levels and use cases
+                        - Make topics diverse and non-overlapping""")
 
         topic_plan = await planning_agent.structured_output_async(TopicPlan, prompt)
 
--- a/src/examples/dataset_generator/topic_planning_dataset.py
+++ b/src/examples/dataset_generator/topic_planning_dataset.py
@@ -1,6 +1,7 @@
 import asyncio
 
 from strands import Agent
+
 from strands_evals.evaluators.output_evaluator import OutputEvaluator
 from strands_evals.generators.dataset_generator import DatasetGenerator
 
--- a/tests/strands_evals/generators/test_topic_planner.py
+++ b/tests/strands_evals/generators/test_topic_planner.py
@@ -1,6 +1,7 @@
 from unittest.mock import AsyncMock, patch
 
 import pytest
+
 from strands_evals.generators.topic_planner import Topic, TopicPlan, TopicPlanner
 
 


src/strands_evals/generators/dataset_generator.py
```diff
@@ -1,14 +1,16 @@
 import asyncio
 import logging
+import math
 
 from pydantic import create_model
 from strands import Agent
-from typing_extensions import Any, Generic, TypeVar
+from typing_extensions import Any, Generic, TypeVar, Optional
```
poshinchen: let's use `| None` instead

src/strands_evals/generators/dataset_generator.py
```diff
@@ -128,24 +130,99 @@ async def _case_worker(self, queue: asyncio.Queue, prompt: str, message_history:
...
+                break
+                
+            topic_cases = min(cases_per_topic, remaining)
+            topic_prompt = f"""{base_prompt}
+
+    Focus on this topic:
+    - {topic.title}: {topic.description}
+    - Key aspects: {', '.join(topic.key_aspects)}"""
+            
+            prompt_specs.append((topic_prompt, topic_cases))
```
poshinchen: These look a bit of off the format.

src/strands_evals/generators/dataset_generator.py
jjbuck: Fixed.

src/strands_evals/generators/dataset_generator.py
jjbuck: Fixed.

src/strands_evals/generators/dataset_generator.py
```diff
@@ -138,14 +140,73 @@ async def generate_cases_async(
...
+                break
+
+            topic_cases = min(cases_per_topic, remaining)
+            topic_prompt = f"""{base_prompt}
+
+Focus on this topic:
+- {topic.title}: {topic.description}
+- Key aspects: {', '.join(topic.key_aspects)}"""
+
+            prompt_specs.append((topic_prompt, topic_cases))
```
poshinchen: This still looks a bit off the format?

src/strands_evals/generators/topic_planner.py
```diff
@@ -0,0 +1,68 @@
+import math
+from typing import List
```
poshinchen: same as above, prefer to use `list`

jjbuck: Ack - scrubbed all uses of `List`, `Dict`, and `Optional`.

src/strands_evals/generators/dataset_generator.py
poshinchen: Wondering what would be the diff if we make it:

```
topic_prompt = f"""{base_prompt}
Focus on this topic:
- {topic.title}: {topic.description}
- Key aspects: {', '.join(topic.key_aspects)}"""
```

With textwarp.dedent the leading whitespaces should not be an issue.

src/strands_evals/generators/topic_planner.py
```diff
@@ -0,0 +1,67 @@
...
+    DEFAULT_PLANNING_PROMPT = """You are a test scenario planner for AI agents. 
+Your role is to analyze agent configurations and generate strategic topic plans 
+that comprehensively evaluate agent capabilities.
+
+Your topics should:
+- Cover different aspects of the agent's capabilities
+- Test edge cases and common scenarios
+- Vary in complexity and scope
+- Ensure comprehensive coverage of available tools and features
+- Be diverse and non-overlapping"""
```
poshinchen: Does it make sense to move the prompt into ./prompt_templates?

src/strands_evals/generators/dataset_generator.py
```diff
@@ -138,14 +140,73 @@ async def generate_cases_async(
...
+            List of (prompt, num_cases) tuples. Always returns at least one prompt.
+        """
+        if num_topics is None:
+            return [(base_prompt, num_cases)]
+
+        topic_planner = TopicPlanner(model=self.model)
+
+        try:
+            topic_plan = await topic_planner.plan_topics_async(
+                context=base_prompt, task_description="", num_topics=num_topics, num_cases=num_cases
```
poshinchen, line 180: why is `task_description` always empty?

src/strands_evals/generators/dataset_generator.py
```diff
@@ -138,14 +140,73 @@ async def generate_cases_async(
...
+        except Exception as e:
+            logger.warning(f"Topic planning failed: {e}. Using single prompt.")
+            return [(base_prompt, num_cases)]
+
+        # Distribute cases across topics
+        cases_per_topic = math.ceil(num_cases / len(topic_plan.topics))
+        prompt_specs: list[tuple[str, int]] = []
+
+        for topic in topic_plan.topics:
+            remaining = num_cases - sum(count for _, count in prompt_specs)
```
poshinchen: This makes O(N*2). 

This could be cleaner and faster:
```
num_generated_cases = 0

for topic in topic_plan.topics:
    remaining = num_cases - num_generated_cases
    if remaining <= 0:
        break

    topic_cases = min(cases_per_topic, remaining)
    topic_prompt = <prompt>

    prompt_specs.append((topic_prompt, topic_cases))
    num_generated_cases += topic_cases

return prompt_specs
```

src/strands_evals/generators/dataset_generator.py
jjbuck: **TLDR This is a CX question stemming from some asymmetry between the `from_context` and `from_cases` interfaces. Take a look at the code snippets below and let me know what you think is most aligned with the Strands Evals CX.**

The current implementation supports both direct TopicPlanner usage with separated `context`/`task_description` fields for optimal prompt structure, and internal usage from `from_context_async` where both pieces of information are consolidated into the `context` parameter (with `task_description=""` since it's already embedded in the combined prompt).

Ultimately this has to do with some of the refactoring and adaption work we've done to incorporate the topic planner, and the inherent asymmetry between `from_context_async` and `generate_cases_async`. When the entrypoint is `from_context_async`, it essentially "bakes the cake early" and creates a prompt that has the task description already baked in alongside the context, which is then passed as an argument to `generate_cases_async`. 
```
from_context_async(context="tools...", task_description="booking assistant")
  │
  ├─ Creates: "Create test cases with context: tools... for this task: booking assistant"
  │  (COMBINED_PROMPT - task_description is baked in)
  │
  └─> generate_cases_async(prompt=COMBINED_PROMPT)
       └─> plan_topics_async(context=COMBINED_PROMPT, task_description="")
```
This gives us a few options:

### Option 1 is to thread parameters through from `generate_cases_async` (less ideal IMO)

Keep context/task_description separated throughout the call chain, e.g.,
```
async def generate_cases_async(
    self, 
    prompt: str,
    num_cases: int = 5,
    message_history: list | None = None,
    num_topics: int | None = None,
    context: str | None = None,  # NEW - for topic planning
    task_description: str | None = None,  # NEW - for topic planning
) -> list[Case]:
    prompt_specs = await self._prepare_generation_prompts(
        base_prompt=prompt,
        num_cases=num_cases,
        num_topics=num_topics,
        context=context,
        task_description=task_description,
    )
```
Then from_context_async passes both the combined prompt **and** the originals.
```
cases = await self.generate_cases_async(
    prompt=combined_prompt,  # For actual generation
    num_cases=num_cases,
    num_topics=num_topics,
    context=context,  # For topic planning
    task_description=task_description,  # For topic planning
)
```

This is much more invasive, and likely causes the `generate_cases_async` function signature to deviate from its design intent. 

### Option 2 is to simplify the Topic Planner interface

This would simplify the TopicPlanner interface and system prompt to more closely correspond to that of `from_cases_async`. There's no reason this *wouldn't* work from a *technical* perspective, but it's arguably less clear of an interface in that it doesn't make it clear to users that the task description is supposed to be embedded in the `prompt`. In other words, this makes the `plan_topics` interface *less* similar to that of `from_context` and *more* similar to that of `generate_cases`.
```
async def plan_topics_async(
    self, 
    prompt: str,  # Just one unified prompt
    num_topics: int, 
    num_cases: int
) -> TopicPlan:
    planning_prompt = f"""Generate {num_topics} diverse topics.

Context and Task:
{prompt}

Requirements: ..."""
```

### Option 3 is to maintain an explicit `task_description` argument but document the flexibility
```
# Usage for direct invocation

# User has separated values
topic_plan = await planner.plan_topics_async(
    context="tools: book_flight...",
    task_description="booking assistant",
    num_topics=6,
    num_cases=30
)

# Internal usage (current)
# from_context_async has already combined them
topic_plan = await planner.plan_topics_async(
    context=combined_prompt,  # Context includes task_description
    task_description="",  # Implicit in context
    num_topics=6,
    num_cases=30
)
```

src/strands_evals/generators/dataset_generator.py
jjbuck: Addressed this by setting all the text with a common indentation level for legibility, and then using `dedent()` to avoid spurious whitespace in the text content.

src/strands_evals/generators/dataset_generator.py
jjbuck: (For posterity) We connected separately, did a code walkthrough, and decided to leave the current interface as-is, with the possibility to revisit prior to release.

src/strands_evals/generators/dataset_generator.py
jjbuck: Good callout. I also had to clean up something in the calling function above where it was doing extra work when `num_cases` / `num_topics` had a non-zero remainder.

src/strands_evals/generators/topic_planner.py
jjbuck: Yep, makes sense to me. I made this change in the latest update.

src/strands_evals/generators/topic_planner.py
```diff
@@ -0,0 +1,58 @@
...
+
+Task Description:
+{task_description}
+
+Requirements:
+- Create exactly {num_topics} distinct topics
+- Each topic will generate approximately {cases_per_topic} test cases
+- Include 2-5 key aspects per topic that test cases should explore
+- Ensure topics span different complexity levels and use cases
+- Make topics diverse and non-overlapping"""
```
poshinchen: can you use dedent here too? the rest of them look good to me!

jjbuck: Addressed in the latest edit.

PR #16, opened 2025-10-14
Implement trace-based evaluation with helpfulness evaluator.

jjbuck:
## Description

Add trace parsing infrastructure and HelpfulnessEvaluator as first trace-based
evaluator, enabling turn-level and tool-level evaluation of agent interactions.

Key additions:
- Session/Trace/Span observability primitives for structured trace data
- Base Evaluator extended with _parse_turn_level() and _parse_tool_level()
- StrandsInMemoryMapper for converting Strands agent traces to eval format
- HelpfulnessEvaluator with 7-point categorical scale (normalized to 0-1)
- Versioned prompt template system (helpfulness_v0.py)
- EvaluationLevel enum (TURN_LEVEL, TOOL_LEVEL, CONVERSATION_LEVEL)
- TurnLevelInput and ToolLevelInput structured types for parsed trace data
- Comprehensive test coverage for parsing, conversion, and evaluation logic
- Example usage demonstrating trace-based evaluation workflow

This establishes the foundation for enabling sophisticated trace-based evaluation
in Strands Eval.

## Related Issues

N/A

## Documentation PR

N/A 

## Type of Change

New feature

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

PR #22, opened 2025-10-29
feat: add GoalSuccessRateEvaluator for conversation goal tracking

jjbuck:
## Description
Add conversation-level evaluator that assesses whether all user goals
were successfully achieved using binary Yes/No scoring with structured
reasoning output.

## Related Issues

1. Related to #16 .

## Documentation PR

N/A

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/evaluators/evaluator.py
+++ b/src/strands_evals/evaluators/evaluator.py
@@ -8,8 +8,10 @@ from ..types.trace import (
     AgentInvocationSpan,
     AssistantMessage,
     Conversation,
+    ConversationLevelInput,
     EvaluationLevel,
     Session,
+    SpanInfo,
     TextContent,
     ToolConfig,
     ToolExecutionSpan,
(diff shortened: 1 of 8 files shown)


src/strands_evals/evaluators/goal_success_rate_evaluator.py
```diff
@@ -0,0 +1,109 @@
...
+        for trace in trajectory.traces:
+            for span in trace.spans:
+                if hasattr(span, "available_tools") and span.available_tools:
+                    available_tools = span.available_tools
+                    break
+            if available_tools:
+                break
+
+        # Format available tools
+        if available_tools:
```
poshinchen: nit: will these logic be re-used in other evaluators? If so we can move it to helper / util.

src/strands_evals/evaluators/goal_success_rate_evaluator.py
```diff
@@ -0,0 +1,109 @@
...
+
+class GoalSuccessRateEvaluator(Evaluator[InputT, OutputT]):
+    """Evaluates whether all user goals were successfully achieved in a conversation."""
+
+    evaluation_level = EvaluationLevel.CONVERSATION_LEVEL
+
+    _score_mapping = {
+        GoalSuccessScore.YES: 1.0,
+        GoalSuccessScore.NO: 0.0,
+    }
```
poshinchen, line 38: just to understand, it must match all goal otherwise it'll be NO?

jjbuck: Yes, that's the idea. The feedback that we got back when we designed and developed this approach was that this notion of goal success only made sense if it was "all or nothing." We had some signal that anything in between would just be confusing, e.g., "What does it mean for something to be 'partially successful'? Did the agent work or didn't it?"

src/strands_evals/evaluators/goal_success_rate_evaluator.py
jjbuck: Yes, this will be used by other evaluators that need to process tool calls. I'll refactor this into the base class in the next rev.

src/strands_evals/evaluators/evaluator.py
```diff
@@ -154,6 +156,50 @@ def _parse_tool_level(self, session: Session) -> list[ToolLevelInput]:
 
         return evaluator_inputs
 
+    def _parse_conversation_level(self, session: Session) -> ConversationLevelInput:
```
poshinchen, line 159: Discussed offline: follow-up action is to move this logic to extractor.

PR #24, opened 2025-10-30
refactor: extract trace parsing into composable TraceExtractor class

jjbuck:
## Description
This PR refactors trace parsing logic from the base evaluator class and some of the child classes to a new `TraceExtractor` class. In addition to providing a better separation of concerns, this also allows independent use of trace extraction outside of evaluators and facilitates compositionality of extractors within the evaluators.

Changes include:
- Extract trace parsing methods from Evaluator base class into new TraceExtractor
- Move _parse_turn_level, _parse_tool_level, and _parse_conversation_level to TraceExtractor
- TraceExtractor is initialized with EvaluationLevel and provides single extract() method
- Evaluator now uses TraceExtractor internally via composition pattern
- Update evaluator subclasses to call super().__init__() after setting evaluation_level
- Add comprehensive tests for TraceExtractor demonstrating composability
- Update existing tests to use TraceExtractor directly
- Add example demonstrating TraceExtractor composability patterns

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

Other (please describe): refactoring

## Testing

Updated tests and confirmed unit tests pass.

- [x ] I ran `hatch run prepare`

## Checklist
- [ x] I have read the CONTRIBUTING document
- [ x ] I have added any necessary tests that prove my fix is effective or my feature works
- [ x ] I have updated the documentation accordingly
- [ x ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [ x ] My changes generate no new warnings
- [ x ] Any dependent changes have been merged and published

----


src/strands_evals/evaluators/evaluator.py
```diff
@@ -36,6 +23,21 @@ class Evaluator(Generic[InputT, OutputT]):
...
+        Args:
+            trace_extractor: Custom trace extractor. If None and evaluation_level is set,
+                           a default TraceExtractor will be created.
+        """
+        if trace_extractor:
+            self._trace_extractor = trace_extractor
+        elif self.evaluation_level:
+            self._trace_extractor = TraceExtractor(self.evaluation_level)
+        else:
+            self._trace_extractor = None
```
poshinchen: nit: This can be removed using `_trace_extractor: TraceExtractor | None = None`

src/strands_evals/evaluators/evaluator.py
```diff
@@ -36,6 +23,21 @@ class Evaluator(Generic[InputT, OutputT]):
...
+
+    def __init__(self, trace_extractor: TraceExtractor | None = None):
+        """Initialize evaluator with optional custom trace extractor.
+
+        Args:
+            trace_extractor: Custom trace extractor. If None and evaluation_level is set,
+                           a default TraceExtractor will be created.
+        """
+        if trace_extractor:
+            self._trace_extractor = trace_extractor
```
poshinchen, line 36: Hmm this makes me think if we want to allow users to pass their own custom TraceExtractor, at this moment TraceExtractor cannot be overridden.

jjbuck: I had the same thought - I think we likely will want to allow users to pass in a custom TraceExtractor in the future. But it can be overridden, right (it's an optional argument in the evaluator constructor)?

src/strands_evals/evaluators/evaluator.py
jjbuck: Resolved.

PR #25, opened 2025-10-31
feat(conventions): added otel v1.37 conventions mapping

poshinchen:
## Description
Added OTEL v1.37+ semantic conventions mapping for the trace-based evaluator.
- Created `GenAIConventionVersion` enum, which [follows the semantic conventions values](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/):
  - **LEGACY**: Use legacy conventions (v1.36.0 or prior) with `gen_ai.system attribute`
            and separate message events (`gen_ai.user.message`, `gen_ai.choice`, etc.)
  - **LATEST_EXPERIMENTAL**: Use latest experimental conventions (v1.37+) with
           `gen_ai.provider.name` attribute and unified `gen_ai.client.inference.operation.details` events.
            Corresponds to OTEL's "gen_ai_latest_experimental" stability option.
The mapper checks for presence of `gen_ai.provider.name` to determine version.

Compared the transformed session with the old conventions and the latest conventions to make sure that the fields are not missed.

## Documentation PR
TBD

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/mappers/strands_in_memory_session_mapper.py
+++ b/src/strands_evals/mappers/strands_in_memory_session_mapper.py
@@ -2,6 +2,7 @@ import json
 import logging
 from collections import defaultdict
 from datetime import datetime, timezone
+from enum import Enum
 from typing import Any
 
 from opentelemetry.sdk.trace import ReadableSpan
@@ -27,10 +28,42 @@ from .session_mapper import SessionMapper
 logger = logging.getLogger(__name__)
 
 
+class GenAIConventionVersion(Enum):
+    """GenAI semantic convention versions following OTEL_SEMCONV_STABILITY_OPT_IN.
+
+    This enum aligns with OpenTelemetry's semantic convention stability options
+    as defined in OTEL_SEMCONV_STABILITY_OPT_IN environment variable.
+
+    Attributes:
+        LEGACY: Use legacy conventions (v1.36.0 or prior) with gen_ai.system attribute
+            and separate message events (gen_ai.user.message, gen_ai.choice, etc.)
+        LATEST_EXPERIMENTAL: Use latest experimental conventions (v1.37+) with
+            gen_ai.provider.name attribute and unified gen_ai.client.inference.operation.details events.
+            Corresponds to OTEL's "gen_ai_latest_experimental" stability option.
+    """
+
+    LEGACY = "legacy"
+    LATEST_EXPERIMENTAL = "gen_ai_latest_experimental"
+
+
 class StrandsInMemorySessionMapper(SessionMapper):
-    """Maps OpenTelemetry in-memory spans to Session format for evaluation."""
+    """Maps OpenTelemetry in-memory spans to Session format for evaluation.
+
+    Supports both legacy and latest GenAI semantic conventions:
+    - Latest (v1.37+): gen_ai.provider.name with unified gen_ai.client.inference.operation.details events
+    - Legacy: gen_ai.system with separate message events (gen_ai.user.message, gen_ai.choice, etc.)
+
+    The mapper automatically detects the convention version. Default to Legacy.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self._convention_version = GenAIConventionVersion.LEGACY
 
     def map_to_session(self, otel_spans: list[ReadableSpan], session_id: str) -> Session:
+        if otel_spans:
+            self._convention_version = self._detect_convention_version(otel_spans[0])
+
         traces_by_id = defaultdict(list)
         for span in otel_spans:
             trace_id = format(span.context.trace_id, "032x")
@@ -43,6 +76,26 @@ class StrandsInMemorySessionMapper(SessionMapper):
                 traces.append(trace)
 
         return Session(traces=traces, session_id=session_id)
+
+    def _detect_convention_version(self, span: ReadableSpan) -> GenAIConventionVersion:
+        """Detect which GenAI semantic convention version is being used.
+
+        Returns:
+            GenAIConventionVersion.LATEST_EXPERIMENTAL if using latest conventions,
+            GenAIConventionVersion.LEGACY otherwise
+        """
+        if span.attributes and "gen_ai.provider.name" in span.attributes:
+            return GenAIConventionVersion.LATEST_EXPERIMENTAL
+
+        return GenAIConventionVersion.LEGACY
+
+    def _use_latest_conventions(self) -> bool:
+        """Helper method to determine if latest conventions should be used.
+
+        Returns:
+            True if LATEST_EXPERIMENTAL, False if LEGACY
+        """
+        return self._convention_version == GenAIConventionVersion.LATEST_EXPERIMENTAL
 
     def _convert_trace(self, trace_id: str, otel_spans: list[ReadableSpan], session_id: str) -> Trace:
         converted_spans: list[InferenceSpan | ToolExecutionSpan | AgentInvocationSpan] = []
@@ -126,6 +179,16 @@ class StrandsInMemorySessionMapper(SessionMapper):
 
     def _convert_inference_span(self, span: ReadableSpan, session_id: str) -> InferenceSpan:
         span_info = self._create_span_info(span, session_id)
+
+        if self._use_latest_conventions():
+            messages = self._extract_messages_from_inference_details(span)
+        else:
+            messages = self._extract_messages_from_events(span)
+
+        return InferenceSpan(span_info=span_info, messages=messages, metadata={})
+
+    def _extract_messages_from_events(self, span: ReadableSpan) -> list[UserMessage | AssistantMessage]:
+        """Extract messages from legacy event format (gen_ai.user.message, etc.)."""
         messages: list[UserMessage | AssistantMessage] = []
 
         for event in span.events:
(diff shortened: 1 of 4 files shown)


src/strands_evals/mappers/strands_in_memory_session_mapper.py
```diff
@@ -28,9 +28,30 @@
...
+    """Maps OpenTelemetry in-memory spans to Session format for evaluation.
+
+    Supports both legacy and latest GenAI semantic conventions:
+    - Latest (v1.37+): gen_ai.provider.name with unified gen_ai.client.inference.operation.details events
+    - Legacy: gen_ai.system with separate message events (gen_ai.user.message, gen_ai.choice, etc.)
+
+    The mapper can auto-detect the convention version or accept an explicit setting. Default to Legacy.
+    """
+
+    def __init__(self, use_latest_conventions: bool | None = None):
```
jjbuck: Do we need to more more explicit about the exact version? Risks in the boolean flag with the term "latest" are

1. If the standard is incremented, the CX is a little disjointed until we come around and make any necessary upgrades to Strands Evals. E.g., if the "latest" today is vX and v(X+1) comes out tomorrow, then customers would come to Strands Evals and interpret "latest" to mean (X+1)....but in practice Strands Evals would only continue to support X until we make whatever changes are required to support (X+1).

2. Given that the OTEL semantic conventions seem to be explicit about the fact that they do **not** guarantee that they won't make breaking changes, it seems plausible that we might need to support various versions as an explicit enum. E.g., some customers might be on version X, some might be on Y, and some might be on Z. We lose that facility if it's just a boolean for the "latest" convention. And on a related note, it's reasonably clear what the "latest" convention is (with a caveat described above), but what is the CX when `use_latest_conventions = False`? Is that the version immediately before the latest, or some other version before that?

What do you think of the idea of making this an `enum`? That may require a little more configuration and restructuring of things, but I think that is likely the better long-term design pattern.

poshinchen: I've made them `enum`!

src/strands_evals/mappers/strands_in_memory_session_mapper.py
```diff
@@ -27,10 +28,52 @@
...
+            convention_version: GenAI semantic convention version to use.
+                Defaults to AUTO which auto-detects from span attributes.
+        """
+        self.convention_version = convention_version
 
     def map_to_session(self, otel_spans: list[ReadableSpan], session_id: str) -> Session:
+        # Auto-detect convention version if set to AUTO
+        if otel_spans and self.convention_version == GenAIConventionVersion.AUTO:
+            detected_version = self._detect_convention_version(otel_spans[0])
+            self.convention_version = detected_version
```
jjbuck: There's a subtle bug here: `self.convention_version` is mutated during `map_to_session`, which means the mapper cannot be safely reused. Once AUTO is resolved on the first call, subsequent calls will use that resolved version regardless of what spans are passed in.

Consider the following example.

```
# Step #0
mapper = StrandsInMemorySessionMapper(convention_version=GenAIConventionVersion.AUTO)

# Step #1 - First call: spans use v1.36 conventions
session1 = mapper.map_to_session(v1_36_spans, "session-1")
# mapper.convention_version is now V1_36

# Step #2 Second call: spans use v1.37 conventions  
session2 = mapper.map_to_session(v1_37_spans, "session-2") -> Still uses V1_36 because AUTO was already resolved. The user will expect to parse spans according to v1.37, but internally the mapper will use v1.36.
```

I think the mapper should have a fixed, immutable convention version. This provides more predictable behavior, ensuring that the mapper's parsing logic never changes after construction, regardless of what spans are passed to it.


 I know this would limit the ability to dynamically infer the OTEL version (at least as currently written), because as a practical matter the Mapper doesn't know what the version is until it sees a sample span. However, I *do* think there's still a way we can provide that convenience to users. What would you say to tweaking this such that the `_detect_convention_version()` function is a public helper method? This would result in the following CX.
 
 ```
#  Use case #1 - Customer knows their OTEL SDK version
mapper = StrandsInMemorySessionMapper(GenAIConventionVersion.V1_37)
session = mapper.map_to_session(spans, "session-1")
session = mapper.map_to_session(other_spans, "session-2") # -> Still v1.37

mapper = StrandsInMemorySessionMapper(GenAIConventionVersion.V1_38) # -> New mapper instance is v1.38
session = mapper.map_to_session(different_spans, "session-3") # -> parses traces using v1.38

# Use Case #2 - Auto detection helper method
version = StrandsInMemorySessionMapper.detect_convention_version(spans)
mapper = StrandsInMemorySessionMapper(version)
session = mapper.map_to_session(spans, "session-1")
```

## Proposed Changes
1. Make `_detect_convention_version()` a public static method (or class method)
2. Remove the `AUTO` resolution logic from `map_to_session`.

File: src/strands_evals/mappers/strands_in_memory_session_mapper.py [hunk 3]
```diff
@@ -27,10 +28,52 @@
...
+        LEGACY: Use legacy conventions (v1.36.0 or prior) with gen_ai.system attribute
+            and separate message events (gen_ai.user.message, gen_ai.choice, etc.)
+        LATEST_EXPERIMENTAL: Use latest experimental conventions (v1.37+) with
+            gen_ai.provider.name attribute and unified gen_ai.client.inference.operation.details events.
+            Corresponds to OTEL's "gen_ai_latest_experimental" stability option.
+        AUTO: Automatically detect convention version from span attributes.
+            Checks for presence of gen_ai.provider.name to determine version.
+    """
+
+    LEGACY = "legacy"
```
jjbuck, line 45: m still a bit concerned about `LATEST_EXPERIMENTAL` for the same reasons I mentioned with the boolean flag. When OTEL releases v1.38, it's unclear whether LATEST_EXPERIMENTAL refers to v1.37 or v1.38.

The risk is that whenever new versions come out, we'll have to make changes (potentially breaking changes) under the `LATEST_EXPERIMENTAL` code paths that update the behavior to whatever the new version is such that everything that was _previously_ "latest" will have to be moved under a _different_ code path, and we'll need new business logic under the `LATEST_EXPERIMENTAL` code paths. 

Also, from a developer experience perspective, it's not quite clear to a future contributor what "latest" meant when we wrote it.

What do you think about making the versions explicit instead?

```
class GenAIConventionVersion(Enum):
    """OpenTelemetry GenAI semantic convention versions."""
    V1_36 = "v1.36"  # gen_ai.system attribute, separate message events
    V1_37 = "v1.37"  # gen_ai.provider.name attribute, unified inference.operation.details events
    AUTO = "auto"    # Detect version from span attributes
 ```
 
 By making it explicit like this, customers know exactly which version of OTEL they're targeting, we can seamlessly support new versions by adding an enum without changing the definitions of existing enums, and we can safely deprecate old versions explicitly if/when appropriate.

File: src/strands_evals/mappers/strands_in_memory_session_mapper.py (hunk 2 above)
Comment (poshinchen) [reply src/strands_evals/mappers/strands_in_memory_session_mapper.py]: yeah thought of that too:

1. Behaviors of the versions:
  - auto: mapper will detect what the conventions are used and change it every when parsing
  - Legacy: Always use old semantic conventions
  - LATEST_EXPERIMENTAL: Always use the latest semantic conventions (based on https://opentelemetry.io/docs/specs/semconv/gen-ai/)
It will be users responsibility to make sure that the traces conform the conventions if it is set to legacy / latest

> The risk is that whenever new versions come out, we'll have to make changes (potentially breaking changes) under the LATEST_EXPERIMENTAL code paths that update the behavior to whatever the new version is such that everything that was previously "latest" will have to be moved under a different code path, and we'll need new business logic under the LATEST_EXPERIMENTAL code paths.

IMO, if there's any other breaking change, it'll be another value in `OTEL_SEMCONV_STABILITY_OPT_IN`, and we will use that as another enum. For the rest of the "soft changes" we can just include it into  `LATEST_EXPERIMENTAL`. Pinning the enum via version is a bit weird because we are not adding all the versions as enum.

However, this makes me wonder, why don't we auto detect the conventions every time, it is not required for users to configure it, and the mapper is flexible to different conventions of traces.

The only possible reason that `convention_version` would exist is for other classes to use it (which I don't think we need it as of now). How about keeping `_detect_convention_version` and remove all the enums?

Comment (jjbuck) [reply src/strands_evals/mappers/strands_in_memory_session_mapper.py]: Thanks for posting the link. Now that I go back and read that, I see OTEL is explicitly encouraging this nomenclature of "legacy", "gen_ai_latest_experimental", etc. Let's leave these enums as-is.

However, I **_do_** think we need to address the mutation issue with `AUTO`. The instance attribute `self.convention_version` is mutated during map_to_session, which means the mapper can't be safely reused. Once AUTO is resolved on the first call, subsequent calls will use that resolved version regardless of what spans are passed in.

Here's the problematic scenario (copy-pasted from my comment above).
```
# Step #0: Create mapper with AUTO
mapper = StrandsInMemorySessionMapper(convention_version=GenAIConventionVersion.AUTO)

# Step #1: First call with v1.36 spans
session1 = mapper.map_to_session(v1_36_spans, "session-1")
# mapper.convention_version is now LEGACY

# Step #2: Second call with v1.37 spans
session2 = mapper.map_to_session(v1_37_spans, "session-2")
# BUG: Still uses LEGACY because AUTO was already resolved in Step #1
# The v1.37 spans will be parsed incorrectly
```

A few options to fix this:

1. Detect per-call without mutation - When `convention_version == AUTO`, detect the version from spans on each call and thread it through the helper methods without mutating `self.convention_version`. This is the "auto detect conventions every time" concept you mentioned. I think the argument against this is that it would likely require excessive threading/plumbing of the version enum down into all the nitty gritty helper methods.

2. Remove `AUTO` from enum, and provide detection helper. Concretely, make `detect_convention_version()` a public static method and require customers to call it explicitly if they want auto-detection.

3. Document the limitation. Keep the current behavior but clearly document that `AUTO` locks on first use and the mapper shouldn't be reused with different span types.

4. Remove AUTO and detection entirely. Customers must explicitly know and configure their OTEL convention version.

I'm still leaning towards option 2. That removes the bug/gotcha, keeps the mapper simpler (no threading of the version specification), and still provides detection as a utility method. What do you think?

Comment (poshinchen) [reply src/strands_evals/mappers/strands_in_memory_session_mapper.py]: updated based on the discussion offline.

File: src/strands_evals/mappers/strands_in_memory_session_mapper.py [hunk 4]
```diff
@@ -27,10 +28,38 @@
 logger = logging.getLogger(__name__)
 
 
+class GenAIConventionVersion(Enum):
```
zastrowm, line 31: Can Strands customers use this, or is it purely internal?

src/strands_evals/mappers/strands_in_memory_session_mapper.py
```diff
@@ -44,6 +73,26 @@ def map_to_session(self, otel_spans: list[ReadableSpan], session_id: str) -> Ses
...
+
+        return GenAIConventionVersion.LEGACY
+
+    def _use_latest_conventions(self) -> bool:
+        """Helper method to determine if latest conventions should be used.
+
+        Returns:
+            True if LATEST_EXPERIMENTAL, False if LEGACY
+        """
+        return self._convention_version == GenAIConventionVersion.LATEST_EXPERIMENTAL
```
zastrowm, line 98: I think we either have to initialize `self._convention_version` in `__init__` or do a check to make sure the attribute is set first - otherwise we run the risk of `self._convention_version` not being available and throwing an exception?

src/strands_evals/mappers/strands_in_memory_session_mapper.py
poshinchen: Purely internal as of now. Users shouldn't get bothered about which version of the traces.

PR #27, opened 2025-11-04
feat: Add FaithfulnessEvaluator

jjbuck:
## Description

1. Add FaithfulnessEvaluator to assess whether agent responses conflict with conversation history
2. Implement 5-level scoring system (Not At All, Not Generally, Neutral/Mixed, Generally Yes, Completely Yes
3. Add faithfulness prompt templates (v0) with evaluation guidelines
4. Include example usage and comprehensive unit tests
5. Update naming of low-level data types (turn -> trace, conversation -> session) to align better with OTEL conventions.

## Related Issues

1. #22 (goal success rate)
6. #16 (helpfulness)

## Documentation PR

N/A

## Type of Change

1. New feature ("Faithfulness Evaulator")
2. Refactor Update naming of low-level data types (turn -> trace, conversation -> session) to align better with OTEL conventions.

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/evaluators/faithfulness_evaluator.py
+++ b/src/strands_evals/evaluators/faithfulness_evaluator.py
@@ -1,0 +1,105 @@
+from enum import Enum
+
+from pydantic import BaseModel, Field
+from strands import Agent
+from typing_extensions import TypeVar
+
+from ..types.evaluation import EvaluationData, EvaluationOutput
+from ..types.trace import EvaluationLevel, TraceLevelInput
+from .evaluator import Evaluator
+from .prompt_templates.faithfulness import get_template
+
+InputT = TypeVar("InputT")
+OutputT = TypeVar("OutputT")
+
+
+class FaithfulnessScore(str, Enum):
+    """Categorical faithfulness ratings."""
+
+    NOT_AT_ALL = "Not At All"
+    NOT_GENERALLY = "Not Generally"
+    NEUTRAL = "Neutral/Mixed"
+    GENERALLY_YES = "Generally Yes"
+    COMPLETELY_YES = "Completely Yes"
+
+
+class FaithfulnessRating(BaseModel):
+    """Structured output for faithfulness evaluation."""
+
+    reasoning: str = Field(description="Step by step reasoning to derive the final score")
+    score: FaithfulnessScore = Field(description="Categorical faithfulness rating")
+
+
+class FaithfulnessEvaluator(Evaluator[InputT, OutputT]):
+    """Evaluates faithfulness of agent responses against conversation history."""
+
+    evaluation_level = EvaluationLevel.TRACE_LEVEL
+
+    _score_mapping = {
+        FaithfulnessScore.NOT_AT_ALL: 0.0,
+        FaithfulnessScore.NOT_GENERALLY: 0.25,
+        FaithfulnessScore.NEUTRAL: 0.5,
+        FaithfulnessScore.GENERALLY_YES: 0.75,
+        FaithfulnessScore.COMPLETELY_YES: 1.0,
+    }
+
+    def __init__(
+        self,
+        version: str = "v0",
+        model: str | None = None,
+        system_prompt: str | None = None,
+    ):
+        super().__init__()
+        self.system_prompt = system_prompt if system_prompt is not None else get_template(version).SYSTEM_PROMPT
+        self.version = version
+        self.model = model
+
+    def evaluate(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
+        parsed_input = self._get_last_turn(evaluation_case)
+        prompt = self._format_prompt(parsed_input)
+        evaluator_agent = Agent(model=self.model, system_prompt=self.system_prompt, callback_handler=None)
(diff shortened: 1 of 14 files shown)


src/strands_evals/evaluators/helpfulness_evaluator.py
```diff
@@ -88,23 +88,30 @@ def _get_last_turn(self, evaluation_case: EvaluationData[InputT, OutputT]) -> Tu
...
         """Format evaluation prompt from parsed turn data."""
         parts = []
 
-        if parsed_input.conversation_history:
+        if parsed_input.session_history:
             history_lines = []
-            for msg in parsed_input.conversation_history:
+            for msg in parsed_input.session_history:
+                if isinstance(msg, list):
+                    continue  # Skip tool execution lists
```
poshinchen, line 126: Why do we skip the `ToolExecutions`? Is it because this evaluator is supposed to evaluate only the `User / Assistant` messages?

src/strands_evals/evaluators/helpfulness_evaluator.py
```diff
@@ -88,23 +88,30 @@ def _get_last_turn(self, evaluation_case: EvaluationData[InputT, OutputT]) -> Tu
...
-        if parsed_input.conversation_history:
-            last_msg = parsed_input.conversation_history[-1]
-            if hasattr(last_msg, "content") and last_msg.content and hasattr(last_msg.content[0], "text"):
+        if parsed_input.session_history:
+            last_msg = parsed_input.session_history[-1]
+            if (
+                not isinstance(last_msg, list)
+                and hasattr(last_msg, "content")
+                and last_msg.content
+                and hasattr(last_msg.content[0], "text")
```
poshinchen: nit, don't have to do it now:
We can have a private method for this validation with clearer naming (maybe `_has_content_text` or `_has_text_content`).

```
and hasattr(last_msg, "content")
                and last_msg.content
                and hasattr(last_msg.content[0], "text")
                ```

File: src/strands_evals/evaluators/faithfulness_evaluator.py [hunk 3]
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    def __init__(
+        self,
+        version: str = "v0",
+        model: str | None = None,
+        system_prompt: str | None = None,
+    ):
+        super().__init__()
+        self.system_prompt = system_prompt if system_prompt is not None else get_template(version).SYSTEM_PROMPT
```
poshinchen, line 53: nit: do we need `if system_prompt is not None` or just `if system_prompt`? 
This more of like "if user pass empty string as system prompt, do we want to take it as is".

src/strands_evals/evaluators/faithfulness_evaluator.py
```diff
@@ -0,0 +1,105 @@
...
+                "No turn-level inputs could be parsed from the trajectory. "
+                "Ensure actual_trajectory is a Session with at least one AgentInvocationSpan."
+            )
+        return parsed_inputs[-1]
+
+    def _format_prompt(self, parsed_input: TraceLevelInput) -> str:
+        """Format evaluation prompt from parsed turn data."""
+        parts = []
+
+        if parsed_input.session_history:
```
poshinchen, line 89: hmm, `session_history` always exists right? you don't need if statement then? 

https://github.com/strands-agents/evals/blob/db136bf218f3019d636dd34a3e36ecd9ed865795/src/strands_evals/types/trace.py#L165

src/strands_evals/evaluators/helpfulness_evaluator.py
jjbuck: Yes, you've got the right idea. The HelpfulnessEvaluator is focused on evaluating the conversational flow (user/assistant messages), not the tool execution details. 

Although your comment here is making me realize that this is probably a little too generic of a type check. It does technically work, because the type of `session_history` is defined as `session_history: list[Union[UserMessage, list[ToolExecution], AssistantMessage]]`....so as currently implemented anything that's a `list` will inherently be `ToolExecution` data....but that's a bit of a technicality. I'd like to tighten this up by making the check more explicit.

src/strands_evals/evaluators/helpfulness_evaluator.py
jjbuck: Ack - updated.

src/strands_evals/evaluators/faithfulness_evaluator.py
jjbuck: I think you've got the right read. As written the behavior will accept an empty string as a valid value for `system_prompt`.....that would be kind of a peculiar choice, but one that would technically be possible. We could use `if system_prompt`, but the empty string would evaluate as falsy....we can revisit that if we want to. At the very least, what's written here is consistent with the implementation for the other evaluators.

src/strands_evals/evaluators/faithfulness_evaluator.py
jjbuck: It's true that `session_history` always exists (it's a required field), but it can be an empty list. The `if parsed_input.session_history:` check is verifying the list is non-empty before adding the history section to the prompt. If the list is empty, we skip adding the "# Previous turns:" header entirely.

PR #28, opened 2025-11-06
feat: add ActorSimulator for multi-turn conversation evaluation

jjbuck:
## Description
Introduces ActorSimulator framework for simulating realistic actors (typically users)  in multi-turn conversations with agents under test. Enables systematic evaluation of conversational agents through synthetic user interactions.

Key capabilities:
- Generic ActorSimulator class configurable with arbitrary system prompts
- from_case_for_user_simulator() factory method to condition simulator to act as a user on the basis of a given `Case`
- Automatic profile generation from test cases using LLM inference
- Built-in goal completion assessment tool for conversation evaluation
- Support for custom tools and behaviors

Design principles:
- Generic base class (ActorSimulator) with specialized factory methods
- Clear separation: __init__() for generic construction, factory for specialization
- Optional task_description in Case metadata (handles vague initial queries)

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/simulation/README.md
+++ b/src/strands_evals/simulation/README.md
@@ -1,0 +1,323 @@
+# Actor Simulator
+
+A framework for simulating realistic multi-turn conversations with AI-powered actors for agent evaluation.
+
+## Overview
+
+ActorSimulator creates realistic actor personas that interact with agents in multi-turn conversations. It automatically generates actor profiles from test cases, maintains conversation context, and produces contextually appropriate responses aligned with the actor's goals and traits.
+
+## Quick Start
+
+```python
+from strands import Agent
+from strands_evals import ActorSimulator, Case
+
+# Create agent under test
+agent = Agent(system_prompt="You are a helpful travel assistant.", callback_handler=None)
+
+# Create test case
+case = Case(
+    input="I want to plan a trip to Tokyo with hotel and activities",
+    metadata={"task_description": "Complete travel package arranged"}
+)
+
+# Create user simulator with max_turns
+user_sim = ActorSimulator.from_case_for_user_simulator(case=case, max_turns=5)
+
+# Run conversation
+user_message = case.input
+while user_sim.has_next():
+    agent_response = agent(user_message)
+    user_result = user_sim.act(str(agent_response))
+    user_message = str(user_result.structured_output.message)
+```
+
+## How It Works
+
+1. **Profile Generation**: Creates a realistic actor profile with traits, context, and goals from the test case
+2. **Conversation Initialization**: Sets up conversation with a greeting and the actor's initial query
+3. **Contextual Responses**: Generates responses that maintain consistency with the actor's profile and goals
+4. **Goal Tracking**: Built-in tool allows actors to assess progress toward their goals
+
+## API Reference
+
+### ActorSimulator
+
+Main class for simulating actor behavior in conversations.
+
+#### Factory Method (Recommended)
+
+```python
+ActorSimulator.from_case_for_user_simulator(
+    case: Case,
+    system_prompt_template: str | None = None,
+    tools: list | None = None,
+    model: str | None = None,
+    max_turns: int = 10
+) -> ActorSimulator
+```
+
+Creates an ActorSimulator configured as a user simulator from a test case. Automatically generates a realistic actor profile from `case.input` and optionally `case.metadata["task_description"]`.
+
+**Parameters:**
+- `case`: Test case with input (initial query) and optional task_description in metadata
+- `system_prompt_template`: Custom system prompt template (uses default if None)
+- `tools`: Additional tools for the actor (defaults to goal completion tool only)
+- `model`: Model identifier (uses Strands default if None)
+- `max_turns`: Maximum number of conversation turns (default: 10)
+
+**Example:**
+```python
+case = Case(
+    input="I need help booking a flight to Paris",
+    metadata={"task_description": "Book round-trip flight under $800"}
+)
+
+user_sim = ActorSimulator.from_case_for_user_simulator(
+    case=case,
+    max_turns=5
+)
+```
+
+#### Direct Initialization
+
+```python
+ActorSimulator(
+    actor_profile: ActorProfile,
+    initial_query: str,
+    system_prompt_template: str,
+    tools: list | None = None,
+    model: str | None = None,
+    max_turns: int = 10
+)
+```
+
+Initialize with an existing actor profile. Use this when you have a pre-defined profile instead of generating one from a test case.
+
+**Parameters:**
+- `actor_profile`: ActorProfile object with traits, context, and actor_goal
+- `initial_query`: The actor's first query or message
+- `system_prompt_template`: Template string for actor behavior (formatted with profile)
+- `tools`: Additional tools for the actor
+- `model`: Model identifier
+- `max_turns`: Maximum number of conversation turns (default: 10)
+
+#### Methods
+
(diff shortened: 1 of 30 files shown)


src/strands_evals/types/simulation/actor.py
poshinchen, line 1: Let's move this to types

src/strands_evals/simulation/profiles/__init__.py
poshinchen, line 1: not sure about the path pattern, should we use `/defaults` as directory name? or `/profiles`

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -0,0 +1,177 @@
...
+        result = profile_agent(profile_prompt, structured_output_model=ActorProfile)
+        return result.structured_output
+
+    def __init__(
+        self,
+        actor_profile: ActorProfile,
+        initial_query: str,
+        system_prompt_template: str,
+        tools: list | None = None,
+        model: str | None = None,
```
poshinchen, line 143: As discussed, let's add max_turn either in `__init__` or in `act` (personally I would prefer in `__init__`.

As of stopping point, let's use `has_next(self)` so that user doesn't need to do `While True:` but `While simulator.has_next()` to validate. How do you feel?

jjbuck: Great idea. Making this change in the next rev so the developer experience will be significantly improved.

src/strands_evals/types/simulation/actor.py
jjbuck: Fixed in next rev.

src/strands_evals/simulation/profiles/__init__.py
jjbuck: I like the idea of `/profiles`. `/defaults` implies something more generic than what this directory actually contains.

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -0,0 +1,199 @@
...
+    ActorSimulator wraps a Strands Agent configured to behave as a specific actor
+    (typically a user) in conversation scenarios. It maintains conversation history,
+    generates contextually appropriate responses, and can assess goal completion.
+
+    Attributes:
+        agent: The underlying Strands Agent configured with actor behavior.
+        actor_profile: The actor's profile containing traits, context, and goal.
+        initial_query: The actor's first query in the conversation.
+        conversation_history: List of conversation messages in Strands format.
+        model_id: Model identifier for the underlying agent.
```
poshinchen, line 32: nit: consider putting examples similar to our [agent class](https://github.com/strands-agents/sdk-python/blob/main/src/strands/agent/agent.py#L636-L641).

src/strands_evals/simulation/README.md
```diff
@@ -0,0 +1,374 @@
...
+    max_turns = 5
+    
+    # Initial User act
+    user_message = case.input
+    conversation.append({"role": "user", "content": user_message})
+    
+    for _ in range(max_turns):
+        # Check for completion
+        if "<stop/>" in user_message:
+            break
```
poshinchen: Will need to update these lines.

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -0,0 +1,199 @@
...
+
+        Returns:
+            AgentResult containing the actor's structured response with reasoning and message.
+        """
+        response = self.agent(agent_message.strip(), structured_output_model=ActorResponse)
+        self._turn_count += 1
+        self._last_message = str(cast(ActorResponse, response.structured_output).message)
+        return response
+
+    def has_next(self, max_turns: int = 10) -> bool:
```
poshinchen: What's the benefit of having `max_turns` here instead of during `__init__`? Are we expecting the max_turns to be dynamic in the long-term?

src/strands_evals/simulation/tools/goal_completion.py
```diff
@@ -0,0 +1,105 @@
...
+
+from strands import Agent, tool
+from typing_extensions import Any
+
+from strands_evals.simulation.prompt_templates.goal_completion import GOAL_COMPLETION_PROMPT
+
+logger = logging.getLogger(__name__)
+
+
+class ToolExecutionError(Exception):
```
poshinchen: Follow-up: we should move this into /types/exceptions (e.g. [strands sdk-python](https://github.com/strands-agents/sdk-python/blob/3061116ebe839c1c8a3182eb736429c3fc4411b0/src/strands/types/exceptions.py))

src/strands_evals/simulation/tools/goal_completion.py
```diff
@@ -0,0 +1,105 @@
...
+    """Exception raised when tool execution fails."""
+
+    def __init__(self, message: str, tool_name: str | None = None):
+        self.tool_name = tool_name
```
poshinchen: hmm `tool_name` could be None, do you want a default tool_name?

src/strands_evals/simulation/tools/goal_completion.py
```diff
@@ -0,0 +1,105 @@
...
+            raise ToolExecutionError("No valid conversation turns found", tool_name="get_conversation_goal_completion")
+
+        return "\n\n".join(formatted_turns)
+
+    except Exception as e:
+        if isinstance(e, ToolExecutionError):
+            raise
+        raise ToolExecutionError(
+            f"Error formatting conversation: {str(e)}", tool_name="get_conversation_goal_completion"
+        ) from e
```
poshinchen: nit:

```
except ToolExecutionError:
    raise
except Exception as e:
    raise ToolExecutionError(
        f"Error formatting conversation: {str(e)}", 
        tool_name="get_conversation_goal_completion"
    ) from e
```

And, ` {str(e)}` may not be required since the original exception message will be in the trace, right?

src/strands_evals/simulation/README.md
jjbuck: Good catch - need to update the README.

src/strands_evals/simulation/actor_simulator.py
jjbuck: For any reasonable use case we have ahead of us, `max_turns` will likely not be dynamic. I'll just make sure this is a private instance attribute so that it's clear to the client that this isn't something we expect them to mess around with explicitly.

src/strands_evals/simulation/actor_simulator.py
jjbuck: Ack - will revise to improve the developer experience.

src/strands_evals/simulation/tools/goal_completion.py
jjbuck: I'm actually not a fan of these custom exceptions here. They really don't add anything meaningful and aren't handled any differently than standard `ValueError` and `RuntimeError`, so I'm going to remove this in the next rev (which will also reduce the surface area of this feature a little bit, too).

src/strands_evals/simulation/actor_simulator.py
poshinchen: This can be the follow-up work. We have others to add together!

src/strands_evals/simulation/tools/goal_completion.py
jjbuck: Will remove.

src/strands_evals/simulation/tools/goal_completion.py
jjbuck: Will refactor when I update the exceptions.

PR #29, opened 2025-11-07
feat: ToolSelectionAccuracyEvaluator 

smeetd159:
## Description

- Add ToolSelectionAccuracyEvaluator to assess whether agent responses conflict with conversation history
- Implement 2-level scoring system (Yes, No)
- Add tool selection prompt templates (v0) with evaluation guidelines
- Include example usage and comprehensive unit tests
- Separate tool level execution span in trace_extractor to change append order of UserMessage objects and tool details to the session history 

## Related Issues

1. (https://github.com/strands-agents/evals/pull/27) (goal success rate)
2.  #16 (helpfulness)

## Documentation PR

NA

## Type of Change

1. New feature
3. Separate tool level execution span in trace_extractor to change append order of UserMessage objects and tool details to the session history 

## Testing

## Checklist

----

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,8 +1,0 @@
-# Default ignored files
-/shelf/
-/workspace.xml
-# Editor-based HTTP Client requests
-/httpRequests/
-# Datasource local storage ignored files
-/dataSources/
-/dataSources.local.xml


src/strands_evals/extractors/trace_extractor.py
poshinchen, line 1: I believe you've tested the results and confirmed the structured-data is the same?

src/strands_evals/evaluators/tool_selection_accuracy_evaluator.py
```diff
@@ -0,0 +1,113 @@
...
+
+        return results
+
+    async def evaluate_async(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
+        tool_inputs = self._parse_trajectory(evaluation_case)
+        results = []
+
+        for tool_input in tool_inputs:
+            prompt = self._format_prompt(tool_input)
+            evaluator_agent = Agent(model=self.model, system_prompt=self.system_prompt, callback_handler=None)
```
poshinchen, line 75: nit: hmm if there are 50 tools, this will initiate the Agent 50 times.... 

I would be fine with it now, but I am interesting if the performance will be impacted if you only reset the messages and the state.

.idea/.gitignore
poshinchen, line 1: why do we need this file?

src/strands_evals/extractors/trace_extractor.py
jjbuck: Yes, tested via the unit tests and re-running all the example scripts.

.idea/.gitignore
smeetd159: Missed it. I'll remove this.

src/strands_evals/evaluators/tool_selection_accuracy_evaluator.py
smeetd159: Strands Agent doesn't have any option to clear the state, or the message history via method. There is however a way to clear it by setting `agent.messages = []` but if the state history is leaked between the tool spans, it'll affect the evaluation outcome.

PR #40, opened 2025-11-17
feat: allow different providers for evaluators

poshinchen:
## Description
Allow users to provide strands Model so the evaluator Agent can use different model providers.

## Related Issues
N/A

## Documentation PR
N/A

## Type of Change

New feature

## Testing

## Checklist

----


src/strands_evals/evaluators/evaluator.py
```diff
@@ -38,6 +41,26 @@ def __init__(self, trace_extractor: TraceExtractor | None = None):
         elif self.evaluation_level:
             self._trace_extractor = TraceExtractor(self.evaluation_level)
 
+    def _get_model_id(self, model: Union[Any, str, None]) -> str:
```
jjbuck: Based on L57 below, it looks like we can accept an input of type `Model`. Should we make that one of that the enumerated types in the union?

poshinchen: done!

Issue #41, opened 2025-11-17
[FEATURE] Allow customer to pass file_path for dataset generation and report


poshinchen:
### Problem Statement

Allow customer to pass file_path for dataset generation and report

### Use Case

Allow customer to pass file_path for dataset generation and report

PR #42, opened 2025-11-17
chore: allow passing path into file_name

poshinchen:
## Description
Allow customer to pass file path so the dataset & report will be generated correctly.

Didn't add unit tests because it needs to mock some underlying dependencies. But I tested locally with different scenarios:

1. `dataset.to_file("async_bank_tools_tool_select_dataset")`
    - generated **/path/to/code/async_bank_tools_tool_select_dataset.json**
2. `dataset.to_file("/path/to/my-dataset/async_bank_tools_tool_select_dataset")`
    - generated **/path/to/my-dataset/async_bank_tools_tool_select_dataset.json**
3. `dataset.to_file("test-dataset/async_bank_tools_tool_select_dataset")`
    - generated **/path/to/code/test-dataset/async_bank_tools_tool_select_dataset.json**

## Related Issues
https://github.com/strands-agents/evals/issues/41

## Documentation PR

N/A

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/dataset.py
+++ b/src/strands_evals/dataset.py
@@ -1,7 +1,7 @@
 import asyncio
 import json
-import os
 from collections.abc import Callable
+from pathlib import Path
 
 from typing_extensions import Any, Generic, TypeVar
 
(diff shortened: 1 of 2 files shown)


src/strands_evals/dataset.py
```diff
@@ -439,13 +439,31 @@ def to_file(self, file_name: str, format: str = "json", directory: str = "datase
         Write the dataset to a file.
 
         Args:
-            file_name: Name of the file without extension.
-            format: The format of the file to be saved.
-            directory: Directory to save the file (default: "dataset_files").
+            file_name: Either a file name without extension (e.g., "my_dataset") or a full file path
```
jjbuck: I think the arguments here are starting to lead to over-specification.

1. The `directory` has a default value but is ignored when `file_name` contains a path separator. It's a small thing, but this hits the developer experience a little bit by requiring more cognitive overhead to make sense of things.

2. The`file_name` accepts filenames, relative paths, AND absolute paths. Strictly speaking, that's not a filename.

3. I know this is technically outside the scope of this PR, but since it's all part of a larger idea I'm building towards to suggest refactoring this -- Only JSON is supported, so why do we parameterize it? We should either hardcode it or actually make it drive serialization behavior. Today, the serialization is implicitly the `self.to_dict()` method. This `to_file` method should be agnostic to file type -- it's the job of whatever serializes the data to know about the supported file formats. Either this `to_file` method should have the logic to assign the appropriate format to a possible selection of serializers, or this `to_file` method should just take a format from the user and plumb it through to the appropriate serializer. Right now it takes in an argument from the user, but it has no effect on the serializer, so the file extension is just an arbitrary suffix on the filename.

4. Replacing `os.path` operations with `Path` is cleaner.

 If we have time to make a small breaking change prior to public launch, I'd recommend we try to radically simplify this.
**Recommendation**: Simplify to a single `path` parameter. Let the caller decide the path structure. Something like:

```python
def to_file(self, path: Path | str):
    """Write dataset to JSON file. Path can be relative or absolute."""
    path = Path(path).with_suffix(".json")
    path.parent.mkdir(parents=True, exist_ok=True)
    with path.open("w") as f:
        json.dump(self.to_dict(), f, indent=2)
```

poshinchen: Updated a bit based on our discussion!

src/strands_evals/dataset.py
```diff
@@ -434,21 +434,45 @@ def to_dict(self) -> dict:
...
+                  Only .json format is supported.
+
+        Raises:
+            ValueError: If the path has a non-JSON extension.
         """
-        os.makedirs(directory, exist_ok=True)
-        if format == "json":
-            with open(f"{directory}/{file_name}.json", "w") as f:
-                json.dump(self.to_dict(), f, indent=2)
+        if "." in os.path.basename(path):
```
jjbuck: I think we can streamline the implementation by taking the input `path` , leveraging `pathlib.Path` and letting that class do all the heavy lifting, e.g.,
```
path = Path(path)
path = path.with_suffix(".json")
path.parent.mkdir(parents=True, exist_ok=True)
```

src/strands_evals/dataset.py
```diff
@@ -434,21 +434,45 @@ def to_dict(self) -> dict:
...
+            file_path = path
         else:
-            raise Exception(f"Format {format} is not supported.")
+            file_path = f"{path}.json"
+
+        # Handle absolute vs relative paths
+        if os.path.isabs(file_path):
+            final_path = file_path
+        else:
+            final_path = os.path.join(os.getcwd(), file_path)
```
jjbuck: This is a "nit" if we end up using `pathlib.Path`, but do we need to explicitly use `os.getcwd()`? I think the two approaches below are functionally equivalent. What does stacking `os.path.join(os.getcwd(),..)` achieve? I could be wrong about the intended use case!

```
# Approach #1
with open(os.path.join(os.getcwd(), "data/file.json"), "w") as f:
    # do something

# Approach #2
with open("data/file.json", "w") as f:
    # do something
    ```

PR #67, opened 2025-12-08
fix: replace deprecated structured_output methods with new API

jsamuel1:
Update OutputEvaluator to use the new `structured_output_model` parameter instead of the deprecated `structured_output` and `structured_output_async` methods.

## Changes
- Replace `evaluator_agent.structured_output(EvaluationOutput, prompt)` with `evaluator_agent(prompt, structured_output_model=EvaluationOutput)`
- Replace `await evaluator_agent.structured_output_async(EvaluationOutput, prompt)` with `await evaluator_agent.invoke_async(prompt, structured_output_model=EvaluationOutput)`

## Testing
All existing tests pass.


uv.lock
afarntrog, line 1: PR looks great. Let's just exclude this `uv.lock` file

jsamuel1:
Done - removed uv.lock from the PR.

Issue #61, opened 2025-12-03
[FEATURE] Enable retry for task execution


poshinchen:
### Problem Statement

Add retry for task execution otherwise throttling / serviceUnavailable exception will cause the evaluation to fail.

### Proposed Solution

Retry the task with min / max delays for each task execution

### Use Case

Retry the task with min / max delays for each task execution

PR #107, opened 2026-01-27
Feat/retry throttled tenacity

poshinchen:
## Description
When executing the evaluation, users would get throttled in task execution / evaluator execution and the whole test case becomes meaningless. Therefore adding the retry mechanism to retry the case execution.

This PR adds retry logic with exponential backoff for handling throttling exceptions during evaluation execution using [tenacity](https://github.com/jd/tenacity). The changes cover both **task execution** and **evaluator execution** can be retried from throttling errors while maintaining proper error isolation.

### Retry Configuration
  - `_MAX_RETRY_ATTEMPTS = 6`
  - `_INITIAL_RETRY_DELAY = 4` seconds
  - `_MAX_RETRY_DELAY = 240` seconds (4 minutes)

### Changes
**- Task Execution Retry Logic**
  - **Async path (`_worker` method)**: Added retry loop around task execution with exponential backoff (4s → 8s → 16s → 32s → 64s → 128s, max 6 attempts)
  - **Sync path (`run_evaluations` method)**: Added retry loop around task execution matching async behavior
  - Both paths now retry on throttling exceptions and fail immediately on non-throttling exceptions

**- Evaluator Error Isolation**
  - Added outer try-catch blocks around evaluator retry loops in both sync and async paths
  - Non-throttling evaluator exceptions are now caught and recorded as failed evaluations (score=0)
  - One evaluator failing no longer prevents other evaluators from running
  - Preserves existing retry logic for throttling exceptions during evaluation

**- Exceptions**
  - Detected the following exception types to retry:
    - strands.types.exceptions.ModelThrottledException
    - strands.types.exceptions.EventLoopException
    - botocore.exceptions.ClientError with ThrottlingException, TooManyRequestsException, RequestLimitExceeded, ServiceUnavailable, 

**- Minor changes**
  - **Division by zero in `_default_aggregator`**: Returns `(0.0, False, "No evaluation outputs produced")` for empty outputs list

## Related Issues

https://github.com/strands-agents/evals/issues/61

## Documentation PR
N/A

## Type of Change
New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/experiment.py
+++ b/src/strands_evals/experiment.py
@@ -2,14 +2,18 @@ import asyncio
 import json
 import logging
 import os
-import time
 from collections.abc import Callable
 from pathlib import Path
 from typing import cast
 
-from botocore.exceptions import ClientError
 from opentelemetry.trace import format_trace_id
-from strands.types.exceptions import EventLoopException, ModelThrottledException
+from tenacity import (
+    RetryError,
+    retry,
+    retry_if_exception,
+    stop_after_attempt,
+    wait_exponential,
+)
 from typing_extensions import Any, Generic, TypeVar
 
 from .case import Case
@@ -21,6 +25,7 @@ from .telemetry import get_tracer, serialize
 from .telemetry._cloudwatch_logger import _send_to_cloudwatch
 from .types.evaluation import EvaluationData
 from .types.evaluation_report import EvaluationReport
+from .utils import is_throttling_error
 
 InputT = TypeVar("InputT")
 OutputT = TypeVar("OutputT")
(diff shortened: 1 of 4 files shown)


src/strands_evals/experiment.py
```diff
@@ -24,6 +34,46 @@
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
+# Retry configuration for handling throttling
+MAX_RETRY_ATTEMPTS = 6
+INITIAL_RETRY_DELAY = 4
+MAX_RETRY_DELAY = 240  # 4 minutes
```
Unshure: We ran into an issue with hardcoded retry configuration in the python sdk that took us a long time to eventually fix. Can we make this configurable from the start?

src/strands_evals/experiment.py
```diff
@@ -24,6 +34,46 @@
...
+THROTTLING_ERROR_CODES = {
+    "ThrottlingException",
+    "TooManyRequestsException",
+    "RequestLimitExceeded",
+    "ServiceUnavailable",
+    "ProvisionedThroughputExceededException",
+}
+
+
+def _is_throttling_error(exception: BaseException) -> bool:
```
afarntrog: do we have a `utils.py` kind of module? Would be great to add this to there so it can be reused in other locations as well.

src/strands_evals/experiment.py
afarntrog: Are we going to make these configurable or can users configure it by just importing and setting it? Fyi, if so, we need to keep in mind that we can change these names or the values once we expose them for public use.

src/strands_evals/experiment.py
```diff
@@ -240,92 +318,166 @@ async def _worker(self, queue: asyncio.Queue, task: Callable, results: list):
...
-                        }
+
+                @retry(
+                    retry=retry_if_exception(_is_throttling_error),
+                    stop=stop_after_attempt(MAX_RETRY_ATTEMPTS),
+                    wait=wait_exponential(multiplier=INITIAL_RETRY_DELAY, max=MAX_RETRY_DELAY),
+                    reraise=True,
+                )
+                async def _run_task_with_retry(task=task, case=case):
+                    return await self._run_task_async(task, case)
```
afarntrog, line 293: Is there some amount of code duplication in the async and nonasync implementations that we can reduce via some refactoring?

poshinchen: I think I made the recording part as another private function already, we could have another refactoring if needed as a follow-up.

src/strands_evals/experiment.py
Unshure: +1

We ran into an issue with hardcoded retry configuration in the python sdk that took us a long time to eventually fix. Can we make this configurable from the start?

poshinchen: I think I can make a follow-up request to make it configurable.
The reason why I don't do it in the beginning is that there's no such configuration at this moment. And we need to think about whether to have the config in experiment level or global level.

src/strands_evals/experiment.py
poshinchen: Will create one.

src/strands_evals/experiment.py
afarntrog: If we can't make this configurable from the start and we plan on doing so in the future I would propose we make these "private" `_`

Issue #93, opened 2026-01-23
[FEATURE] Tool response simulation with LLM


ybdarrenwang:
### Problem Statement

Testing agents against production APIs can be costly, time-consuming, and risky. While static mocking provides basic isolation from external dependencies, it has significant limitations:
- Brittle and Hard to Scale: Static mocks require manual maintenance as APIs and use cases evolve
- Poor Adaptability: Fails to detect new use cases or edge cases that weren't explicitly mocked
- Limited Coverage: Cannot simulate the diversity of real-world API responses
- Unrealistic Testing: Predefined responses don't reflect the dynamic nature of actual agent-tool interactions

LLM-backed tool response simulation enables more realistic end-to-end testing by generating dynamic, context-aware responses — while still avoiding external dependencies, production data exposure, and PII concerns. This approach provides comprehensive testing coverage without the brittleness of static mocks or the risks of hitting real tools.

### Proposed Solution

Implement LLM-backed tool response simulation that integrates seamlessly with the Strands Evals SDK workflow. The solution should support:
1. Stateful tools like tables and databases with shared state management across multiple tool calls
2. Integration with existing Strands Evals SDK workflow including Experiment, Case, and evaluators
3. Configurable tool errors including failure rates and error types
4. Compatibility with static mocking for simple scenarios where dynamic simulation is unnecessary

Example:
```
from strands_evals import ToolSimulator, Case, Experiment
from strands_evals.evaluators import TrajectoryEvaluator

# Define shared state for inventory system
INVENTORY_INITIAL_STATE = """
The inventory system contains 5 products:
- PROD-001: Wireless Bluetooth Headphones (125 units in stock)
- PROD-002: Ergonomic Office Chair (45 units in stock)
- PROD-003: Stainless Steel Water Bottle (0 units - out of stock)
"""

# Decorate tools for simulation with shared state
@ToolSimulator.function_tool(
    initial_state_description=INVENTORY_INITIAL_STATE,
    share_state_id="inventory_system"
)
def get_product_inventory(product_id: str) -> Dict[str, Any]:
    """Get inventory information for a specific product."""
    pass

@ToolSimulator.mcp_tool(
    schema=inventory_update_mcp_schema,
    share_state_id="inventory_system",
    initial_state_description=INVENTORY_INITIAL_STATE
)
def update_inventory(product_id: str, quantity_change: int, reason: str):
    """Update inventory levels via MCP protocol."""
    pass

@ToolSimulator.api_tool(
[...]

PR #111, opened 2026-01-29
feat: add ToolSimulator for tool response simulation

ybdarrenwang:
## Description
Introduces ToolSimulator framework for simulating realistic tool responses during agent evaluation without calling production APIs. Enables systematic testing of agents with function tools through LLM-powered dynamic simulation.

Key capabilities:
- Dynamic (LLM-powered) simulated response generation
- Shared state management across multiple tools via share_state_id for stateful testing scenarios
- Decorator-based registration for function tools
- Integration with Strands Evals workflow including Experiment, Case, and evaluators
- Multi-agent support with tool simulation across sub-agents (agent-as-tool pattern)

Design principles:
- Centralized registry for tool management and state tracking
- Context-aware response generation using initial state descriptions and conversation history
- Seamless integration with existing Strands tool decorator patterns
- Comprehensive unit test coverage for all simulation modes

## Related Issues
#93 

## Documentation PR
https://github.com/strands-agents/docs/pull/500

## Type of Change
New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/simulation/tool_simulator.py
+++ b/src/strands_evals/simulation/tool_simulator.py
@@ -182,17 +182,6 @@ class ToolSimulator:
         self.model = model
         self.state_registry = state_registry or StateRegistry(max_tool_call_cache_size=max_tool_call_cache_size)
         self._registered_tools: dict[str, RegisteredTool] = {}
-        self._initialize_shared_states()
-
-    def _initialize_shared_states(self):
-        """Initialize shared states from registered tools' initial descriptions."""
-        for tool_name, registered_tool in self._registered_tools.items():
-            if registered_tool.initial_state_description:
-                state_key = registered_tool.share_state_id or registered_tool.name
-                self.state_registry.initialize_state_via_description(
-                    registered_tool.initial_state_description, state_key
-                )
-                logger.info(f"Initialized state for tool '{tool_name}' with key '{state_key}'")
 
     def _create_tool_wrapper(self, registered_tool: RegisteredTool):
         """


src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,832 @@
+import inspect
+import json
+import logging
+import warnings
+from datetime import datetime
+from typing import Any, Callable, Dict, List, Optional
```
poshinchen: Can you use Python built-in collection types?

ybdarrenwang: Done in the latest commit

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,852 @@
...
+            framework: Agent framework to use (default: "strands")
+            max_tool_call_cache_size: Maximum number of tool calls to store per state key.
+                                     Only used when creating a new StateRegistry (ignored if state_registry
+                                     is provided). Older calls are automatically evicted when limit is exceeded.
+                                     Default is 20.
+        """
+        # Store framework selection
+        self.framework = framework
+        # Store model configuration for creating internal agents
+        self.model_id = model
```
poshinchen: nit: we could keep it as `model` instead of `model_id`.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,852 @@
...
+
+        # Set custom prompts or use defaults
+        if function_tool_prompt is None:
```
poshinchen: I'm fine with static loading so the code can be cleaner.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,852 @@
...
+
+            self.api_tool_prompt = API_TOOL_RESPONSE_GENERATION_PROMPT
+        else:
+            self.api_tool_prompt = api_tool_prompt
+
+        # Set up state registry
+        if state_registry:
+            self._state_registry = state_registry
+        elif self._state_registry is None:
+            self._state_registry = StateRegistry(max_tool_call_cache_size=max_tool_call_cache_size)
```
poshinchen: It should be the same as :
```
self._state_registry = state_registry or StateRegistry(max_tool_call_cache_size=max_tool_call_cache_size)
```
If self._state_registry isn't initialized anywhere before.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: Addressed in the latest commit.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: Addressed in the latest commit.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: Addressed in the latest commit.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+
+    def _simulate_tool_call(self, tool_type: ToolType, state_key: str, input_data: Dict[str, Any]) -> Any:
+        """Simulate a tool invocation and return the response."""
+        tool_name = input_data.get("tool_name", "")
+        registered_tool = self._registered_tools.get(tool_name)
+        if not registered_tool:
+            raise ValueError(f"Tool '{tool_name}' not registered")
+
+        # Handle different simulation modes
+        if registered_tool.mode == "static":
```
poshinchen: Since we support 3.10+, let's use `match` instead?

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+
+            # Create agent and generate response with structured output
+            agent = Agent(
+                system_prompt=self.mcp_tool_prompt,
+                tools=[],
+                model=self.model,
+                callback_handler=None,
+            )
+            result = agent(prompt, structured_output_model=MCPToolResponse)
+            if result.structured_output:
```
poshinchen: Is this check required? And is the dump required?

Strands-agent should handle this logic: https://github.com/strands-agents/sdk-python/blob/7db79bbeb53847006c1b6caad84dc6862e836477/src/strands/agent/agent_result.py#L52-L53

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+                system_prompt=self.function_tool_prompt,
+                tools=[],
+                model=self.model,
+                callback_handler=None,
+            )
+            result = agent(prompt, structured_output_model=None)
+
+            # Parse JSON response for function tools since they vary based on function signature
+            response_text = (
+                getattr(result, "response", None) or str(result.content) if hasattr(result, "content") else str(result)
```
poshinchen: The class of the result should be AgentResult, which you can check message.content? It should be possible to simplify this logic a bit.

https://github.com/strands-agents/sdk-python/blob/7db79bbeb53847006c1b6caad84dc6862e836477/src/strands/agent/agent_result.py#L32

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+            initial_state_description = current_state.get("initial_state", "No initial state provided.")
+
+            prompt = self.function_tool_prompt.format(
+                tool_name=tool_name,
+                parameters=json.dumps(parameters, indent=2) if parameters else "{}",
+                initial_state_description=initial_state_description,
+                previous_responses=json.dumps(current_state, indent=2) or "{}",
+            )
+
+            # Create agent and generate response with structured output
```
poshinchen: > Create agent and generate response with structured output
It says structured_output but below it sets structured_output_model=None. Is it expected? Probably this is the reason why you need json loads below?

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+            raise RuntimeError("State registry is not initialized")
+
+        try:
+            # Get initial state description from state registry
+            current_state = self._state_registry.get_state(state_key)
+            initial_state_description = current_state.get("initial_state", "No initial state provided.")
+
+            prompt = self.function_tool_prompt.format(
+                tool_name=tool_name,
+                parameters=json.dumps(parameters, indent=2) if parameters else "{}",
```
poshinchen: `if parameters else "{}",` seems to be redundant.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+                initial_state_description=initial_state_description,
+                previous_responses=json.dumps(current_state, indent=2) or "{}",
```
poshinchen: `or "{}"` seems to be redundant

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+            raise RuntimeError("State registry is not initialized")
+
+        try:
+            # Get initial state description from state registry
+            current_state = self._state_registry.get_state(state_key)
+            initial_state_description = current_state.get("initial_state", "No initial state provided.")
+
+            prompt = self.mcp_tool_prompt.format(
+                tool_name=tool_name,
+                mcp_payload=json.dumps(input_mcp_payload, indent=2) if input_mcp_payload else "{}",
```
poshinchen: `if input_mcp_payload else "{}",` is redundant.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+                initial_state_description=initial_state_description,
+                previous_responses=json.dumps(current_state, indent=2) or "{}",
```
poshinchen: same here

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+                tool_name, state_key, ToolType.MCP, response_data, input_mcp_payload=input_mcp_payload
+            )
+
+            return response_data
+
+        except Exception as e:
+            logger.error(f"Error generating MCP response: {e}")
+            return {"isError": True, "content": [{"type": "text", "text": f"Error generating response: {str(e)}"}]}
+
+    def _handle_api_tool(self, input_data: Dict[str, Any], state_key: str) -> Dict[str, Any]:
```
poshinchen: Same comments here from previous functions

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: For function tool we don't have specific data model, unlike MCP and API tools. That's why for function tools we simply attempt to parse the text response into json dict. (I tried dynamically create base class based on python function signature, but the codes seems long and error prone to me.)

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: The logic at agent_result.py#L52-L53 seems to handle cases when the agent could return structured output or NL output (and prioritize the former if present). I think it's different from our case cuz for Tool Simulator we mandate the response to follow the predetermined format, otherwise it's seen as a failed simulation.

But you're right that the check and dump can be simplified given agent_result's support.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+    Attributes:
+        model: Provider for running inference or model identifier for Bedrock.
+        _registered_tools: Class-level registry for all registered tools.
+        _state_registry: Registry for maintaining tool state across calls.
+    """
+
+    # Class-level registry for all registered tools
+    _registered_tools: dict[str, RegisteredTool] = {}
+    _state_registry: StateRegistry | None = None
+    _global_instance: Union["ToolSimulator", None] = None
```
poshinchen: "ToolSimulator | None" = None

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+                    case ToolType.API:
+                        return self._handle_api_tool(input_data, state_key)
+                    case _:
+                        raise ValueError(f"Tool type '{tool_type}' not supported")
+            case _:
+                raise ValueError(f"Tool simulation mode '{registered_tool.mode}' not supported")
+
+    def _handle_function_tool(self, input_data: dict[str, Any], state_key: str) -> dict[str, Any]:
+        """Handle function tool simulation."""
+        tool_name = input_data.get("tool_name", "")
```
poshinchen: I see that the `_handle_xx_tool` functions have duplicate code, is it possible to at least extract the beginning checks to a private function?

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,838 @@
...
+                callback_handler=None,
+            )
+            result = agent(prompt, structured_output_model=None)
+
+            # Parse JSON response for function tools since they vary based on function signature
+            response_text = str(result) or "No response"
+            try:
+                response_data = json.loads(response_text)
+            except json.JSONDecodeError:
+                response_data = {"result": response_text}
```
poshinchen: Same for these json.loads' try except duplicates, let's create an private function _parse_json_response?

poshinchen: hmm, Strands handles the serialization already. It should work directly as `response_data = { "result": response_text}`.

src/strands_evals/types/simulation/tool.py
```diff
@@ -0,0 +1,95 @@
...
+    mcp_schema: dict[str, Any] | None = Field(default=None, description="MCP tool schema")
+    api_path: str | None = Field(default=None, description="API endpoint path")
+    api_method: str | None = Field(default=None, description="HTTP method")
+    initial_state_description: str | None = Field(
+        default=None, description="Initial state description for the tool's context"
+    )
+    simulator_kwargs: dict[str, Any] | None = Field(
+        default_factory=dict, description="Additional simulator configuration"
+    )
+    mode: str = Field(default="dynamic", description="Simulation mode: dynamic, static, mock")
```
poshinchen: Should we do `Literal["dynamic", "static", "mock"] = Field(default="dynamic", description="Simulation mode: dynamic, static, mock")`? then you don't need to do another `match _:` for ValueError.

src/strands_evals/types/simulation/tool.py
```diff
@@ -0,0 +1,95 @@
...
+        initial_state_description: Initial state description for the tool's context.
+        simulator_kwargs: Additional simulator configuration parameters.
+    """
+
+    name: str = Field(..., description="Name of the tool")
+    tool_type: ToolType = Field(..., description="Type of the tool")
+    function: Callable | None = Field(default=None, description="Function callable", exclude=True)
+    mcp_schema: dict[str, Any] | None = Field(default=None, description="MCP tool schema")
+    api_path: str | None = Field(default=None, description="API endpoint path")
+    api_method: str | None = Field(default=None, description="HTTP method")
```
poshinchen: curious, what happens if user passes "get", "post", ... instead of GET / POST.

src/strands_evals/types/simulation/tool.py
ybdarrenwang: How about a `class ToolSimulationMode(Enum)`?

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: The intended behavior here is: The tool simulation agent should return a json string that's `json.loads`-able. But looking at Strands Agent's str() parsing:
```
        if not result and self.structured_output:
            result = self.structured_output.model_dump_json()

        return result
``` 
, I think there're 3 possibilities:
1. result == "" -- in such case, yes `response_data = { "result": response_text}` works, telling the agent that the response is empty.
2. result != "", and self.structured_output does not exist -- in such case, yes `response_data = { "result": response_text}` works, telling the agent that the response is non-empty but also non-json (so we return raw text instead).
3. result != "", and self.structured_output exists -- in such case, we want `response_data` to be the json dict loaded from response_text, without the additional layer of `{ "result":`. That's why the try part is needed.

src/strands_evals/types/simulation/tool.py
ybdarrenwang: One potential risk is the consistency of the stored previous calls -- switching between upper and lower cases may or may not confuse backbone LLM. That shouldn't break the simulation; but I can make them all uppercase just for consistency.

src/strands_evals/types/simulation/tool.py
poshinchen: Sure!

src/strands_evals/types/simulation/tool.py
```diff
@@ -0,0 +1,112 @@
+from enum import Enum
+from typing import Any, Callable
+
+from pydantic import BaseModel, Field
+
+
+class ToolType(Enum):
```
Unshure: Why do we need these different tool types? Why isnt everything just a FUNCTION tool?

cc: @poshinchen

src/strands_evals/types/simulation/tool.py
```diff
@@ -0,0 +1,112 @@
...
+        MCP: Model Context Protocol tools with structured schemas.
+        API: REST API endpoints with HTTP methods and paths.
+    """
+
+    FUNCTION = "function"
+    MCP = "mcp"
+    API = "api"
+
+
+class ToolSimulationMode(Enum):
```
Unshure: Chatted offline with @poshinchen about this one as well, confused why we need STATIC and MOCK simulation modes? Dynamic makes sense, but im missing the value in having STATIC and MOCK

poshinchen: It was the discussion before checking whether STATIC is useful, but apparently it can be done directly when creating the mock tool. So we decided to remove MOCK and STATIC and keep the Dynamic.

src/strands_evals/types/simulation/tool.py
ybdarrenwang: Despite eventually all tools would be wrapped as a function, to be attached to Strands agent say with @tool decorator, some specific tool types such as API (with OpenAPI schema), MCP (with MCP schema) etc. have very specific schema underneath. We tailored prompts and instructions specifically for them so LLM can simulate their responses reliably.

c.c. @vinayakarannil

ybdarrenwang: After discussion with @poshinchen, we've confirmed that the scope of this PR shall contain only function tool. We will validate the necessity of MCP and API specific prompts and schemas, and create a follow-up PR.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,444 @@
...
+            function_tool_prompt: Optional custom prompt for tool response generation
+            model: Provider for running inference or a string representing the model-id for Bedrock to use
+            framework: Agent framework to use (default: "strands")
+            max_tool_call_cache_size: Maximum number of tool calls to store per state key.
+                                     Only used when creating a new StateRegistry (ignored if state_registry
+                                     is provided). Older calls are automatically evicted when limit is exceeded.
+                                     Default is 20.
+        """
+        # Store framework selection
+        self.framework = framework
```
poshinchen: What's the plan /timeline of adding tool mock for different framework? If we're supporting strands in the near future, we don't need this.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,444 @@
...
+                           a new StateRegistry will be created with max_tool_call_cache_size.
+            function_tool_prompt: Optional custom prompt for tool response generation
+            model: Provider for running inference or a string representing the model-id for Bedrock to use
+            framework: Agent framework to use (default: "strands")
+            max_tool_call_cache_size: Maximum number of tool calls to store per state key.
+                                     Only used when creating a new StateRegistry (ignored if state_registry
+                                     is provided). Older calls are automatically evicted when limit is exceeded.
+                                     Default is 20.
+        """
+        # Store framework selection
```
poshinchen: Some of these comments are not useful we can just skip them.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,444 @@
...
+
+        # Use framework-specific method to create the tool wrapper
+        if self.framework == "strands":
+            return self._create_strands_tool_wrapper(registered_tool, wrapper)
+        else:
+            raise ValueError(f"Framework '{self.framework}' is not supported. Only 'strands' is currently supported.")
+
+    def _create_strands_tool_wrapper(self, registered_tool: RegisteredTool, wrapper: Callable):
+        """Create a Strands-specific DecoratedFunctionTool wrapper."""
+        from strands.tools.decorator import DecoratedFunctionTool, FunctionToolMetadata
```
poshinchen: could this be imported at the top?

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,444 @@
...
+            wrapper.__name__ = registered_tool.name
+            wrapper.__doc__ = f"Simulated {registered_tool.name} tool"
+
+        # Use framework-specific method to create the tool wrapper
+        if self.framework == "strands":
+            return self._create_strands_tool_wrapper(registered_tool, wrapper)
+        else:
+            raise ValueError(f"Framework '{self.framework}' is not supported. Only 'strands' is currently supported.")
+
+    def _create_strands_tool_wrapper(self, registered_tool: RegisteredTool, wrapper: Callable):
```
poshinchen: Can we force user to do this:

```
@simulator.tool(....)
@tool
def search_tool()dict
"""Tool definitions and inputSchema"""
pass
```

So we can use the  `@tool` decorator logics to load the tool?

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: removed from latest commit

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: removed from latest commit

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: done

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: Yes great suggestion. I've incorporate this into latest commit.

tests/strands_evals/simulation/test_tool_simulator.py
```diff
@@ -0,0 +1,419 @@
...
+
+def test_tool_decorator_registration():
+    """Test tool decorator registration."""
+    simulator = ToolSimulator()
+
+    @simulator.tool()
+    @tool
+    def test_function(x: int, y: str) -> dict:
+        """A sample function for testing."""
+        return {"x": x, "y": y}
```
Unshure: This feels a bit clunky to need both of these annotations to get the tool working. Can we have it so only `@simulator.tool` is needed?
```suggestion
    @simulator.tool
    def test_function(x: int, y: str) -> dict:
        """A sample function for testing."""
        return {"x": x, "y": y}
```

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,420 @@
...
+from strands_evals.types.simulation.tool import RegisteredTool
+
+from .prompt_templates.tool_response_generation import FUNCTION_TOOL_RESPONSE_GENERATION_PROMPT
+
+logger = logging.getLogger(__name__)
+
+
+class StateRegistry:
+    """
+    State registry for managing shared state between tool simulators.
```
Unshure, line 23: Curious what the use cases for shared state are? Why does one tool need to know the state of another? If tools need to share state, what are some other approaches we can take to doing this?

One that comes to mind is to provide the same agent to multiple tool simulators, so that same agent has the context of all of the invocations, and we dont need this `StateRegistry` class

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,420 @@
...
+        Creates an empty state dictionary to track tool calls and responses
+        across different simulation sessions. Tool call cache is automatically
+        bounded to prevent excessive memory usage.
+
+        Args:
+            max_tool_call_cache_size: Maximum number of tool calls to store per state key.
+                                      Older calls are automatically evicted when limit is exceeded.
+                                      Default is 20.
+        """
+        self.max_tool_call_cache_size = max_tool_call_cache_size
```
Unshure, line 46: Do we need a cache? Im worried we are overengineering this feature before we see how folks will use it

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,420 @@
...
+
+class ToolSimulator:
+    """
+    Simulates tool behavior with decorator-based registration system for agent evaluation.
+
+    ToolSimulator provides a decorator for tools and maintains a registry of all
+    registered tools. It can be configured to override tool behavior for simulation purposes,
+    enabling controlled testing scenarios.
+
+    IMPORTANT: This simulator expects functions to be decorated with Strands' @tool decorator first.
```
Unshure: Why do we need this? Seems like its a bit clunky of a devex

tests/strands_evals/simulation/test_tool_simulator.py
ybdarrenwang: Stacking `@simulator.tool()` on `@tool` was suggested by @poshinchen. The benefit is that `@simulator.tool()` can reuse function metadata already extracted by `@tool`, thus shorter implementation code.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: A minimal use case that needs this is when a tool is database/table getter and the other setter. The getter needs to know what setter has done in previous calls so it fetches the latest data.

We do use Strands agent for LLM calls underneath. So reusing its memory is actually a good suggestion.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: Stacking `@simulator.tool()` on `@tool` was suggested by @poshinchen. The benefit is that `@simulator.tool()` can reuse function metadata already extracted by `@tool`, thus shorter implementation code.

tests/strands_evals/simulation/test_tool_simulator.py
ybdarrenwang: See https://github.com/strands-agents/evals/pull/111#discussion_r2790160046

ybdarrenwang: Chatted with @poshinchen offline and resolved the confusion. The decorator stacking has been removed in latest commit.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: Chatted with @poshinchen offline and resolved the confusion. The decorator stacking has been removed in latest commit.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: We can use a simple array and just keep appending, removing the limit of `max_tool_call_cache_size` and leave it to future.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: After discussion offline, we decided to keep the StateRegistry as of now, and we can have a follow-up action to move to strands memory later on.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,418 @@
...
+            model: Provider for running inference or a string representing the model-id for Bedrock to use
+            max_tool_call_cache_size: Maximum number of tool calls to store per state key.
+                                     Only used when creating a new StateRegistry (ignored if state_registry
+                                     is provided). Older calls are automatically evicted when limit is exceeded.
+                                     Default is 20.
+        """
+        self.model = model
+        self.state_registry = state_registry or StateRegistry(max_tool_call_cache_size=max_tool_call_cache_size)
+        self._registered_tools: dict[str, RegisteredTool] = {}
+        self._initialize_shared_states()
```
poshinchen: Given the current setup / example, will `_initialize_shared_states` do anything?

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -0,0 +1,418 @@
...
+        response_data = self._parse_simulated_response(result)
+
+        self.state_registry.cache_tool_call(
+            registered_tool.name, state_key, response_data, parameters=json.loads(parameters_string)
+        )
+        return response_data
+
+    def tool(
+        self,
+        output_schema: type[BaseModel],
```
poshinchen: hmm, can it be optional, and default to a schema?

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: It calls StateRegistry's `initialize_state_via_description`, stores the lliteral `initial_state_description` and create the previous call cache.

src/strands_evals/simulation/tool_simulator.py
ybdarrenwang: Based on our tests, the output schema must be provided to the simulation prompt; otherwise the returned simulated tool response can be arbitrary (even empty).

ybdarrenwang: I tested with more examples and found it possible to provide a bare bone output schema with limited impact to the simulated response quality. Checked in with the latest commit.

src/strands_evals/simulation/tool_simulator.py
poshinchen: only when `_registered_tools` is not empty, but it is always empty when you initialize the class regardless of the state_registry?

ybdarrenwang: ok I see what you mean. That's right. I've removed `_initialize_shared_states` in latest commit.

Issue #101, opened 2026-01-26
[FEATURE] Implement CoherenceEvaluator


poshinchen:
### Problem Statement

Integrate CoherenceEvaluator from agentcore to Strands-evals

### Use Case

N/A

PR #125, opened 2026-02-09
Added CoherenceEvaluator

poshinchen:
## Description
Added Coherence evaluator which evaluates the logical cohesion and reasoning quality of agent responses with five-level scoring:

- NOT_AT_ALL = "Not At All"
- NOT_GENERALLY = "Not Generally"
- NEUTRAL_MIXED = "Neutral/Mixed"
- GENERALLY_YES = "Generally Yes"
- COMPLETELY_YES = "Completely Yes"

The evaluator is trace_level, which only contains the last prompt and the response.

## Related Issues

https://github.com/strands-agents/evals/issues/101

## Documentation PR

TBD

## Type of Change
New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/evaluators/coherence_evaluator.py
+++ b/src/strands_evals/evaluators/coherence_evaluator.py
@@ -1,0 +1,152 @@
+from enum import Enum
+from typing import cast
+
+from pydantic import BaseModel, Field
+from strands import Agent
+from strands.models.model import Model
+from typing_extensions import Union
+
+from ..types.evaluation import EvaluationData, EvaluationOutput, InputT, OutputT
+from ..types.trace import EvaluationLevel, TextContent, ToolExecution, TraceLevelInput
+from .evaluator import Evaluator
+from .prompt_templates.coherence import get_template
+
+
+class CoherenceScore(str, Enum):
+    """Categorical coherence ratings."""
+
+    NOT_AT_ALL = "Not At All"
+    NOT_GENERALLY = "Not Generally"
+    NEUTRAL_MIXED = "Neutral/Mixed"
+    GENERALLY_YES = "Generally Yes"
+    COMPLETELY_YES = "Completely Yes"
+
+
+class CoherenceRating(BaseModel):
+    """Structured output for coherence evaluation."""
+
+    reasoning: str = Field(description="step by step reasoning to derive the final score, using no more than 250 words")
+    score: CoherenceScore = Field(description="Categorical coherence rating")
+
+
+class CoherenceEvaluator(Evaluator[InputT, OutputT]):
+    """Evaluates the logical cohesion of the assistant's response.
+    
+    This evaluator assesses whether the assistant's response maintains logical consistency,
+    flows naturally, and presents ideas in a well-organized manner. It uses an LLM-as-judge
+    approach to provide categorical ratings that are then normalized to numeric scores.
+    
+    Scores:
+    - NOT_AT_ALL (0.0): Response is completely incoherent or contradictory
+    - NOT_GENERALLY (0.25): Response has significant logical gaps or inconsistencies
+    - NEUTRAL_MIXED (0.5): Response has both coherent and incoherent elements
+    - GENERALLY_YES (0.75): Response is mostly coherent with minor reasoning issues
+    - COMPLETELY_YES (1.0): Response is fully coherent and logically consistent
+    """
+
+    evaluation_level = EvaluationLevel.TRACE_LEVEL
+
+    _score_mapping = {
+        CoherenceScore.NOT_AT_ALL: 0.0,
+        CoherenceScore.NOT_GENERALLY: 0.25,
+        CoherenceScore.NEUTRAL_MIXED: 0.5,
+        CoherenceScore.GENERALLY_YES: 0.75,
+        CoherenceScore.COMPLETELY_YES: 1.0,
+    }
+
+    def __init__(
+        self,
+        version: str = "v0",
+        model: Union[Model, str, None] = None,
+        system_prompt: str | None = None,
+        include_inputs: bool = True,
+    ):
+        super().__init__()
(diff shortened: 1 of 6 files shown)


src/strands_evals/evaluators/coherence_evaluator.py
```diff
@@ -0,0 +1,140 @@
...
+
+    NOT_AT_ALL = "Not At All"
+    NOT_GENERALLY = "Not Generally"
+    NEUTRAL_MIXED = "Neutral/Mixed"
+    GENERALLY_YES = "Generally Yes"
+    COMPLETELY_YES = "Completely Yes"
+
+
+class CoherenceRating(BaseModel):
+    """Structured output for coherence evaluation."""
```
dbschmigelski, line 26: nitL it might be helpful to pass the description as a doc string here if it is included. If it is not, then inthe Field description might benefit too

```
- Not At All: Too many errors of reasoning, contradictions, or major gaps.
- Not Generally: A few instances of coherent reasoning, but errors reduce quality.
- Neutral/Mixed: Unclear whether the reasoning is correct or not.
- Generally Yes: Small reasoning issues, but the main point is well-argued.
- Completely Yes: No issues with logical cohesion. The reasoning is sound and consistent
- ```

File: src/strands_evals/evaluators/coherence_evaluator.py [hunk 2]
```diff
@@ -0,0 +1,140 @@
...
+        result = await evaluator_agent.invoke_async(prompt, structured_output_model=CoherenceRating)
+        return self._create_evaluation_output(result)
+
+    def _create_evaluation_output(self, result) -> list[EvaluationOutput]:
+        rating = cast(CoherenceRating, result.structured_output)
+        normalized_score = self._score_mapping[rating.score]
+        return [
+            EvaluationOutput(
+                score=normalized_score,
+                test_pass=normalized_score >= 0.5,
```
dbschmigelski, line 90: nit: this threshold should probably be configurable

src/strands_evals/evaluators/coherence_evaluator.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        parts = []
+
+        if parsed_input.session_history:
+            history_lines = []
+            for msg in parsed_input.session_history:
+                if isinstance(msg, list) and msg and isinstance(msg[0], ToolExecution):
+                    continue  # Skip tool execution lists
+                if not isinstance(msg, list) and self._has_text_content(msg):
+                    first_content = msg.content[0]
```
dbschmigelski, line 143: why only text and why are we only taking the first content block even if Text is the second?

src/strands_evals/evaluators/coherence_evaluator.py
```diff
@@ -0,0 +1,140 @@
...
+    reasoning: str = Field(description="step by step reasoning to derive the final score, using no more than 250 words")
+    score: CoherenceScore = Field(description="Categorical coherence rating")
+
+
+class CoherenceEvaluator(Evaluator[InputT, OutputT]):
+    """Evaluates the logical cohesion of the assistant's response."""
+
+    evaluation_level = EvaluationLevel.TRACE_LEVEL
+
+    _score_mapping = {
```
dbschmigelski, line 49: Does the agent perform better just doing a continuous scale of 0-1 and then we map compared to only giving these options?

src/strands_evals/evaluators/coherence_evaluator.py
```diff
@@ -0,0 +1,140 @@
...
+        CoherenceScore.NOT_AT_ALL: 0.0,
```
dbschmigelski, line 50: Are these level names pre agreed upon or up for debate? What is the performance of thgese versus 

low, low-medium, medium, medium-high, high or 1,2,3,4,5 for example

src/strands_evals/evaluators/coherence_evaluator.py
poshinchen: Yes, we'll have a follow-up task to refactor them

src/strands_evals/evaluators/coherence_evaluator.py
poshinchen: yes, this is more of like a scientific question I believe. Instead of telling people 0.8 / 0.9,  it would be more useful to take them which category it is.

src/strands_evals/evaluators/coherence_evaluator.py
poshinchen: This is following the Agentcore evaluators to make sure users can use them in both places. So I think it's pre-agreed before their launch!

src/strands_evals/evaluators/coherence_evaluator.py
dbschmigelski: Yeah I agree, but telling the user is different than the LLM telling us. What about both btw? Like we are effectively providing a description. Here is the meaning of the score, but does the LLM actually perform more reliably one way or another

src/strands_evals/evaluators/coherence_evaluator.py
dbschmigelski: Gotcha ok that makes sense then.

src/strands_evals/evaluators/coherence_evaluator.py
poshinchen: lol, we just got an issue open: https://github.com/strands-agents/evals/issues/130

src/strands_evals/evaluators/coherence_evaluator.py
poshinchen: > Here is the meaning of the score, but does the LLM actually perform more reliably one way or another

I think it is performing better when mentioning the scores mapping to the LLM. But i'll ask the team to double check!

src/strands_evals/evaluators/coherence_evaluator.py
dbschmigelski: ok as long as we handle it there

Issue #130, opened 2026-02-10
[BUG] eval functions _convert_agent_invocation_span and _process_user_message do not support multimodal input


goDylanMorgan:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands and evals
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

0.1.5

### Strands Evals Version

1.24.0

### Python Version

3.12.3

### Operating System

Ubuntu 24.04.3 LTS

### Installation Method

None

### Steps to Reproduce

Minimal Repro Script
```
from strands import Agent                                                                                                              
  from strands_evals import Case, Experiment          
  from strands_evals.evaluators import GoalSuccessRateEvaluator                                                                          
  from strands_evals.mappers import StrandsInMemorySessionMapper
  from strands_evals.telemetry import StrandsEvalsTelemetry                                                                              
                                                                         
  telemetry = StrandsEvalsTelemetry().setup_in_memory_exporter()
  memory_exporter = telemetry.in_memory_exporter

  def task_fn(case):
      memory_exporter.clear()
      agent = Agent(
          trace_attributes={"session.id": case.session_id, "gen_ai.conversation.id": case.session_id},
          callback_handler=None,
      )
      result = agent([
          {"document": {"name": "invoice", "format": "pdf", "source": {"bytes": open("invoice.pdf", "rb").read()}}},
          {"text": "Analyze this document"},
      ])
      mapper = StrandsInMemorySessionMapper()
      session = mapper.map_to_session(memory_exporter.get_finished_spans(), session_id=case.session_id)

      # Bug: session.traces[0].spans[0].user_prompt will be "" because
      # the mapper does content_list[0].get("text", "") and index 0 is the document block
      print(f"user_prompt: '{session.traces[0].spans[0].user_prompt}'")

      return {"output": str(result), "trajectory": session}

  experiment = Experiment(
      cases=[Case(input="Analyze this document")],
      evaluators=[GoalSuccessRateEvaluator()],
  )
  reports = experiment.run_evaluations(task=task_fn)
  reports[0].run_display()
```

### Expected Behavior

A multimodal (or any input prompt with multiple content blocks) should be analyzed by the evaluator.  Absent that behavior, common enterprise use cases like agentic document processing cannot be tested effectively because the evaluators don't see what the model saw so they can't effectively evaluate on Faithfulness, Goal Success etc...

[...]

PR #133, opened 2026-02-11
fix: retrieve multiple text contentBlock in messageConent

poshinchen:
## Description
Our sessionMapper only reads and parse the first text contentBlock, but there are cases where multiple multiple text contentBlocks exists. Therefore adjust the parsing logic to detect multiple text contentBlocks.

`Note`:
multi-modal contentBlcoks parsing is out of scope as multi-modal support will be visited in the future.

## Related Issues

https://github.com/strands-agents/evals/issues/130

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/evaluators/evaluator.py
+++ b/src/strands_evals/evaluators/evaluator.py
@@ -131,14 +131,37 @@ class Evaluator(Generic[InputT, OutputT]):
             msg: Message object to check (UserMessage or AssistantMessage)
 
         Returns:
-            True if msg has content attribute with at least one item that is TextContent
+            True if msg has content attribute with at least one TextContent block.
+            Note: TextContent may not be at index 0 due to tool calls or other content types.
         """
-        return (
-            hasattr(msg, "content")
-            and bool(msg.content)
-            and len(msg.content) > 0
-            and isinstance(msg.content[0], TextContent)
-        )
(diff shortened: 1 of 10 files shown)


src/strands_evals/evaluators/evaluator.py
```diff
@@ -131,14 +131,37 @@ def _has_text_content(self, msg: UserMessage | AssistantMessage) -> TypeGuard[Us
...
-            and len(msg.content) > 0
-            and isinstance(msg.content[0], TextContent)
-        )
+        if not hasattr(msg, "content") or not msg.content:
+            return False
+
+        # Check if ANY content block is TextContent, not just the first
+        return any(isinstance(content_block, TextContent) for content_block in msg.content)
+
+    def _extract_text_content(self, msg: UserMessage | AssistantMessage) -> str:
```
mkmeral, line 143: is this the best devX? if I am creating custom evaluators, am I supposed to use this? 

If so, why is it private? 

If not, why don't we create utils to allow everyone to use these methods? Because we would be forcing them to replicate the logic we have

poshinchen: This is private function which extract the data by traces automatically. There's another pr to extract everything as util.

PR #137, opened 2026-02-12
fix: handle parallel tool calls during tool extraction

clareliguori:
## Description
The previous tool use extraction logic assumed a single tool use / result per turn, and silently dropped any parallel tool calls.

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/extractors/tools_use_extractor.py
+++ b/src/strands_evals/extractors/tools_use_extractor.py
@@ -21,51 +21,46 @@ def extract_agent_tools_used_from_messages(agent_messages):
     for i, message in enumerate(agent_messages):
         if message.get("role") == "assistant":
             message_info = message.get("content")
-            if len(message_info) > 0:
-                tools = []
-                for message in message_info:
-                    if "toolUse" in message:
-                        tools.append(message.get("toolUse"))
+            if message_info:
+                # Collect tool uses from this message
+                tools = [cb.get("toolUse") for cb in message_info if cb.get("toolUse")]
+                if not tools:
+                    continue
+
+                # Build lookup dict of tool results from subsequent user messages
+                tool_ids_needed = {tool.get("toolUseId") for tool in tools}
+                tool_results_by_id: dict[str, dict] = {}
+                for next_message in agent_messages[i + 1 :]:
+                    if next_message.get("role") == "user":
+                        for content_block in next_message.get("content") or []:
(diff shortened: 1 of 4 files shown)


src/strands_evals/extractors/tools_use_extractor.py
```diff
@@ -43,24 +43,21 @@ def extract_agent_tools_used_from_messages(agent_messages):
                             if next_message.get("role") == "user":
                                 content = next_message.get("content")
                                 if content:
-                                    # Find toolResult in content blocks - may not be at index 0
-                                    tool_result_dict = None
+                                    # Find toolResult with matching toolUseId in content blocks
                                     for content_block in content:
                                         if "toolResult" in content_block:
                                             tool_result_dict = content_block.get("toolResult")
```
JackYPCOnline: nit : this tool_result_dict is pointing to last non-matching `toolResult` if not found. This is not a functional bug.

src/strands_evals/extractors/tools_use_extractor.py
JackYPCOnline: nit: For each tool in a parallel batch, the while loop rescans forward from i + 1 to find the matching toolResult. Since all parallel tool results are typically in the same next user message, we're scanning the same message N times (once per tool).

clareliguori: Ooh good point, I didn't look at this closely enough. And actually we don't need to scan forward, we can just build a dict of tool results by ID once.

clareliguori: Oh wait yes we do need to scan forward, some models do re-use tool use IDs

src/strands_evals/extractors/tools_use_extractor.py
```diff
@@ -21,51 +21,46 @@ def extract_agent_tools_used_from_messages(agent_messages):
...
-                        tools.append(message.get("toolUse"))
+            if message_info:
+                # Collect tool uses from this message
+                tools = [cb.get("toolUse") for cb in message_info if cb.get("toolUse")]
+                if not tools:
+                    continue
+
+                # Build lookup dict of tool results from subsequent user messages
+                tool_ids_needed = {tool.get("toolUseId") for tool in tools}
+                tool_results_by_id: dict[str, dict] = {}
```
JackYPCOnline, line 32: Good Optimization!

Issue #97, opened 2026-01-26
[FEATURE] Create TraceClient Interface with interface definition


poshinchen:
### Problem Statement

Design and implement a TraceClient interface to retrieve traces and will be further used for task execution.

### Proposed Solution

Possible APIs
- get_session(session_id) - Retrieve complete session
- get_trace(trace_id) - Retrieve single trace
- list_sessions(filters) - List with filtering
- Async variants for all methods (TBD)

### Use Case

Users can create its own client which inherits the base client.

PR #140, opened 2026-02-16
feat: trace provider interface

afarntrog:
## Description
Add TraceProvider interface for observability backends

Add abstract TraceProvider that retrieves agent trace data from
observability backends and returns Session/Trace types the evals
system already consumes.

- TraceProvider ABC with get_session() (required), list_sessions()
  and get_session_by_trace_id() (optional, raise NotImplementedError)
- SessionFilter dataclass for time-range and limit-based discovery
- Exception hierarchy: TraceProviderError base with
  SessionNotFoundError, TraceNotFoundError, ProviderError
- Export providers module from strands_evals package

## Related Issues

https://github.com/strands-agents/evals/issues/97

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/providers/trace_provider.py
+++ b/src/strands_evals/providers/trace_provider.py
@@ -1,26 +1,8 @@
 """TraceProvider interface for retrieving agent trace data from observability backends."""
 
 from abc import ABC, abstractmethod
-from collections.abc import Iterator
-from dataclasses import dataclass, field
-from datetime import datetime
-from typing import Any
 
 from ..types.evaluation import TaskOutput
-
-
-@dataclass
-class SessionFilter:
-    """Filter criteria for discovering sessions.
-
-    Universal fields are defined here. Provider-specific parameters
-    go in `additional_fields`.
-    """
-
-    start_time: datetime | None = None
-    end_time: datetime | None = None
-    limit: int | None = None
-    additional_fields: dict[str, Any] = field(default_factory=dict)
 
 
 class TraceProvider(ABC):
@@ -50,49 +32,3 @@ class TraceProvider(ABC):
             ProviderError: If the provider is unreachable or returns an error
         """
         ...
-
-    def list_sessions(
-        self,
-        session_filter: SessionFilter | None = None,
-    ) -> Iterator[str]:
-        """Discover session IDs matching filter criteria.
-
-        Returns session IDs that can be fed to get_evaluation_data().
-        Not abstract — providers override to enable session discovery.
-
-        Args:
-            session_filter: Optional filter. If None, provider-specific defaults apply.
-
-        Yields:
-            Session ID strings
-
-        Raises:
-            NotImplementedError: If the provider does not support session discovery
-            ProviderError: If the provider is unreachable or returns an error
-        """
-        raise NotImplementedError(
-            "This provider does not support session discovery. "
-            "Use get_evaluation_data() with a known session_id instead."
-        )
-
-    def get_evaluation_data_by_trace_id(self, trace_id: str) -> TaskOutput:
-        """Fetch a single trace and return its evaluation data.
-
-        Useful when someone has a trace_id (e.g., from a Langfuse link) but
-        not a session_id, or for single-shot agent runs without sessions.
-        Not abstract — providers override to enable trace-level retrieval.
-
-        Args:
-            trace_id: The unique trace identifier
-
-        Returns:
-            TaskOutput with 'output' and 'trajectory' for the single trace
-
-        Raises:
-            NotImplementedError: If the provider does not support trace-level retrieval
-            TraceNotFoundError: If trace doesn't exist
-            ProviderError: If the provider is unreachable or returns an error
-        """
-        raise NotImplementedError(
-            "This provider does not support trace-level retrieval. Use get_evaluation_data() with a session_id instead."
-        )
--- a/src/strands_evals/providers/__init__.py
+++ b/src/strands_evals/providers/__init__.py
@@ -5,13 +5,11 @@ from .exceptions import (
     TraceProviderError,
 )
 from .trace_provider import (
-    SessionFilter,
     TraceProvider,
 )
 
 __all__ = [
     "ProviderError",
-    "SessionFilter",
     "SessionNotFoundError",
     "TraceNotFoundError",
     "TraceProvider",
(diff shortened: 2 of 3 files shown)


src/strands_evals/providers/trace_provider.py
```diff
@@ -0,0 +1,98 @@
...
+            TaskOutput with 'output' (final agent response) and
+            'trajectory' (Session containing all traces/spans)
+
+        Raises:
+            SessionNotFoundError: If no traces found for session_id
+            ProviderError: If the provider is unreachable or returns an error
+        """
+        ...
+
+    def list_sessions(
```
JackYPCOnline: nit: list_sessionIds?

src/strands_evals/providers/trace_provider.py
```diff
@@ -0,0 +1,98 @@
...
+        Raises:
+            NotImplementedError: If the provider does not support session discovery
+            ProviderError: If the provider is unreachable or returns an error
+        """
+        raise NotImplementedError(
+            "This provider does not support session discovery. "
+            "Use get_evaluation_data() with a known session_id instead."
+        )
+
+    def get_evaluation_data_by_trace_id(self, trace_id: str) -> TaskOutput:
```
JackYPCOnline: Any example of how you call this function? if we call this under trace namespace, may be `get_evaluation_data_by_id` is good enough?

src/strands_evals/providers/trace_provider.py
```diff
@@ -0,0 +1,98 @@
...
+
+class TraceProvider(ABC):
+    """Retrieves agent trace data from observability backends for evaluation.
+
+    Implementations handle authentication, pagination, and conversion from
+    provider-native formats to the types the evals system consumes.
+    """
+
+    @abstractmethod
+    def get_evaluation_data(self, session_id: str) -> TaskOutput:
```
JackYPCOnline, line 16: Also output will not be none right?

src/strands_evals/providers/trace_provider.py
afarntrog: I think it's worth keeping `list_sessions` similar to S3 that has `list_buckets`: https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html

src/strands_evals/providers/trace_provider.py
afarntrog: A Session contains one or more trace ids. So the idea over here is that you can provide a single trace id and we wrap it in a session object so that it can be passed as the `trajectory` as the `TaskOutput`. This can be useful if one wants to evaluate from a particular point in a session

src/strands_evals/providers/trace_provider.py
afarntrog: What do you mean? The output should always be a `TaskOutput`

JackYPCOnline: all output will not be like : Optional[TaskOutput] ? also for other functions

afarntrog: I see. So the current implementation is that if there is no trace data then we raise a `TraceNotFoundError` otherwise we return the `TaskOutput`. I would like some kind of case before we make it `Optional`.

src/strands_evals/providers/trace_provider.py
```diff
@@ -0,0 +1,98 @@
...
+    go in `additional_fields`.
+    """
+
+    start_time: datetime | None = None
+    end_time: datetime | None = None
+    limit: int | None = None
+    additional_fields: dict[str, Any] = field(default_factory=dict)
+
+
+class TraceProvider(ABC):
```
Unshure, line 8: Have you implemented this abstract class for any provider yet, or how this will be used in evaluations? Im missing how each of these methods are used, so its hard to give feedback on the shape of this class.

afarntrog: See here: https://github.com/afarntrog/evals/tree/trace_provider_cloudwatch/src/strands_evals/providers thanks!

src/strands_evals/providers/trace_provider.py
Unshure: Trying to understand the customer use case here. Will customers write code that looks like this?
```
from strands_evals.evaluators import CoherenceEvaluator, OutputEvaluator
from strands_evals.providers import CloudWatchProvider

provider = CloudWatchProvider(log_group="/aws/...", region="us-east-1")

def task(case: Case) -> dict:
    return provider.get_evaluation_data(case.input)

cases = [Case(name="session_1", input="my-session-id", expected_output="any")]
evaluators = [
    OutputEvaluator(rubric="Score 1.0 if the output is coherent. Score 0.0 otherwise."),
    CoherenceEvaluator(),
]

experiment = Experiment(cases=cases, evaluators=evaluators)
reports = experiment.run_evaluations(task)

for report in reports:
    print(f"{report.overall_score:.2f} - {report.reasons}")
```

In this case, we only need `get_evaluation_data` as a required method, right?

src/strands_evals/providers/trace_provider.py
Unshure: When using this in evals, when do customers need to list sessions? Are sessions present in every trace provider? Or is trace_id sufficient?

src/strands_evals/providers/trace_provider.py
afarntrog: Yup! Only the `get_evaluation_data` method in the `TraceProvider` is required. Customers can optionally implement the other. Here's an example of where we implement it. https://github.com/afarntrog/evals/blob/trace_provider_cloudwatch/src/strands_evals/providers/langfuse_provider.py#L124-L139

src/strands_evals/providers/trace_provider.py
afarntrog: I don't know about 'every' trace provider, however, I know at least Langfuse and Cloudwatch both have sessions. I imagine a customer flow that looks like this:
```python
# Batch eval run:
Run traces on all traces within the last 24hrs 
session_ids = list_sessions(...)
for sid in session_ids:
   run...
```

src/strands_evals/providers/trace_provider.py
Unshure: If they are optional, can we exclude them? Customers can add their own convenience methods if they need them. Worried customers will see these and get confused why they need them.

afarntrog: Discussed offline. https://github.com/strands-agents/evals/issues/143

Issue #109, opened 2026-01-29
[EPIC] Add Definitive Evaluation / assertions


poshinchen:
## Problem Statement

Strands Evals currently requires LLM invocations for ALL evaluations, even for simple assertions that should be deterministic. This feature request proposes adding definitive (deterministic, non-LLM) evaluation capabilities to complement existing LLM-based evaluators, enabling fast, reliable, and cost-free assertions for structural and factual checks.

### Basic
- Equals
- Contains
- MaxDuration / MinDuration
- HasMatchingSpan
- StartsWith
- IsInstance

### Span-level
- HasSpan(name) - Check if span with name exists
- HasAttributes
- HasSpanWithAttributes(name, attributes) - Span with specific attributes
- HasMatchingSpan(query) - Complex query-based check
- SpanSequence
- HassErrorFlag

### Tool-level
- ToolCalled
- ToolSucceeded
- ToolResultContains
- ToolResultEquals

### Generic
- Custom: definitive callback

...

### Solution 1: Add definitive matching method as part of evaluators
The implementation will look like:

```
evaluators = [
    # Definitive checks (fast, free, reliable)
    Equals(value=expected_output),
    HasSpanWithName(name="calculator"),
    MaxDuration(seconds=5.0),
    Contains(value="Paris"),
    
    # LLM-based checks (quality assessment)
    HelpfulnessEvaluator(),
    OutputEvaluator(rubric="Assess response quality"),
]
```

### Alternative 2: Separate Assertion Phase
Additional assertions field, need to think about the score and failure model (evaluator runs only when all definitive assertions pass).

```
assertions = [
    Equals(value=expected_output),
    HasSpanWithName(name="calculator"),
    MaxDuration(seconds=5.0),
]
evaluators = [
    HelpfulnessEvaluator(),
    OutputEvaluator(rubric="Assess response quality"),
]
experiment = Experiment(
    cases=test_cases,
    assertions=assertions,
    evaluators=evaluators
)
```

## Use Case

- As a developer testing my agent locally, I want instant feedback when my agent fails basic checks without LLM costs

- As a developer ensuring trace quality, I want to verify execution structure and metadata to catch instrumentation and configuration issues

- I want fast, cheap health checks so that I can continuously verify basic functionality

## Alternatives Solutions

Users can now set up its own custom evaluator with definitive assertions

PR #154, opened 2026-03-10
feat: add deterministic evaluators for output and trajectory checks

afarntrog:
## Description

Adds new evaluators containing deterministic evaluators that don't require LLM calls:

- **Output evaluators**:  for checking actual_output against expected values or custom logic
- **Trajectory evaluator**:  for verifying specific tools were invoked during agent execution

Includes comprehensive unit tests for all new evaluators and exports them from the top-level evaluators package.

## Related Issues

#109 

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/evaluators/deterministic/output.py
+++ b/src/strands_evals/evaluators/deterministic/output.py
@@ -78,5 +78,3 @@ class StartsWith(Evaluator[InputT, OutputT]):
 
     async def evaluate_async(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
         return self.evaluate(evaluation_case)
-
-
--- a/tests/strands_evals/evaluators/deterministic/test_output.py
+++ b/tests/strands_evals/evaluators/deterministic/test_output.py
@@ -223,5 +223,3 @@ class TestStartsWith:
         assert d["evaluator_type"] == "StartsWith"
         assert d["value"] == "Hello"
         assert d["case_sensitive"] is False
-
-
--- a/tests/strands_evals/test_experiment.py
+++ b/tests/strands_evals/test_experiment.py
@@ -9,7 +9,6 @@ from strands.types.exceptions import EventLoopException, ModelThrottledException
 from strands_evals import Case, Experiment
 from strands_evals.evaluators import (
     Contains,
-    Custom,
     Equals,
     Evaluator,
     InteractionsEvaluator,
@@ -1543,20 +1542,11 @@ def test_multiple_deterministic_evaluators_in_experiment():
             Equals(),
             Contains(value="Paris"),
             StartsWith(value="The capital"),
-            Custom(
-                fn=lambda case: [
-                    EvaluationOutput(
-                        score=1.0 if case.actual_output and len(str(case.actual_output)) < 500 else 0.0,
-                        test_pass=case.actual_output is not None and len(str(case.actual_output)) < 500,
-                        reason="Response length check",
-                    )
-                ]
-            ),
         ],
     )
     reports = experiment.run_evaluations(_simulate_agent)
 
-    assert len(reports) == 4
+    assert len(reports) == 3
     for report in reports:
         assert report.scores == [1.0]
         assert report.test_passes == [True]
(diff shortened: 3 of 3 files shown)


src/strands_evals/evaluators/deterministic/output.py
```diff
@@ -0,0 +1,96 @@
...
+    def __init__(self, value: Any | None = None):
+        super().__init__()
+        self.value = value
+
+    def evaluate(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
+        expected = self.value if self.value is not None else evaluation_case.expected_output
+        match = evaluation_case.actual_output == expected
+        return [
+            EvaluationOutput(
+                score=1.0 if match else 0.0,
```
poshinchen, line 19: nit: `score = float(match)` should work. It applies to other score below.

src/strands_evals/evaluators/deterministic/output.py
```diff
@@ -0,0 +1,96 @@
...
+                test_pass=found,
+                reason=f"actual_output {'starts with' if found else 'does not start with'} '{self.value}'",
+            )
+        ]
+
+    async def evaluate_async(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
+        return self.evaluate(evaluation_case)
+
+
+class Custom(Evaluator[InputT, OutputT]):
```
poshinchen: I'm wondering the difference between this one and the CustomEvaluator lol. Aren't they the same?

src/strands_evals/evaluators/deterministic/trajectory.py
```diff
@@ -0,0 +1,47 @@
...
+class ToolCalled(Evaluator[InputT, OutputT]):
+    """Checks if a specific tool was called in the trajectory."""
+
+    def __init__(self, tool_name: str):
+        super().__init__()
+        self.tool_name = tool_name
+
+    def evaluate(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
+        trajectory = evaluation_case.actual_trajectory
+        if trajectory is None:
```
poshinchen, line 15: nit: Do we check `expected_trajectory`?

src/strands_evals/evaluators/deterministic/trajectory.py
```diff
@@ -0,0 +1,47 @@
...
+            return [EvaluationOutput(score=0.0, test_pass=False, reason="no trajectory provided")]
+
+        if isinstance(trajectory, Session):
+            found = self._check_session(trajectory)
```
poshinchen, line 19: This checks whether the tool is called given the whole conversation (i.e. if there are 3 invocations in a conversation, all of them will be checked), should we just check the last one? Or we allow users to set the level to check individual / full session?

src/strands_evals/evaluators/deterministic/output.py
afarntrog: Can you link to the `CustomEvaluator`?

src/strands_evals/evaluators/deterministic/output.py
afarntrog: True. However, i don't think it's intuitive or explicit :) 
> "Explicit is Better Than Implicit”

src/strands_evals/evaluators/deterministic/trajectory.py
afarntrog: True. I would hold off of just checking the last one. Because why the last one and not second to last one etc? This kinda comes back to our multi-turn conversation we had the other day. I would think that we should keep this as the user is basically evaluating, "was this tool_a called within my agent run?"

src/strands_evals/evaluators/deterministic/trajectory.py
afarntrog: I don't think we want it here. Being that the user will pass in the `tool_name` in the constructor I don't think it's necessary to check for `expected_trajectory` it's basically redundant.

src/strands_evals/evaluators/deterministic/output.py
afarntrog: As we discussed offline - we can allow users to create their own custom evaluator by basically doing what we have here. So we can remove this as it really just helps with a bit of the scaffolding.

Issue #110, opened 2026-01-29
[EPIC] Add Environment_State Validation


poshinchen:
### Problem Statement

Strands Evals currently focuses on evaluating agent behavior through trajectories (sequences of actions/tool calls) and outputs (final responses). However, it cannot verify the actual side effects that agents produce in their environment - such as changes to databases, file systems, API states, or other external systems.

### Example
When an agent performs actions that modify external state (e.g., creates a file, updates a database record, sends an email), evaluators can only see:
- What the agent said it did (output)
- What tools it called (trajectory)

But evaluators **cannot** verify:
- Whether the action actually succeeded
- What the final state of the environment looks like
- Whether side effects match expectations

### Phase 1: Environment State Callback Pattern
Concept: Add an optional environment_state_callback to the task function that captures environment state after task execution. This has minimal API changes, and backward compatible

- Task functions returns {"output": ..., "trajectory": ..., "environment_states": {...}}
- The environment_state field contains arbitrary data representing the post-execution environment
- Evaluators receive this state in EvaluationData and can verify it against expectations
- Cases can include expected_environment_state for comparison

### Phase 2: Environment State extractors to get the default states from different clients
- Provide built-in extractors for common scenarios
- Allow custom evaluators to access environment states by key

### Use Case

- As a robotics engineer, I want to evaluate whether my agent successfully moved objects to target locations, so I can verify task completion beyond just checking if the agent called the right movement commands.

- As a database application developer, I want to verify that my agent correctly created, updated, or deleted database records, so I can ensure data integrity beyond just checking if SQL commands were executed.

- As a workflow orchestration developer, I want to verify the final state of a multi-step process (e.g., all artifacts present, correct configuration applied), so I can ensure end-to-end success rather than just individual step completion.
...

PR #156, opened 2026-03-11
feat: add environment state evaluation support

afarntrog:
## Description
Add EnvironmentState type and evaluators for assessing agent side effects on external environments (e.g., file systems, databases, APIs).

- Add EnvironmentState model with name/state fields
- Add expected_environment_state field to Case
- Add actual/expected environment state fields to EvaluationData
- Add StateEquals deterministic evaluator for exact state matching
- Add EnvironmentStateEvaluator (LLM-based) for semantic state evaluation
- Include comprehensive tests for all new functionality

## Related Issues

#110 

## Type of Change

New feature

## Testing

Tested locally and also added comprehensive unit and integ tests.

## Checklist
- [ ] I have read the CONTRIBUTING document

----


src/strands_evals/evaluators/environment_state_evaluator.py
```diff
@@ -0,0 +1,74 @@
...
+from strands.models.model import Model
+from typing_extensions import Union
+
+from ..types.evaluation import EvaluationData, EvaluationOutput, InputT, OutputT
+from .evaluator import Evaluator
+from .prompt_templates.case_prompt_template import compose_test_prompt
+from .prompt_templates.prompt_templates import judge_environment_state_template as SYSTEM_PROMPT
+
+
+class EnvironmentStateEvaluator(Evaluator[InputT, OutputT]):
```
poshinchen: qq: what's the benefit of having `EnvironmentStateEvaluator` separately, instead of having the EnvironmentState into OutputEvaluators' system prompt? We're covering the case where user doesn't care about the output but environmentStates only?

afarntrog: Yeah, the main justification is the user only cares about environment state, not output case. Do you think this warrants another evaluator? I've always seen evaluators as 'cheap' so why not? There will be maintenance as we're vending this one but it can also help guide users how to create one.

afarntrog: As discussed offline we can remove this and repurpose the `OutputEvaluator`

Issue #91, opened 2026-01-23
[EPIC] TraceMappers to map traces from different framework to supported structure


poshinchen:
### Problem Statement

Developers can use various agent frameworks (LangChain, CrewAI, AutoGen, etc.) and need to evaluate agents built with these frameworks. Currently, manual trace conversion is required, creating friction and limiting adoption for strands-evals.

### Proposed Solution

1. LangChainMapper - Convert LangChain traces to Strands conventions
2. OpenInferenceMapper - Convert OpenInference conventions to Strands conventions
3. BraintrustMapper
4. CrewAIMapper - Map CrewAI multi-agent traces
5. AutoGenMapper - Handle AutoGen conversation traces

### Use Case

* As a LangChain user, I want to evaluate my agent with Strands-evals without rewriting my code
* As a framework maintainer, I want to provide native Strands-evals support to my users

joaquinariasco-lab:
Collaboration opportunity for multi-agent automation

Hi Poshinchen,

I saw your work on the Strands TraceMappers EPIC
 and your focus on enabling interoperability between LangChain, CrewAI, AutoGen, and Strands-evals. I’m building a platform where LLM-based agents like yours can automatically discover, cooperate, and exchange work/value with other agents, without manual integration or conversion.

Your expertise in multi-agent systems and framework mappings seems like a perfect fit to be one of the first developers shaping this automated agent economy. I’d love to show you a demo and explore how your agents could connect with others in a shared environment.

Would you be open to a quick call or demo this week?

Best,
Joaquín

PR #164, opened 2026-03-16
feat(mapper): added framework detection for traces from CloudWatch

poshinchen:
## Description
Users could have non-Strands traces in CloudWatch (via the collector). Our current CloudWatchProvider is using CloudWatchSessionMapper which only works for Strands.

Since we're going to have the mapper which is "framework / convention specific", we will deprecate CloudWatchSessionMapper and route the traces / spans to the correct mapper via `detect_otel_mapper` method.

### What has been changed                                                                                                                                                                             
  - **CloudWatchProvider now queries** `aws/spans` log group for **span hierarchy** (spanId → parentSpanId), enriching event records that previously had parent_span_id: None for all spans. 
  This is a best-effort enrichment — failure falls back gracefully.                                                                                                                
  - **LangChainOtelSessionMapper — ADOT span type detection** fixed:                                                                                                                   
    - Inference spans: `role == "unknown"` with raw string content                                                                                                                   
    - Tool execution spans: `input_str` in parsed body (skips `tool_call_with_context` duplicates)                                                                                     
    - Agent invocation spans: `kwargs.name == "LangGraph"`                                                                                                                           
  - **Multi-agent AgentInvocationSpan dedup**:
    - In multi-agent LangGraph systems, each nested sub-graph produces its own LangGraph record. Only the last (root graph) is kept, matching 
  live/Traceloop behavior.                                                                                                                                                         
  - **Fixed leaking mutable state**: 
    - `_trace_tools_map` replaced with local `trace_tools` dict. Added per-call caches (_span_messages_cache, _adot_body_cache) reset each map_to_session() 
  call.                                                                                                                                                                            
  - **available_tools back-fill**: When llm.request.functions.* attributes are missing (ADOT path), builds ToolConfig from converted ToolExecutionSpans.                             
  - **Fixed AgentInvocationSpan creation**: if not user_query → if user_query is None (empty string is valid). _get_last_message_text() now skips tool messages and searches backward  
  for non-empty text.                                                                                                                                                              
  - **Fixed ADOT double-encoded empty strings:** Raw "" (2-char literal) decoded properly, empty assistant messages now return None instead of AssistantMessage(content=[]).           
  - **Extracted hard-coded strings** into module-level constants (_ATTR_*, _KIND_*, _ADOT_*).                                                                                          
  - **Cleanup:** Removed unused log_group parameter from _run_logs_insights_query, removed dead _ATTR_GEN_AI_PREFIX constant, fixed tool_status default from "" to None.

### Tested
  - `TestMapperAutoDetection` — verifies auto-detection for Strands, LangChain (OpenLLMetry), OpenInference scopes, explicit mapper override, and default None mapper
  - `TestScopePreservation` — verifies CloudWatchLogsParser preserves scope.name for LangChain (OpenLLMetry), OpenInference, Strands, spans with events, and synthetic spans
  - Existing `TestSpanConversion` and `TestSessionBuilding` updated to use pytest fixtures (deprecation warning compat)
  - **Tested and compared the mapped sessions using langchain/traceloop, and in_memory_exporter langchain/traceloop.**

## Related Issues

https://github.com/strands-agents/evals/issues/91

## Documentation PR

N/A

## Type of Change
Bug fix
Breaking change

## Testing

## Checklist

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,9 @@ dependencies = [
     "moto>=5.1.0,<6.0.0",
 ]
 
+[[tool.hatch.envs.hatch-test.matrix]]
+python = ["3.13", "3.12", "3.11", "3.10"]
+
 [tool.hatch.envs.default.scripts]
 list = [
     "echo 'Scripts commands available for default env:'; hatch env show --json | jq --raw-output '.default.scripts | keys[]'"
--- a/src/strands_evals/mappers/cloudwatch_parser.py
+++ b/src/strands_evals/mappers/cloudwatch_parser.py
@@ -88,7 +88,12 @@ class CloudWatchLogsParser:
         return result
 
     def _is_event(self, record: dict) -> bool:
-        """Check if record is an OTEL event (has event.name attribute)."""
+        """Check if record is an OTEL event.
+
+        Recognized patterns:
+        1. Has explicit event name (EventName, eventName, or attributes.event.name)
+        2. Has body dict with input/output keys (body-format OTEL log record)
+        """
         if not isinstance(record, dict):
             return False
 
@@ -100,18 +105,34 @@ class CloudWatchLogsParser:
         if isinstance(attrs, dict) and "event.name" in attrs:
             return True
 
+        # Body-format records (body dict with input/output) are also events
+        body = record.get("body")
+        if isinstance(body, dict) and ("input" in body or "output" in body):
+            return True
+
         return False
 
     def _is_span(self, record: dict) -> bool:
-        """Check if record is an OTEL span (has start/end time)."""
+        """Check if record is an OTEL span.
+
+        Recognized patterns:
+        1. Has start/end timestamps (startTimeUnixNano/endTimeUnixNano or start_time/end_time)
+        2. Already normalized format with span_events list (pass-through)
+        """
         if not isinstance(record, dict):
             return False
 
         # Spans have startTimeUnixNano or start_time
         has_start = "startTimeUnixNano" in record or "start_time" in record
         has_end = "endTimeUnixNano" in record or "end_time" in record
+        if has_start and has_end:
+            return True
 
-        return has_start and has_end
+        # Already-normalized spans have span_events list
+        if "span_events" in record and isinstance(record.get("span_events"), list):
+            return True
+
+        return False
 
     def _normalize_event(self, record: dict) -> dict | None:
         """Normalize a CloudWatch event record."""
@@ -175,7 +196,7 @@ class CloudWatchLogsParser:
                     "version": scope.get("version", ""),
                 },
                 "status": {"code": status.get("code", "UNSET")},
-                "span_events": [],  # Will be populated later
+                "span_events": record.get("span_events", []),  # Preserve existing or populate later
             }
         except Exception as e:
             logger.warning(f"Failed to normalize span: {e}")
@@ -183,18 +204,23 @@ class CloudWatchLogsParser:
 
     def _create_synthetic_span(self, span_id: str, events: list[dict], first_event: dict) -> dict:
         """Create a synthetic span from events when no span records exist."""
-        # Get trace_id from the raw logs (need to find matching record)
+        # Get trace_id and scope from the raw logs
         trace_id = ""
+        scope_name = ""
         for record in self.raw_logs:
             if record.get("spanId") == span_id or record.get("span_id") == span_id:
-                trace_id = record.get("traceId") or record.get("trace_id", "")
-                break
+                trace_id = trace_id or record.get("traceId") or record.get("trace_id", "")
+                raw_scope = record.get("scope")
+                if isinstance(raw_scope, dict):
+                    scope_name = scope_name or raw_scope.get("name", "")
 
-        # Get scope from event name
-        event_name = first_event.get("event_name", "")
-        scope_name = first_event.get("attributes", {}).get("event.name", event_name)
+        # Fall back to event name as scope hint
+        if not scope_name:
+            event_name = first_event.get("event_name", "")
+            scope_name = first_event.get("attributes", {}).get("event.name", event_name)
 
         # Use event timestamp for span times
+        event_name = first_event.get("event_name", "")
         timestamp = first_event.get("timestamp", 0)
 
         return {
--- a/src/strands_evals/mappers/langchain_otel_session_mapper.py
+++ b/src/strands_evals/mappers/langchain_otel_session_mapper.py
@@ -37,6 +37,26 @@ logger = logging.getLogger(__name__)
 
 SCOPE_NAME = "opentelemetry.instrumentation.langchain"
 
+# --- OTEL semantic convention attribute keys ---
+_ATTR_LLM_REQUEST_TYPE = "llm.request.type"
+
+# --- Traceloop/OpenLLMetry attribute keys ---
+_ATTR_TRACELOOP_SPAN_KIND = "traceloop.span.kind"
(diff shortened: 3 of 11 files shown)


src/strands_evals/mappers/cloudwatch_parser.py
```diff
@@ -183,18 +183,24 @@ def _normalize_span(self, record: dict) -> dict | None:
...
         trace_id = ""
+        scope_name = ""
         for record in self.raw_logs:
             if record.get("spanId") == span_id or record.get("span_id") == span_id:
-                trace_id = record.get("traceId") or record.get("trace_id", "")
-                break
+                trace_id = trace_id or record.get("traceId") or record.get("trace_id", "")
+                # Prefer scope.name from raw record if available
+                raw_scope = record.get("scope", {})
+                if isinstance(raw_scope, dict) and raw_scope.get("name"):
```
afarntrog: what other types can `raw_scope` be and should we just ignore it if it's not a dict? Why are we not handling that?

tests/strands_evals/mappers/test_cloudwatch_parser.py
```diff
@@ -208,6 +208,73 @@ def test_multiple_spans_multiple_traces(self):
         assert span_ids == {"s1", "s2"}
 
 
+class TestScopePreservation:
+    """Task 0.2: Verify scope.name is preserved for all frameworks."""
```
afarntrog: nit: `Task 0.2:`

pyproject.toml
```diff
@@ -14,11 +14,11 @@ authors = [
 ]
 
 dependencies = [
-    "pydantic>=2.0.0,<3.0.0",
+    "pydantic>=2.4.0,<3.0.0",
     "rich>=14.0.0,<15.0.0",
     "strands-agents>=1.0.0",
     "strands-agents-tools>=0.1.0,<1.0.0",
-    "typing-extensions>=4.0",
+    "typing-extensions>=4.13.2,<5.0.0",
```
afarntrog: Is upgrading this a risk? Can it break our customers?

tests/strands_evals/providers/test_cloudwatch_provider.py
```diff
@@ -350,3 +351,80 @@ def test_output_from_last_agent_invocation(self, provider, mock_logs_client):
...
+        _setup_query_results(mock_logs_client, [span])
+
+        with patch("boto3.client", return_value=mock_logs_client):
+            p = CloudWatchProvider(log_group="/test/group")
+            with patch.object(OpenInferenceSessionMapper, "map_to_session") as mock_map:
+                mock_map.return_value = Session(session_id="sess-1", traces=[])
+                try:
+                    p.get_evaluation_data("sess-1")
+                except Exception:
+                    pass
```
afarntrog: Is this `try/except` defeating the purpose of the test?

tests/strands_evals/providers/test_cloudwatch_provider.py
```diff
@@ -350,3 +351,80 @@ def test_output_from_last_agent_invocation(self, provider, mock_logs_client):
...
+        _setup_query_results(mock_logs_client, [span])
+
+        with patch("boto3.client", return_value=mock_logs_client):
+            p = CloudWatchProvider(log_group="/test/group")
+            with patch.object(LangChainOtelSessionMapper, "map_to_session") as mock_map:
+                mock_map.return_value = Session(session_id="sess-1", traces=[])
+                try:
+                    p.get_evaluation_data("sess-1")
+                except Exception:
+                    pass
```
afarntrog: Is this `try/except` defeating the purpose of the test?

tests/strands_evals/mappers/test_cloudwatch_parser.py
poshinchen: LOLLL

pyproject.toml
poshinchen: This is what we currently have in Strands, but yeah I can omit them

src/strands_evals/mappers/cloudwatch_parser.py
poshinchen: yeah this is the type thingy where the scope (if exists) should always be a dict, but since it's raw data it could fail if the scope doesn't exist.

src/strands_evals/providers/cloudwatch_provider.py
```diff
@@ -96,20 +112,44 @@ def _discover_log_group(self, agent_name: str) -> str:
...
             raise SessionNotFoundError(f"CloudWatch: no spans found for session_id='{session_id}'")
 
-        session = self._mapper.map_to_session(span_dicts, session_id)
+        # Normalize raw CloudWatch log records into the standard span format.
+        span_dicts = CloudWatchLogsParser(raw_records).parse()
+
+        if not span_dicts:
+            raise SessionNotFoundError(f"CloudWatch: no normalizable spans found for session_id='{session_id}'")
+
+        # Enrich with hierarchy from aws/spans (best-effort).
```
afarntrog, line 143: Is best effort worth it? If we can't consistently get build the hierarchy should we just never do it so that we can be consistent?

src/strands_evals/providers/cloudwatch_provider.py
```diff
@@ -119,19 +159,44 @@ def get_evaluation_data(self, session_id: str) -> TaskOutput:
...
         now = datetime.now(tz=timezone.utc)
-        end_time = now
         start_time = now - timedelta(days=self._lookback_days)
 
         try:
             response = self._client.start_query(
                 logGroupName=self._log_group,
                 startTime=int(start_time.timestamp()),
-                endTime=int(end_time.timestamp()),
+                endTime=int(now.timestamp()),
```
afarntrog, line 199: `now.timestamp())` -> I think we should accept a end time as someone can run evals on runs from weeks ago. let's consider a `range(start, end)`

src/strands_evals/mappers/langchain_otel_session_mapper.py
```diff
@@ -37,6 +37,26 @@
...
+# --- Traceloop/OpenLLMetry attribute values ---
+_KIND_WORKFLOW = "workflow"
+_KIND_TOOL = "tool"
+_LLM_TYPE_CHAT = "chat"
+
+# --- ADOT body field values (LangChain serialization) ---
+_ADOT_ROLE_UNKNOWN = "unknown"
+_ADOT_LANGGRAPH_NAME = "LangGraph"
+_ADOT_TOOL_CALL_WITH_CONTEXT = "tool_call_with_context"
+_ADOT_INPUT_STR_KEY = "input_str"
```
afarntrog, line 58: Does it make sense to have these in a separate config file?

src/strands_evals/providers/cloudwatch_provider.py
poshinchen: yeah it's kind of required to build the hierarchy so that the evaluator would not be messed up. When you're using in_memory exporter the parentSpanId is there, but it isn't in the cloudwatch adot logs.

src/strands_evals/providers/cloudwatch_provider.py
poshinchen: oh yeah, startTime is from weeks ago, and the end time is now. But I'll be dine allowing users to pass end_time too.

src/strands_evals/mappers/langchain_otel_session_mapper.py
poshinchen: I can start from creating a internal constant file, then we can decide what to be moved to config!

Issue #159, opened 2026-03-12
[FEATURE] Update Tracing to support Langfuse Observation types


lilianskim:
### Problem Statement

Langfuse has updated their spans to support more types for filtering 
https://langfuse.com/docs/observability/features/observation-types
Currently strands only supports a few of these types

### Proposed Solution

Have a new decorator or update the existing tool decorator to support these additional types. 

### Use Case

For better filtering for evals and metrics against these different observation types

poshinchen:
Hi, are you storing the traces in Langfuse? If so, could you check if this PR works for you?
https://github.com/strands-agents/evals/pull/165/changes

PR #165, opened 2026-03-16
feat(mappers): langfuse provider changes to support newer version of langfuse

poshinchen:
## Description
### LangfuseProvider multi-framework support                                                                                                                                     
  - Makes LangfuseProvider work with traces from any framework (LangChain, Strands, LlamaIndex, CrewAI, etc.) — not just Strands.                                                  

This PR makes the provider framework-agnostic by routing on **obs.type** (universal across all frameworks in Langfuse) with a Strands-specific fallback for backward compatibility.                                                                                                                                                                   
                                                            
### Changes                                                                                                                                                                      
- `langfuse_provider.py` — Multi-framework observation routing:
- **_convert_observation()**: Route by universal obs.type (GENERATION → InferenceSpan, TOOL → ToolExecutionSpan, CHAIN/AGENT root → AgentInvocationSpan) with Strands SPAN name-based fallback preserved                                                                                                                                                    
- **_convert_tool_execution()**: Handle both Strands format (obs.input.name) and LangChain/universal format (obs.name + obs.input as arguments)                              
- **_parse_tool_result():** Support LangChain ToolMessage format ({"content": "...", "type": "tool"})                                                                        
- **_convert_message():** Parse LangChain tool_calls field on assistant messages                                                                                             
- **_extract_user_prompt():** Handle LangChain CHAIN formats ({"input": "..."}, {"messages": [...]}) and Strands message-list format via OTEL ([{"role": "user", "content":  '[{"text": "..."}]'}])                                                                                                                                                           
- **_extract_agent_response():** Handle LangChain CHAIN output formats ({"output": "..."}, {"messages": [...]}) and Strands message-list output format                       
- **_extract_text_from_content():** New helper to unwrap Strands content (plain string, JSON-encoded [{"text": "..."}], or parsed list)                                      
- **pyproject.toml:** Cap langfuse dependency at <4

## Related Issues

https://github.com/strands-agents/evals/issues/91
https://github.com/strands-agents/evals/issues/159

## Documentation PR

N/A

## Type of Change
New feature

## Testing

## Checklist

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dev = [
     "ruff>=0.13.0,<0.15.0",
 ]
 
-langfuse = ["langfuse>=2.0.0,<3"]
+langfuse = ["langfuse>=2.0.0,<4"]
 otel = ["opentelemetry-exporter-otlp-proto-http>=1.30.0,<2.0.0"]
 langchain = [
     "langchain>=0.3.0",
--- a/src/strands_evals/providers/langfuse_provider.py
+++ b/src/strands_evals/providers/langfuse_provider.py
@@ -195,12 +195,19 @@ class LangfuseProvider(TraceProvider):
     def _convert_observation(self, obs: Any, session_id: str) -> Any:
         """Route a single Langfuse observation to the appropriate span converter.
 
-        Langfuse observation fields used for routing:
-            obs.type: str  — "GENERATION" | "SPAN" | "EVENT" | ...
-            obs.name: str  — e.g. "execute_tool calc", "invoke_agent my_agent", "chat"
+        Langfuse normalizes traces from ALL frameworks into its own Observation
+        format. The `obs.type` field is universal across frameworks:
+
+        - GENERATION — LLM call (LangChain, Strands, LlamaIndex, etc.)
+        - TOOL — Tool invocation (LangChain sends these)
+        - CHAIN — Orchestration/agent (root chain = agent invocation)
+        - SPAN — Strands-specific spans (fallback by ``obs.name``)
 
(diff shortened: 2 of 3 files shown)


src/strands_evals/mappers/cloudwatch_parser.py
```diff
@@ -183,18 +183,24 @@ def _normalize_span(self, record: dict) -> dict | None:
 
     def _create_synthetic_span(self, span_id: str, events: list[dict], first_event: dict) -> dict:
         """Create a synthetic span from events when no span records exist."""
-        # Get trace_id from the raw logs (need to find matching record)
+        # Get trace_id and scope from the raw logs
```
afarntrog, line 207: these are the same changes as https://github.com/strands-agents/evals/pull/164/changes :) are they dependent on each other?

pyproject.toml
```diff
@@ -44,7 +44,7 @@ dev = [
     "ruff>=0.13.0,<0.15.0",
 ]
 
-langfuse = ["langfuse>=2.0.0,<3"]
+langfuse = ["langfuse>=2.0.0,<5"]
```
afarntrog: this seems like a wide range. Also, my understanding is that the versions have drastic differences which would be incompatible with evals sdk (one of the reasons we introduced this PR). I would recommend a range of: `langfuse>=4.0.0,<5`

poshinchen: I think I would be fine with `langfuse>=3,< 5`, since it basically means that older version of langfuse will be deprecated. v2/3 has similar concept/component, but v3/v4 has some major diffs and I prefer to keep both versions.

PR #166, opened 2026-03-18
feat: Add EvaluationPlugin for agent invocation evaluation and retry

afarntrog:
## Description

Adds a new `EvaluationPlugin` that integrates with the Strands agents plugin system to evaluate agent output after each invocation and automatically retry with improved system prompts on failure.

### Changes

**New: `strands_evals.plugins` module**

- **`EvaluationPlugin`** — A Strands `Plugin` that wraps agent invocations to:
  - Run configured evaluators against agent output after each call
  - On evaluation failure, use an LLM to generate an improved system prompt based on the failure reasons
  - Automatically retry the invocation with the improved prompt (configurable `max_retries`)
  - Restore the original system prompt after all attempts complete
  - Return a structured `ImprovementSuggestion` (reasoning + system prompt) from the improvement LLM
- **`improvement_suggestion.py`** — Prompt template for the improvement suggestion agent that analyzes evaluation failures and proposes system prompt modifications
- Expected output and trajectory can be set at construction time or overridden per-invocation via `invocation_state`

**Other changes**

- Bumped `strands-agents` minimum version from `>=1.0.0` to `>=1.28.0` (required for the plugin interface)
- Added ruff line-length exemption for `plugins/prompt_templates/`

### Testing

- 480 lines of unit tests covering: initialization, agent wrapping, evaluator execution, retry behavior, system prompt restoration, improvement suggestion generation, partial failure handling, max retries, and multi-evaluator scenarios

---

Let me know if you want any tweaks to this.

## Type of Change

New feature

## Testing

## Checklist

----


src/strands_evals/plugins/evaluation_plugin.py
```diff
@@ -0,0 +1,205 @@
...
+
+        def wrapped_call(self_agent: Any, prompt: Any = None, **kwargs: Any) -> Any:
+            return plugin._invoke_with_evaluation(self_agent, original_call, prompt, **kwargs)
+
+        wrapped_class = type(
+            agent.__class__.__name__,
+            (agent.__class__,),
+            {"__call__": wrapped_call},
+        )
+        agent.__class__ = wrapped_class
```
afarntrog, line 83: We need to do this monkey patching because hooks alone will not give us the retry pattern we need:

What EvaluationPlugin needs to do:
  1. Intercept an agent invocation
  2. Let it run
  3. Evaluate the result
  4. If evaluation fails, modify the system prompt and run the agent again
  5. Repeat up to `max_retries` times

What hooks give you:
  - `BeforeInvocationEvent` fires before the agent runs. You can modify messages. No result yet.
  - `AfterInvocationEvent` fires after the agent runs. You get the result. But you're inside `_run_loop's` finally block, and `stream_async` still holds the invocation lock. If you call `agent()` from here → deadlock.

Hooks give you "before" and "after", not "around". There's no way to say "evaluate this result and, if it's bad, run the whole thing again" from within a hook. The retry requires calling `agent()`, which requires the lock, which is held by the invocation that's firing your hook.  `__class__` swap helps because:
This replaces `__call__` — every call to `agent(prompt)` now goes through `_invoke_with_evaluation`, which runs evaluators and handles retries.

PR #168, opened 2026-03-18
feat: add pre/post call hooks to ToolSimulator (#167)

### LINKED ISSUE #167: [FEATURE] Pre/post call hooks on ToolSimulator for fault injection and response modification
### Problem Statement

When using `ToolSimulator` to test agent resilience (not just workflow correctness), there's no way to intercept tool calls before or after the LLM generates a response. This makes it impossible to programmatically inject faults like rate limits, timeouts, or partial outages.

We tried several approaches:
- Putting failure instructions in the state description ("return a 429 error on call number 4"). The LLM ignores this consistently due to its helpfulness bias, even with strong prompting. Even when the state explicitly says "MUST return error" and "if you skip the error the test fails," the LLM produces a successful response. This is also unreliable due to LLM non-determinism: on rare occasions it might comply, but you can't depend on it for repeatable testing.
- Wrapping the function with a decorator before registration. Doesn't work because the simulator only uses the function for its name, signature, and docstring. At call time, `_create_tool_wrapper` replaces the function entirely with LLM inference, so the decorator never executes.

Additionally, when tools are called concurrently by the agent, there's no way to coordinate fault behavior across parallel calls using the state-based approach. The LLM generating each tool response operates independently and has no awareness of what other concurrent calls are doing or returning.

The only working approach is subclassing `ToolSimulator` and overriding `_call_tool`, which couples to a private method.

### Proposed Solution

Add optional `pre_call_hook` and `post_call_hook` parameters to `ToolSimulator.__init__`:

```python
simulator = ToolSimulator(
    model=model,
    pre_call_hook=my_fault_injector,
    post_call_hook=my_response_modifier,
)
```

The `pre_call_hook` is called before the LLM generates a response. It receives the tool name, parameters, state key, and previous call history. If it returns a non-None dict, that dict is returned as the tool response (short-circuiting the LLM call). If it returns None, normal simulation proceeds. The fault response should still be cached via `state_registry.cache_tool_call` so subsequent calls see the failure in their context.

```python
def my_fault_injector(tool_name, parameters, state_key, previous_calls):
    if random.random() < 0.3:
        return {"error": {"code": "QuotaExceeded", "retryAfterSeconds": 2}}
    return None
```

The `post_call_hook` is called after the LLM generates a response but before it's cached. It receives the same context plus the response dict, and returns a (possibly modified) response.

```python
def my_response_modifier(tool_name, parameters, state_key, response):
    response["_simulated_latency_ms"] = random.randint(50, 500)
    return response
```

This is a small change (two optional params on `__init__`, a few lines in `_call_tool`), fully backward compatible, and follows the same hook pattern the SDK uses at the agent level.

### Use Case

- Rate limiting: randomly return 429 errors to test agent retry logic
- Partial failures: specific tools fail intermittently while others work
- Timeouts: hard cutoff after N calls to test graceful degradation
- Response modification: inject latency metadata, add missing fields, corrupt responses for robustness testing
- Chaos testing: compose multiple fault types to simulate degraded API conditions

### Alternatives Solutions

- State-based LLM instructions: unreliable, LLM ignores error instructions
- Subclassing `ToolSimulator` and overriding `_call_tool`: works but couples to a private method

### Additional Context

Related to [issue #114](https://github.com/strands-agents/evals/issues/114) (Chaos/Resiliency Evaluation of Agents). The hooks proposed here would provide the low-level extension point that a higher-level chaos testing library (as described in #114) could build on.

We have a working implementation using the subclass approach and are happy to contribute a PR for the hooks.

kaghatim:
## Description

Add optional `pre_call_hook` and `post_call_hook` parameters to `ToolSimulator.__init__` for fault injection and response modification during tool simulation.

- `pre_call_hook`: invoked before the LLM generates a response. Receives a `PreCallHookEvent` dataclass. Returns a dict to short-circuit the LLM call, or `None` to proceed normally. Short-circuited responses are cached in the state registry.
- `post_call_hook`: invoked after the LLM generates a response but before caching. Receives a `PostCallHookEvent` dataclass. Returns a (possibly modified) response dict.

Hook events follow the Strands SDK pattern (`@dataclass` events, before/after separation) and live in `types/simulation/hook_events.py`.

Both hooks are optional and default to `None`, making this fully backward compatible.

## Related Issues

Closes #167

## Documentation PR

N/A - No new docs needed for this change. Hook usage is documented in the `__init__` docstring.

## Type of Change

New feature

## Testing

- Added 12 unit tests covering: short-circuit, passthrough, response modification, caching of hook responses, call history accumulation, shared state, both hooks together, post-hook skipped on short-circuit, and backward compatibility.
- All 24 tests pass via `hatch test tests/strands_evals/simulation/test_tool_simulator.py`

## Checklist

----

### DIFF
--- a/src/strands_evals/simulation/tool_simulator.py
+++ b/src/strands_evals/simulation/tool_simulator.py
@@ -11,6 +11,7 @@ from strands.agent import AgentResult
 from strands.models.model import Model
 from strands.tools.decorator import DecoratedFunctionTool, FunctionToolMetadata
 
+from strands_evals.types.simulation.hook_events import PostCallHookEvent, PreCallHookEvent
 from strands_evals.types.simulation.tool import DefaultToolResponse, RegisteredTool
 
 from .prompt_templates.tool_response_generation import TOOL_RESPONSE_PROMPT_TEMPLATE
@@ -166,6 +167,8 @@ class ToolSimulator:
         state_registry: StateRegistry | None = None,
         model: Model | str | None = None,
         max_tool_call_cache_size: int = 20,
+        pre_call_hook: Callable | None = None,
+        post_call_hook: Callable | None = None,
     ):
         """
         Initialize a ToolSimulator instance.
@@ -178,10 +181,21 @@ class ToolSimulator:
                                      Only used when creating a new StateRegistry (ignored if state_registry
                                      is provided). Older calls are automatically evicted when limit is exceeded.
                                      Default is 20.
+            pre_call_hook: Optional callable invoked before the LLM generates a tool response.
+                          Receives a PreCallHookEvent with tool_name, parameters, state_key,
+                          and previous_calls. If it returns a non-None dict, that dict is used
+                          as the tool response (short-circuiting the LLM call) and cached in
+                          the state registry. If it returns None, normal LLM simulation proceeds.
+            post_call_hook: Optional callable invoked after the LLM generates a tool response
+                           but before it is cached. Receives a PostCallHookEvent with tool_name,
+                           parameters, state_key, and response. Must return a (possibly modified)
+                           response dict.
         """
         self.model = model
         self.state_registry = state_registry or StateRegistry(max_tool_call_cache_size=max_tool_call_cache_size)
         self._registered_tools: dict[str, RegisteredTool] = {}
+        self._pre_call_hook = pre_call_hook
+        self._post_call_hook = post_call_hook
 
     def _create_tool_wrapper(self, registered_tool: RegisteredTool):
         """
@@ -245,7 +259,35 @@ class ToolSimulator:
         return response_data
 
     def _call_tool(self, registered_tool: RegisteredTool, parameters_string: str, state_key: str) -> dict[str, Any]:
-        """Simulate a tool invocation and return the response."""
+        """Simulate a tool invocation and return the response.
+
+        If a pre_call_hook is configured and returns a non-None dict, that dict is used
+        as the tool response (short-circuiting the LLM call). The response is still cached.
+
+        If a post_call_hook is configured, it receives the LLM-generated response before
+        caching and may modify it.
+        """
+        parameters = json.loads(parameters_string)
+        current_state = self.state_registry.get_state(state_key)
+
+        # Pre-call hook: may short-circuit the LLM call
+        if self._pre_call_hook is not None:
+            event = PreCallHookEvent(
+                tool_name=registered_tool.name,
+                parameters=parameters,
+                state_key=state_key,
+                previous_calls=current_state.get("previous_calls", []),
+            )
+            hook_response = self._pre_call_hook(event)
+            if hook_response is not None:
+                if not isinstance(hook_response, dict):
+                    raise TypeError(f"pre_call_hook must return a dict or None, got {type(hook_response).__name__}")
+                self.state_registry.cache_tool_call(
+                    registered_tool.name, state_key, hook_response, parameters=parameters
+                )
+                return hook_response
+
+        # Normal LLM simulation
         # Get input schema from Strands tool decorator
         input_schema_dict = registered_tool.function.tool_spec.get("inputSchema", {}).get("json", {})
         input_schema = json.dumps(input_schema_dict, indent=2)
@@ -253,8 +295,6 @@ class ToolSimulator:
         # Get output schema as JSON string using Pydantic's model_json_schema (output_schema is mandatory)
         output_schema = registered_tool.output_schema.model_json_schema()
         output_schema_string = json.dumps(output_schema, indent=2)
-
-        current_state = self.state_registry.get_state(state_key)
 
         prompt = TOOL_RESPONSE_PROMPT_TEMPLATE.format(
             tool_name=registered_tool.name,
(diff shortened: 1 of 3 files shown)


src/strands_evals/simulation/tool_simulator.py
```diff
@@ -245,7 +259,33 @@ def _parse_simulated_response(self, result: AgentResult) -> dict[str, Any]:
...
+        # Pre-call hook: may short-circuit the LLM call
+        if self._pre_call_hook is not None:
+            event = PreCallHookEvent(
+                tool_name=registered_tool.name,
+                parameters=parameters,
+                state_key=state_key,
+                previous_calls=current_state.get("previous_calls", []),
+            )
+            hook_response = self._pre_call_hook(event)
+            if hook_response is not None:
```
ybdarrenwang, line 282: Should also verify if hook_response is a dict?

kaghatim: Yeah good point - will add a dict type check for both hooks so it fails fast with a clear message.

src/strands_evals/simulation/tool_simulator.py
```diff
@@ -245,7 +259,35 @@ def _parse_simulated_response(self, result: AgentResult) -> dict[str, Any]:
         return response_data
 
     def _call_tool(self, registered_tool: RegisteredTool, parameters_string: str, state_key: str) -> dict[str, Any]:
-        """Simulate a tool invocation and return the response."""
+        """Simulate a tool invocation and return the response.
```
ybdarrenwang, line 262: This implementation reminds me of decorator pattern, which also applies changes pre and post a decorated function. Wondering if it might be better to refactor this with decorator pattern?
```
    def _apply_pre_call_hook(self, func: Callable) -> Callable:
        """Internal decorator that applies pre_call_hook logic."""
        if self._pre_call_hook is None:
            return func
        
        @wraps(func)
        def wrapper(registered_tool: RegisteredTool, parameters_string: str, state_key: str) -> dict[str, Any]:
            parameters = json.loads(parameters_string)
            current_state = self.state_registry.get_state(state_key)
            
            # Invoke pre-call hook
            event = PreCallHookEvent(
                tool_name=registered_tool.name,
                parameters=parameters,
                state_key=state_key,
                previous_calls=current_state.get("previous_calls", []),
            )
            hook_response = self._pre_call_hook(event)
            
            # Short-circuit if hook returns a response
            if hook_response is not None:
                if not isinstance(hook_response, dict):
                    raise TypeError(f"pre_call_hook must return a dict or None, got {type(hook_response).__name__}")
                self.state_registry.cache_tool_call(
                    registered_tool.name, state_key, hook_response, parameters=parameters
                )
                return hook_response
            
            # Otherwise, proceed with normal execution
            return func(registered_tool, parameters_string, state_key)
        
        return wrapper

    def _apply_post_call_hook(self, func: Callable) -> Callable:
        """Internal decorator that applies post_call_hook logic."""
        if self._post_call_hook is None:
            return func
        
        @wraps(func)
        def wrapper(registered_tool: RegisteredTool, parameters_string: str, state_key: str) -> dict[str, Any]:
            # Execute the function to get response
            response_data = func(registered_tool, parameters_string, state_key)
            
            # Apply post-call hook to modify response
            parameters = json.loads(parameters_string)
            event = PostCallHookEvent(
                tool_name=registered_tool.name,
                parameters=parameters,
                state_key=state_key,
                response=response_data,
            )
            modified_response = self._post_call_hook(event)
            
            if not isinstance(modified_response, dict):
                raise TypeError(f"post_call_hook must return a dict, got {type(modified_response).__name__}")
            
            return modified_response
        
        return wrapper

......

    def _create_tool_wrapper
        ......
        def wrapper(*args, **kwargs):
                # Apply decorators in reverse order (post first, then pre)
                # This ensures execution order is: pre -> core logic -> post
                call_func = self._call_tool
                call_func = self._apply_post_call_hook(call_func)
                call_func = self._apply_pre_call_hook(call_func)
                ......
                return call_func
```

I think the pros are a) self._call_tool is not touched, b) more modularized, c) it seems to me that the pre/post hooks need access to many members of TS. Instead of relying on Event classes we might rather give them full access to self. Though I also worry this seems an overkill instead of just modifying _call_tool. @poshinchen thoughts?

kaghatim: Yeah the reason I went with hooks over subclassing was specifically because `_call_tool` is private. I didn't want to couple to its internals or commit its signature as public API. The event classes keep the contract narrow so `_call_tool` can be refactored freely. But I'm open to a different approach if you have a preference. I would be happy to rework it.

Edit: Good point on the decorator pattern. I think the inline approach is simpler here, it avoids re-parsing `parameters_string` across decorators and keeps the flow in one place. Happy to wait for @poshinchen's take.

src/strands_evals/types/simulation/hook_events.py
```diff
@@ -0,0 +1,38 @@
...
+@dataclass
+class PostCallHookEvent:
+    """
+    Event passed to post_call_hook after the LLM generates a tool response.
+
+    Attributes:
+        tool_name: Name of the tool that was called.
+        parameters: Parsed parameters for the tool call.
+        state_key: Key for the state (tool_name or share_state_id).
+        response: The LLM-generated response dict, which the hook may modify.
```
ybdarrenwang, line 32: Also just some food for thoughts:
- What if we only add 1 new `class CallHookEvent`, just fetch response after it's cached into `previos_calls[-1]`, and update it if post hook is applied?
- What if we just throw the entire TS self to the hooks?

My worry is these event types are ad-hoc and bonded to TS changes. Say if in the future we want access to more TS internal members and states, one has to touch these types again.

kaghatim: On one class with `previous_calls[-1]`: right now the post-hook runs before caching so it returns a modified response and that's what gets stored. Caching first and having the hook mutate `previous_calls[-1]` in place would work, but it makes the hook a side-effect function instead of a simple transform. I went with the return-a-value pattern to keep hooks easier to test and reason about.

On passing `self`: I kept the events narrow to avoid coupling hooks to TS internals, but I hear the concern about the types growing over time. Happy to rework if that's the preference.

ybdarrenwang:
Confirmed this feature can be realized by existing Plugin capability.

Issue #95, opened 2026-01-26
[EPIC] Enable Using ground truth for evaluators


poshinchen:
### Problem Statement

I would like strands-evals to be able to take the ground truth to help evaluators make the evaluation decision.

### Proposed Solution

Allow evaluators to take the ground truth and format them into the system prompt for decision. This requires API changes in the evaluator to whether take the dataset into consideration or not.

### Use Case

Given multiple evaluation results telling the ground truth, the evaluators can further make better decisions for further evaluations.

PR #180, opened 2026-03-26
feat: add ground truth assertion support to Goal Success Rate evaluator

ybdarrenwang:
## Description
This PR enhances the GoalSuccessRateEvaluator to support assertion-based evaluation mode alongside the existing conversation-analysis mode. The evaluator now accepts ground truth assertions via metadata and validates agent behavior against these explicit success criteria.

## Related Issues
#95 

## Type of Change
New feature

## Testing

## Checklist

----


poshinchen (dismissed):
1. I would prefer the assertions to be `expected_criteria` / `expected_goal` to align with what other evaluators have, and to avoid confusion between deterministic assertions
  1.2. Add `expected_criteria : str | None = None` to EvaluationData and Case
2. How useful is the `additional_context`?
3. Can't the prompt be goal_success_rate_v1? what's the major diff?

src/strands_evals/evaluators/goal_success_rate_evaluator.py
```diff
@@ -36,19 +59,52 @@ class GoalSuccessRateEvaluator(Evaluator[InputT, OutputT]):
...
+        if self._has_assertions(evaluation_case):
+            return await self._evaluate_with_assertions_async(session_input, evaluation_case)
+
+        return await self._evaluate_basic_async(session_input)
+
+    def _evaluate_basic(self, session_input: SessionLevelInput) -> list[EvaluationOutput]:
+        """Evaluate goal success using the basic prompt (no assertions)."""
         prompt = self._format_prompt(session_input)
         evaluator_agent = Agent(model=self.model, system_prompt=self.system_prompt, callback_handler=None)
         result = evaluator_agent(prompt, structured_output_model=GoalSuccessRating)
```
afarntrog, line 103: The `sync` and `async` are essentially the same. We refactored the underlying logic so you technically only need the implement one because we **always** call the `async` version now.
- If your code is async then just implement the async version
- If you code is sync then just implement the sync version. We automatically delegate it correctly.

ybdarrenwang: Will remove async methods in the next revision.

afarntrog (approved):

PR #185, opened 2026-04-02
feat: add correctness evaluator, trace-based and reference-based

ybdarrenwang:
## Description
Add correctness evaluator, supporting both trace-based and reference-based evaluation

## Related Issues

#95 

## Documentation PR

https://github.com/strands-agents/docs/pull/714

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/evaluators/coherence_evaluator.py
+++ b/src/strands_evals/evaluators/coherence_evaluator.py
@@ -74,13 +74,6 @@ class CoherenceEvaluator(Evaluator[InputT, OutputT]):
         result = evaluator_agent(prompt, structured_output_model=CoherenceRating)
         return self._create_evaluation_output(result)
 
-    async def evaluate_async(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
-        parsed_input = self._get_last_turn(evaluation_case)
-        prompt = self._format_prompt(parsed_input)
-        evaluator_agent = Agent(model=self.model, system_prompt=self.system_prompt, callback_handler=None)
-        result = await evaluator_agent.invoke_async(prompt, structured_output_model=CoherenceRating)
-        return self._create_evaluation_output(result)
-
     def _create_evaluation_output(self, result) -> list[EvaluationOutput]:
         rating = cast(CoherenceRating, result.structured_output)
         normalized_score = self._score_mapping[rating.score]
@@ -92,36 +85,6 @@ class CoherenceEvaluator(Evaluator[InputT, OutputT]):
                 label=rating.score,
             )
         ]
-
-    def _get_last_turn(self, evaluation_case: EvaluationData[InputT, OutputT]) -> TraceLevelInput:
-        """Extract the most recent turn from the conversation for evaluation."""
-        parsed_inputs = self._parse_trajectory(evaluation_case)
-        if not parsed_inputs:
-            raise ValueError(
-                "No turn-level inputs could be parsed from the trajectory. "
-                "Ensure actual_trajectory is a Session with at least one AgentInvocationSpan."
-            )
-        return parsed_inputs[-1]
-
-    def _extract_user_prompt(self, parsed_input: TraceLevelInput) -> str:
-        """Extract user prompt from last message in session history.
-
-        Args:
-            parsed_input: Trace-level input containing session history
-
-        Returns:
-            User prompt text, or empty string if not available
-        """
-        if not parsed_input.session_history:
-            return ""
-
-        last_msg = parsed_input.session_history[-1]
-        if not isinstance(last_msg, list) and self._has_text_content(last_msg):
-            first_content = last_msg.content[0]
-            if isinstance(first_content, TextContent):
-                return first_content.text
-
-        return ""
 
     def _format_prompt(self, parsed_input: TraceLevelInput) -> str:
         """Format evaluation prompt from parsed trace data.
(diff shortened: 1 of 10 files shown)


src/strands_evals/evaluators/prompt_templates/correctness/correctness_v0.py
```diff
@@ -0,0 +1,5 @@
+SYSTEM_PROMPT = """You are evaluating the correctness of the Assistant's response. You are given a task and a candidate response. Is this a correct and accurate response to the task?
+
+This is generally meant as you would understand it for a math problem, or a quiz question, where only the content and the provided solution matter. Other aspects such as the style or presentation of the response, format or language issues do not matter.
+
```
poshinchen, line 4: qq: I notice other templates include scoring in the system prompt, but this one does not. Is this intentional?

src/strands_evals/evaluators/coherence_evaluator.py
```diff
@@ -74,13 +74,6 @@ def evaluate(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[Eva
         result = evaluator_agent(prompt, structured_output_model=CoherenceRating)
         return self._create_evaluation_output(result)
 
-    async def evaluate_async(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
```
poshinchen, line 77: If I understand it correctly, `evaluate_async` from the baseEvaluator will call the `evaluate()`, that's why you remove this function?

Have you tested it with example?

src/strands_evals/evaluators/correctness_evaluator.py
```diff
@@ -0,0 +1,164 @@
...
+
+    evaluation_level = EvaluationLevel.TRACE_LEVEL
+
+    _score_mapping = {
+        CorrectnessScore.PERFECTLY_CORRECT: 1.0,
+        CorrectnessScore.PARTIALLY_CORRECT: 0.5,
+        CorrectnessScore.INCORRECT: 0.0,
+    }
+
+    _reference_score_mapping = {
```
poshinchen, line 68: Why is the score different with / without `expected_*` data? Shouldn't the score be static?

src/strands_evals/evaluators/coherence_evaluator.py
ybdarrenwang: Yes this aligns with the intended refactoring.

I tested it with a coherence evaluator example script, checked in here: https://github.com/strands-agents/docs/pull/714/changes#diff-0f114303007c9d21daa526f3d49c4ceb0a49b54e20b5b37ee5295395e28f263b. It's executed successfully.

src/strands_evals/evaluators/correctness_evaluator.py
subbu10123: When we developed the evaluator for the "with reference" mode, the judgment for "partially correct" could not be made arbitrarily. Ideally, when part of the response is incorrect and part is correct, it should be classified as PARTIALLY_CORRECT. However, since there was a reference response to compare the agent response against, the judge models were having difficulty distinguishing between PARTIALLY_CORRECT and INCORRECT responses based on our experiments. Most classifications were falling under PARTIALLY_CORRECT even though important elements of the response that were expected were totally wrong, and they should have been classified as INCORRECT. So a decision was made to use only a binary category: even if one part of the response was incorrect, the total evaluation would be classified as INCORRECT. This problem arises only for the "with reference" mode. In the future, with the availability of more ground truth data for "partially correct" categories, this can be updated.

src/strands_evals/evaluators/prompt_templates/correctness/correctness_v0.py
ybdarrenwang: Yes. Since recent models have significantly better instruction-following capabilities and perform well with concise instructions, for example, for binary labels, simply specifying the labels in the output instruction works perfectly, no need for the more elaborate prompting patterns.

src/strands_evals/evaluators/correctness_evaluator.py
poshinchen: Is this behavior the same in agentcore too?

src/strands_evals/evaluators/correctness_evaluator.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class CorrectnessEvaluator(Evaluator[InputT, OutputT]):
+    """Evaluates the correctness of the assistant's response.
+
+    Supports two modes:
+    - **Basic mode**: Evaluates correctness based on the conversation context and the
+      assistant's response alone. Uses a 3-level scoring rubric
+      (Perfectly Correct=1.0, Partially Correct=0.5, Incorrect=0.0).
+    - **Reference mode**: When ``expected_output`` is provided on the evaluation case,
```
poshinchen: nit: double ``

poshinchen (commented):
🤔  I'm a bit confused about `expected_assertion`, and the `expected_output` now.
What is the difference? Can't we always use one or another in trace-based evaluators?

ybdarrenwang:
> 🤔 I'm a bit confused about `expected_assertion`, and the `expected_output` now. What is the difference? Can't we always use one or another in trace-based evaluators?

After discussion offline, we confirmed that it shouldn't make difference, and we'll use `expected_assertions` for both
- GoalSuccessRateEvaluator with assertion
- CorrectnessEvaluator with reference

poshinchen (approved):

PR #187, opened 2026-04-07
feat: add multimodal evaluators and prompt templates for image-to-text evaluation

sangminwoo:
## Description
This PR adds multimodal (MLLM-as-a-Judge) evaluators for image-to-text tasks with reference-free and reference-based evaluation support.

- Introduced `MultimodalCorrectnessEvaluator`, `MultimodalFaithfulnessEvaluator`, `MultimodalInstructionFollowingEvaluator`, and `MultimodalOverallQualityEvaluator` to assess various aspects of multimodal responses.
- Implemented `MultimodalOutputEvaluator` as a base class for handling multimodal inputs and outputs.
- Created prompt templates for evaluation rubrics including correctness, faithfulness, instruction following, and overall quality.
- Developed a structured approach for composing evaluation prompts with support for media content.
- Added `ImageData` class to manage image sources and formats, enabling flexible input handling for evaluators.
- Established a unified module for multimodal evaluation types and data structures.
- Supports reference-free evaluation (default) and reference-based evaluation (automatically toggled when `expected_output` is provided in the test case). 

## Related Issues

https://github.com/strands-agents/evals/issues/128

## Documentation PR

https://github.com/strands-agents/docs/pull/674

## Type of Change

New feature

## Testing

## Checklist

----


src/strands_evals/evaluators/multimodal_output_evaluator.py
```diff
@@ -0,0 +1,137 @@
...
+from ..types.evaluation import EvaluationData, EvaluationOutput
+from ..types.multimodal import MultimodalInput
+from .output_evaluator import OutputEvaluator
+from .prompt_templates.multimodal_case_prompt_template import compose_multimodal_test_prompt
+from .prompt_templates.multimodal_judge_system_prompt import MLLM_JUDGE_SYSTEM_PROMPT
+
+logger = logging.getLogger(__name__)
+
+
+class MultimodalOutputEvaluator(OutputEvaluator[MultimodalInput, str]):
```
jjbuck: `MultimodalOutputEvaluator` subclasses `OutputEvaluator` but completely overrides `evaluate()` and `evaluate_async()`. The only inherited behavior is         
  attribute storage in `__init__` and `to_dict()`. This doesn't provide strong substitutability behavior. `TrajectoryEvaluator` and `InteractionsEvaluator`   
  face the same situation and subclass `Evaluator` directly. This should too, or better yet, `OutputEvaluator` itself should be extended to handle 
  multimodal inputs.

src/strands_evals/evaluators/multimodal_output_evaluator.py
```diff
@@ -0,0 +1,137 @@
...
+        self.include_media = include_media
+        self.reference_suffix = reference_suffix if reference_suffix is not None else self.DEFAULT_REFERENCE_SUFFIX
+
+    def _select_rubric(self, evaluation_case: EvaluationData[MultimodalInput, str]) -> str:
+        """Select the appropriate rubric based on whether a reference output is available.
+
+        When expected_output is present, appends the reference comparison suffix to the rubric.
+        """
+        if evaluation_case.expected_output is not None:
+            return self.rubric + self.reference_suffix
```
jjbuck, line 64: Nit: It sounds like we're doing away with the reference suffix - is that right? Do we still need this `+ self.rerence_suffix` bit?

src/strands_evals/evaluators/multimodal_output_evaluator.py
```diff
@@ -0,0 +1,137 @@
...
+from ..types.multimodal import MultimodalInput
+from .output_evaluator import OutputEvaluator
+from .prompt_templates.multimodal_case_prompt_template import compose_multimodal_test_prompt
+from .prompt_templates.multimodal_judge_system_prompt import MLLM_JUDGE_SYSTEM_PROMPT
+
+logger = logging.getLogger(__name__)
+
+
+class MultimodalOutputEvaluator(OutputEvaluator[MultimodalInput, str]):
+    """MLLM-as-a-Judge evaluator for multimodal tasks.
```
jjbuck, line 16: I'd like to "stress test" the idea of whether this needs to be its own independent class at all. From the evaluator's perspective, an output is an output. The evaluator builds a prompt,     
  sends it to a judge, returns scores. Whether the prompt contains images is a detail of prompt construction, not a reason for a parallel class hierarchy.
  This design leaves `OutputEvaluator` as a vestigial text-only evaluator. The cleaner path would be to add a `_build_prompt()` template method to `OutputEvaluator` that  
  the concrete multimodal evaluators override. This would simplify things be removing the intermediate base class, the include_media flag, and getting us out of the slog of maintaining this parallel hierarchy in perpetuity.

src/strands_evals/evaluators/multimodal_output_evaluator.py
```diff
@@ -0,0 +1,137 @@
...
+
```
jjbuck, line 17: `MultimodalOutputEvaluator` hardcodes its generics as `OutputEvaluator[MultimodalInput, str]`. Every other evaluator in the SDK keeps `InputT, OutputT.`

src/strands_evals/types/multimodal.py
```diff
@@ -0,0 +1,263 @@
...
+# (VideoData, AudioData, PDFData, DocumentData) can be added to this union.
+AnyMediaData = Union[ImageData]
+
+
+# =============================================================================
+# MultimodalInput TypedDict
+# =============================================================================
+
+
+class MultimodalInput(TypedDict):
```
jjbuck: Every other input/output type in the SDK flows through Pydantic BaseModel (Case, EvaluationData, EvaluationOutput, ActorProfile, etc.). Using TypedDict means we don't get validation on construction. What's the rationale for this choice?

src/strands_evals/types/multimodal.py
```diff
@@ -0,0 +1,263 @@
...
+        elif PIL_AVAILABLE and isinstance(source_val, Image.Image):
+            buffer = BytesIO()
+            fmt = (self.format or "png").upper()
+            if fmt == "JPG":
+                fmt = "JPEG"
+            source_val.save(buffer, format=fmt)
+            source_val = base64.b64encode(buffer.getvalue()).decode("utf-8")
+        return {"source": source_val, "format": self.format, "media_type": self.media_type}
+
+    def to_bytes(self) -> bytes:
```
jjbuck, line 152: Just my two cents, but I think there's too much business logic buried in this method. It silently makes HTTP requests (urllib.request.urlopen), S3 API calls (boto3.client("s3").get_object), and file system reads. Network I/O buried inside a data model's method like this is hard to test, mock, and reason about. A user calling `image.to_bytes()` doesn't expect a 30-second HTTP
  timeout or an AWS credentials error. This should be a separate resolver/loader, not baked into the type.

src/strands_evals/types/multimodal.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def model_post_init(self, __context: Any) -> None:
+        """Auto-detect format and media_type after initialization."""
+        if self.format is None:
+            self.format = _detect_format(self.source)
+        if self.media_type is None and self.format:
+            self.media_type = f"image/{self.format}"
+
+    @model_serializer
+    def _serialize(self) -> dict:
```
jjbuck, line 134: Let's double-check that this works under all conditions. I think that because MultimodalInput is a TypedDict (not a BaseModel), Pydantic's serialization won't automatically invoke ImageData._serialize() when dumping a Case.

src/strands_evals/types/multimodal.py
```diff
@@ -0,0 +1,263 @@
...
+AnyMediaData = Union[ImageData]
+
+
+# =============================================================================
+# MultimodalInput TypedDict
+# =============================================================================
+
+
+class MultimodalInput(TypedDict):
+    """Input structure for multimodal evaluations.
```
jjbuck, line 234: Related to above, EvaluationReport.display() renders input as a string in the `Rich` table. A MultimodalInput dict with binary image data will render as garbled output. We need to handle this.

src/strands_evals/types/multimodal.py
```diff
@@ -0,0 +1,263 @@
+"""Multimodal evaluation types and data structures."""
+
+import base64
+import binascii
+import logging
+import re
+import urllib.request
+from io import BytesIO
+from pathlib import Path
+from typing import Any, List, Literal, Optional, Union, cast
```
jjbuck: @poshinchen @Unshure Do we prefer `typing_extensions` as the standard?

src/strands_evals/types/multimodal.py
```diff
@@ -0,0 +1,263 @@
...
+        ValueError: If the URI is malformed or the download fails.
+    """
+    if not BOTO3_AVAILABLE:
+        raise ImportError(f"boto3 is required to download from S3 URIs ({uri}). Install it with: pip install boto3")
+    match = _S3_URI_PARSE_PATTERN.match(uri)
+    if not match:
+        raise ValueError(f"Malformed S3 URI: {uri}. Expected format: s3://bucket/key")
+    bucket, key = match.group(1), match.group(2)
+    try:
+        s3 = boto3.client("s3")
```
jjbuck: We need to up-level this to a proper, full-blown configurable client (e.g., as written it will use the default region, which may not match the region of the target bucket). See also related comments on refactoring this to provide a proper resolve/dataloader for file fetching.

src/strands_evals/types/multimodal.py
poshinchen: most of these in typing should be replaced by the built-in syntax

src/strands_evals/evaluators/multimodal_output_evaluator.py
sangminwoo: Valid point. We propose adding a `_build_prompt()` template method to `OutputEvaluator`, so `MultimodalOutputEvaluator` only overrides how the prompt is built (not the full evaluation flow). We've prototyped this locally. since this modifies `OutputEvaluator` (shared code), we'd like to confirm alignment from you and @poshinchen before changing this.

src/strands_evals/evaluators/multimodal_output_evaluator.py
sangminwoo: We're keeping `reference_suffix`. it's actually the mechanism that replaced the old separate ref/non-ref rubric pairs. Previously we had e.g. CORRECTNESS_RUBRIC and CORRECTNESS_REF_RUBRIC as two separate strings. Now each evaluator has a single rubric, and `_select_rubric()` appends the `reference_suffix` when `expected_output` is provided. This avoids maintaining duplicate rubrics for every evaluation dimension.

src/strands_evals/evaluators/multimodal_output_evaluator.py
sangminwoo: we've implemented this locally (see [this Comment](https://github.com/strands-agents/evals/pull/187#discussion_r3048382290)). `OutputEvaluator` gets a `_build_prompt()` template method, and `MultimodalOutputEvaluator` only overrides it. no more duplicated `evaluate()`/`evaluate_async()`. Holding this change until we confirm alignment from you and @poshinchen  on modifying `OutputEvaluator`.

src/strands_evals/evaluators/multimodal_output_evaluator.py
sangminwoo: Fixed. Changed to `OutputEvaluator[InputT, OutputT]` throughout.

src/strands_evals/types/multimodal.py
sangminwoo: Converted `MultimodalInput` to `BaseModel` for consistency with other SDK types. This also resolves the serialization concern in [this Comment](https://github.com/strands-agents/evals/pull/187#discussion_r3048458079) Since `MultimodalInput` is now a `BaseModel` (like `ImageData`), calling `model_dump()` on a Case will call ImageData's custom serializer (e.g., converting raw bytes to base64 strings).

src/strands_evals/types/multimodal.py
sangminwoo: Moved the network/file download logic out of `ImageData.to_bytes()` into a standalone `resolve_image_bytes()` function. `to_bytes()` now just calls that function. This way, tests can mock `resolve_image_bytes()` without needing to deal with the `ImageData` model itself.

src/strands_evals/types/multimodal.py
sangminwoo: Resolved by converting `MultimodalInput` to `BaseModel` (see [this Comment](https://github.com/strands-agents/evals/pull/187#discussion_r3048449293)).

src/strands_evals/types/multimodal.py
sangminwoo: Added `_format_input_for_display()` to `EvaluationReport`. It detects multimodal inputs and renders a readable summary (instruction + media descriptor).

src/strands_evals/types/multimodal.py
sangminwoo: Updated. Replaced `List`, `Dict`, `Union`, `Optional` with built-in list, dict, |, X | None. Only `Any`, `Literal`, `cast` remain from typing.

src/strands_evals/types/multimodal.py
sangminwoo: Removed S3 handling entirely from this PR to keep it focused on multimodal evaluation. S3/custom data sources can be added in a follow-up with proper region configuration and a dedicated data loader.

src/strands_evals/evaluators/multimodal_output_evaluator.py
afarntrog: To @jjbuck's point, even with the `_build_prompt()` method, what is the purpose of inheriting from `OutputEvaluator` as opposed to inheriting directly from `Evaluator` since the core `OutputEvaluator` implementation is overridden? (I see the [latest commit](https://github.com/strands-agents/evals/pull/187/changes/4da410a28bcb544a7f895e4720de33508f779e13) has the `_build_prompt()` but I still see all the methods overridden.)

sangminwoo: Hi @afarntrog, good catch! I had already added the `_build_prompt()` template method to `OutputEvaluator` in the previous commit, but hadn't yet pushed the corresponding cleanup in `MultimodalOutputEvaluator` that removes the `evaluate()`/`evaluate_async()` overrides. Just pushed that now. `MultimodalOutputEvaluator` only overrides `_build_prompt()` and fully inherits the evaluation flow from `OutputEvaluator`. Could you take another look?

afarntrog: Nice! Can you just also resolve the merge conflicts?

afarntrog (changes requested):
I would like to see some tests for this new multimodal. Please add unit and integ tests

src/strands_evals/evaluators/multimodal_output_evaluator.py
sangminwoo: Sure, we will follow up shortly!

sangminwoo: Merge conflicts resolved.

sangminwoo:
@afarntrog Added unit and integration tests:
- `test_multimodal.py` - 45 tests for `ImageData`, `MultimodalInput`, `resolve_image_bytes`, format detection
- `test_multimodal_output_evaluator.py` - 18 tests for base evaluator init, rubric selection, prompt building, evaluate/evaluate_async
- `test_multimodal_specialized_evaluators.py` - 23 tests for all four specialized subclasses (Correctness, Faithfulness, InstructionFollowing, OverallQuality)
- `test_multimodal_case_prompt_template.py` - 16 tests for media resolution, content block building, prompt composition
- `test_evaluation_report.py` - 11 tests for `_format_input_for_display` and `flatten()`
- `tests_integ/test_multimodal_output_evaluator.py` - 9 integration tests covering MLLM/LLM modes, reference-based/free, multi-image, file path, sync/async, context, and batch

---

Also, could we add Sungyeon Kim (@sung-yeon-kim) as a collaborator on this repo? He's been co-authoring the work on this PR.

src/strands_evals/evaluators/multimodal_output_evaluator.py
```diff
@@ -0,0 +1,137 @@
+"""Multimodal output evaluator using MLLM-as-a-Judge."""
+
+import logging
+import warnings
+from typing import List, Union, cast
```
poshinchen: let's use built-in list / |

src/strands_evals/evaluators/multimodal_instruction_following_evaluator.py
```diff
@@ -0,0 +1,33 @@
+"""Instruction Following evaluator for multimodal tasks."""
+
+from strands.models.model import Model
+from typing_extensions import Union
```
poshinchen: let's use built-in list / |

src/strands_evals/evaluators/multimodal_faithfulness_evaluator.py
```diff
@@ -0,0 +1,33 @@
+"""Faithfulness evaluator for multimodal tasks."""
+
+from strands.models.model import Model
+from typing_extensions import Union
```
poshinchen: let's use built-in list / |

src/strands_evals/evaluators/multimodal_correctness_evaluator.py
```diff
@@ -0,0 +1,34 @@
+"""Correctness evaluator for multimodal tasks."""
+
+from strands.models.model import Model
+from typing_extensions import Union
```
poshinchen: let's use built-in list / |

src/strands_evals/types/multimodal.py
poshinchen: ? I'm still them here.

sangminwoo: @poshinchen This is already addressed in earlier commit (we used list / | and `Union` / `typing_extensions` are gone). I think the comment was left on the initial commit. Could you re-review against the latest commit?

src/strands_evals/evaluators/multimodal_correctness_evaluator.py
sangminwoo: @poshinchen  typing cleanup is already pushed in earlier commit. I think the comment was left on the previous commit, so the diff view was showing the old version. Could you re-review against the latest commit? Please let me know if anything else needs changes.

jjbuck:
### 1. Drop `include_media` entirely

As noted in one of my previous comments [https://github.com/strands-agents/evals/pull/187#discussion_r3048446548], I still think we should drop this boolean flag. Boolean flags on a class are typically an anti-pattern that make extensibility difficult - they almost always signal the class has two modes and should either be two classes, or the "mode" should be driven by the data. Here it's the latter. The class has a single coherent purpose, i.e., "evaluate a multimodal response with an MLLM judge, using content blocks when media is available." The flag re-introduces optionality that the `_build_prompt` refactor was specifically designed to eliminate.

This would touch all four specialized evaluators, `multimodal_output_evaluator.py`, `multimodal_case_prompt_template.py` Additionally, there's a silent corruption path here wherein `include_media: false` in an evaluator config can round-trip a multimodal case into silent text-only evaluation.

  **What to do:**

  1. Remove `include_media` from `MultimodalOutputEvaluator.__init__`, all four specialized subclasses, and `compose_multimodal_test_prompt`.
  2. Remove the `self.include_media` attribute.
  3. Make dispatch purely data-driven in `compose_multimodal_test_prompt`:
      ```python
      if isinstance(input_, MultimodalInput) and input_.media:
          # build content-block list
      else:
          # text-only prompt
          if isinstance(input_, MultimodalInput) and not input_.media:
              warnings.warn("MultimodalInput has empty media; evaluating as text-only.", ...)

Warn only when the input declares itself multimodal but has no media — that's an actual data inconsistency. Don't warn on plain-text inputs (since nothing is wrong with those)

  4. Remove the duplicated warning at the evaluator level (the prompt-composer is the right place for it.
  

### 2. Silent corruption on save/load round-trip

  File: src/strands_evals/evaluators/multimodal_output_evaluator.py, line ~85

  **Blocker: silent corruption on save/load round-trip.**

  `Experiment.from_dict` (in `experiment.py:662`) calls `Case.model_validate(case_data)` *without* the `[MultimodalInput, str]` parameterization.
  Pydantic has no way to know the original `Case` was parameterized, so `case.input` comes back as a raw `dict` rather than a `MultimodalInput`.
  Then in `_build_prompt`:

  ```python
  if (self.include_media
      and isinstance(evaluation_case.input, MultimodalInput)  # False — input is a dict
      and not evaluation_case.input.media):
```
  the isinstance guard returns False, and we fall through to text-only mode with no warning. You can verify this by saving a multimodal Case to JSON and reloading, where it gets evaluated as text only, with the image data preserved in the file but never reaching the judge. Users see plausible butwrong scores with no indication anything went wrong.

  What I'd recommend is

  1. In compose_multimodal_test_prompt, coerce the input before the isinstance dispatch:
```
      input_ = evaluation_case.input
      if isinstance(input_, dict):
          input_ = MultimodalInput.model_validate(input_)
```
 Do this at one well-defined boundary (the prompt composer) rather than scattering coercion through the evaluators.

  2. Add a round-trip regression test that saves a MultimodalInput-bearing experiment, reloads it, builds the prompt, and asserts the prompt is a list (content blocks), not a str. Asserting on prompt shape is the only way to catch silent fallthroughs like this. Asserting on round-trip JSON equality won't catch it because the JSON content is preserved.


  ### 3. `uses_environment_state` kwarg leak

  File: `src/strands_evals/evaluators/multimodal_output_evaluator.py`, `__init__`


  **Blocker: `uses_environment_state` kwarg leak breaks `from_dict` even with `custom_evaluators` passed.**

  `MultimodalOutputEvaluator.__init__` calls `super().__init__(...)` which sets `self.uses_environment_state = False`. But`MultimodalOutputEvaluator.__init__`'s own signature doesn't have `uses_environment_state`. `Evaluator.to_dict` compares `self.__dict__` against`inspect.signature(self.__class__.__init__)`'s defaults — the attribute isn't in that signature, so it's treated as a non-default value and always serialized.

The result is that `MultimodalCorrectnessEvaluator().to_dict()` emits `uses_environment_state: False`, and feeding that dict back through `from_dict` raises `TypeError: MultimodalCorrectnessEvaluator.init() got an unexpected keyword argument 'uses_environment_state'`

This fails even when the user passes `custom_evaluators=[MultimodalCorrectnessEvaluator]`. There's no user workaround today.

  **Fix:** surface `uses_environment_state` on `MultimodalOutputEvaluator.__init__` and forward it. Same for all four specialized subclasses — they need to accept it and pass it up, keeping the default `False`.

  ```python
  def __init__(
      self,
      rubric: str,
      model: Model | str | None = None,
      include_inputs: bool = True,
      system_prompt: str | None = None,
      reference_suffix: str | None = None,
      uses_environment_state: bool = False,
  ):
      super().__init__(
          rubric=rubric, model=model,
          system_prompt=system_prompt or MLLM_JUDGE_SYSTEM_PROMPT,
          include_inputs=include_inputs,
          uses_environment_state=uses_environment_state,
      )
      ...
```

sangminwoo:
Thanks @jjbuck for the detailed review and suggestions!
                                                                                                                                                                                                                       
### 1. Drop `include_media` entirely

Dispatch is now data-driven in `compose_multimodal_test_prompt`:
- `MultimodalInput` with `media` → MLLM mode
- `MultimodalInput` with empty `media` → LLM mode + warning
- Plain text input → LLM mode, no warning (if someone wants text-only judging, they pass plain-text input instead of a `MultimodalInput`)

Removed `include_media` from `MultimodalOutputEvaluator`, the four specialized subclasses, and `compose_multimodal_test_prompt`, plus the duplicated warning at the evaluator level.

### 2. Silent corruption on save/load round-trip
The fix follows your suggestion: coerce the input in `compose_multimodal_test_prompt`, so there's one place handling it. Also added a regression test: it saves a `MultimodalInput`, reloads it as a raw dict (simulating the `from_dict` path), and checks the built prompt is a `list` with an `image` block, not a plain string. This way, if the coercion ever breaks again, the test will catch it.

### 3. `uses_environment_state` kwarg leak
As you suggested, I've added `uses_environment_state: bool = False` to `MultimodalOutputEvaluator.__init__` and all four subclasses, forwarded to `super().__init__`.

While fixing this, I noticed `reference_suffix` had the same issue. Fixed it in all four subclasses and added a round-trip test covering all four evaluators.

Thanks again!

jjbuck (approved):

poshinchen (approved):

afarntrog (approved):

PR #189, opened 2026-04-08
feat(detectors): detectors interface and failure_detector implementation

poshinchen:
## Description
Here's the Phase 1 to develop detectors (failure_detector).
Summary
  1. **Add strands_evals.detectors package**
    1.1. add single-trace detection capabilities into strands-evals
  2. **Implement detect_failures() for LLM-based semantic failure detection (hallucinations, tool misuse, goal deviation, etc.) with automatic chunking fallback for sessions exceeding context limits**
  3. **Add shared chunking utilities** (token estimation, span splitting, chunk merge) used by all LLM-based detectors
  4. **Add all Pydantic types for the detectors package** (FailureOutput, RCAOutput, DiagnosisResult, and LLM structured output schemas)

### What's included:
- `types/detector`
  - All detector Pydantic models (output types + LLM schemas)
- `detectors/chunking.py`
  - Shared context window management and span chunking
  - Pure-parsing user request extraction (no LLM)
- `detectors/failure_detector.py`
  - detect_failures() with direct → chunked fallback
- `detectors/prompt_templates/failure_detection/`
  - Failure detection prompt
- `detectors/__init__.py`
  - Public exports

### How to Use failure_detector
```
memory_exporter.clear()
create_agents_and_graph(case)
finished_spans = memory_exporter.get_finished_spans()
mapper = StrandsInMemorySessionMapper()
high_conf = detect_failures(flawed_session, confidence_threshold="medium")
```

### Test Scenarios
  - Unit tests for all Pydantic types (test_types.py)
  - Unit tests for chunking utilities (test_chunking.py)
  - Unit tests for user request extraction (test_user_request_extractor.py)
  - Unit tests for failure detector with mocked LLM (test_failure_detector.py)
  - Manual integration test: ran detect_failures() on a flawed multi-agent graph session — correctly detected goal deviation, tool schema errors, and task non-compliance
  - Manual integration test: ran detect_failures() on a clean math agent session

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/detectors/chunking.py
+++ b/src/strands_evals/detectors/chunking.py
@@ -1,0 +1,125 @@
+"""Shared utilities for context window management and span chunking.
+
+Used by detect_failures, summarize_execution, and analyze_root_cause
+to handle sessions that exceed LLM context limits.
+
+Ported from AgentCoreLens failure_detector._split_spans_by_tokens() and
+related helpers, replacing litellm.token_counter() with a conservative
+character-based estimate.
+"""
+
+import json
+import logging
+
+from ..types.detector import ConfidenceLevel, FailureItem
+from ..types.trace import SpanUnion
+from .constants import (
+    CHARS_PER_TOKEN,
+    CHUNK_OVERLAP_SPANS,
+    CONTEXT_SAFETY_MARGIN,
+    DEFAULT_MAX_INPUT_TOKENS,
+    MIN_CHUNK_SIZE,
+)
+
+logger = logging.getLogger(__name__)
+_CONFIDENCE_RANK: dict[ConfidenceLevel, int] = {"low": 0, "medium": 1, "high": 2}
+
+
+def estimate_tokens(text: str) -> int:
+    """Conservative token estimate from text length."""
+    return len(text) // CHARS_PER_TOKEN
+
+
+def would_exceed_context(prompt_text: str, max_input_tokens: int = DEFAULT_MAX_INPUT_TOKENS) -> bool:
+    """Pre-flight check: would this prompt likely exceed context?
+
+    If wrong (false negative), the model returns a context-exceeded error,
+    which triggers the chunking fallback in each detector.
+    """
+    return estimate_tokens(prompt_text) > int(max_input_tokens * CONTEXT_SAFETY_MARGIN)
+
+
+def split_spans_by_tokens(
+    spans: list[SpanUnion],
+    max_tokens: int = DEFAULT_MAX_INPUT_TOKENS,
+    overlap_spans: int = CHUNK_OVERLAP_SPANS,
+) -> list[list[SpanUnion]]:
+    """Split spans into chunks fitting within token limits.
+
+    Adjacent chunks share ``overlap_spans`` spans for context continuity.
+    Ported from Lens ``failure_detector._split_spans_by_tokens()``.
+
+    Args:
+        spans: Flat list of span objects to chunk.
+        max_tokens: Maximum model input tokens.
+        overlap_spans: Number of spans shared between adjacent chunks.
+
+    Returns:
+        List of span chunks. Each chunk fits within the effective token limit.
+    """
+    effective_limit = int(max_tokens * CONTEXT_SAFETY_MARGIN)
+    chunks: list[list[SpanUnion]] = []
+    current_chunk: list[SpanUnion] = []
+    current_tokens = 0
+
+    for span in spans:
+        span_tokens = estimate_tokens(json.dumps(span.model_dump(), default=str))
+        if current_tokens + span_tokens > effective_limit and len(current_chunk) >= MIN_CHUNK_SIZE:
+            chunks.append(current_chunk)
+            overlap = current_chunk[-overlap_spans:] if overlap_spans > 0 else []
+            current_chunk = list(overlap)
+            current_tokens = sum(estimate_tokens(json.dumps(s.model_dump(), default=str)) for s in overlap)
+        current_chunk.append(span)
+        current_tokens += span_tokens
+
+    if current_chunk:
+        chunks.append(current_chunk)
+
+    logger.info(
+        "Split %d spans into %d chunks (max_tokens=%d, overlap=%d)",
+        len(spans),
+        len(chunks),
+        max_tokens,
+        overlap_spans,
+    )
+    return chunks
+
+
+def merge_chunk_failures(chunk_results: list[list[FailureItem]]) -> list[FailureItem]:
+    """Merge failures from overlapping chunks, deduplicating by span_id.
+
+    Keeps highest confidence per category when the same span_id appears
+    in multiple chunks. Ported from Lens ``_merge_chunk_failures()``.
+
+    Args:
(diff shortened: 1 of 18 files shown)


src/strands_evals/detectors/constants.py
```diff
@@ -0,0 +1,22 @@
+"""Shared constants for detectors."""
+
+# Default model for all detectors
+DEFAULT_DETECTOR_MODEL = "us.anthropic.claude-sonnet-4-5-20250929-v1:0"
```
afarntrog, line 4: I'm a bit hesitant to hardcode a default model. We will constantly have to maintain it. Why not use the same model that is being used for the source agent run?

src/strands_evals/detectors/chunking.py
```diff
@@ -0,0 +1,267 @@
...
+def estimate_tokens(text: str) -> int:
+    """Estimate token count for a text string.
+
+    Uses tiktoken (cl100k_base via strands SDK) when available for accurate
+    counts. Falls back to ``len(text) // CHARS_PER_TOKEN`` otherwise.
+    """
+    enc = _get_tiktoken_encoding()
+    if enc is not None:
+        return len(enc.encode(text))
+    return len(text) // CHARS_PER_TOKEN
```
afarntrog: This integer division `len("x") // 3` returns 0 for any string shorter than 3 characters are we willing to accept that? if not we should use something like `max`

src/strands_evals/detectors/failure_detector.py
```diff
@@ -0,0 +1,295 @@
...
+
+from .._async import run_async
+from ..types.detector import ConfidenceLevel, FailureDetectionStructuredOutput, FailureItem, FailureOutput
+from ..types.trace import Session, SpanUnion
+from .chunking import merge_chunk_failures, split_spans_by_tokens, would_exceed_context
+from .constants import DEFAULT_DETECTOR_MODEL
+from .prompt_templates.failure_detection import get_template
+
+logger = logging.getLogger(__name__)
+CONFIDENCE_ORDER: dict[ConfidenceLevel, int] = {"low": 0, "medium": 1, "high": 2}
```
afarntrog: This is duplicate of the other one in this PR. Can we extract them into a `constants.py` file?
```
chunking.py:27
_CONFIDENCE_RANK: dict[ConfidenceLevel, int] = {"low": 0, "medium": 1, "high": 2}
```

src/strands_evals/types/detector.py
```diff
@@ -0,0 +1,153 @@
...
+    model_config = {"populate_by_name": True}
+
+    fix_type: Literal["SYSTEM_PROMPT_FIX", "TOOL_DESCRIPTION_FIX", "OTHERS"] = Field(
+        ..., alias="Fix Type", description="Type of fix needed"
+    )
+    recommendation: str = Field(
+        ...,
+        alias="Recommendation",
+        description="Brief, actionable fix suggestion (1-2 sentences)",
+    )
```
afarntrog: I don't see this being used. Let's remove it and add it when we need it.

tests/strands_evals/detectors/test_types.py
```diff
@@ -0,0 +1,196 @@
...
+    FailureDetectionStructuredOutput,
+    FailureError,
+    FailureItem,
+    FailureOutput,
+    FixRecommendation,
+    RCAItem,
+    RCAOutput,
+    RCAStructuredOutput,
+    RootCauseItem,
+)
```
afarntrog, line 8: Some of these are not being used. Let's remove it and add it when we need it.

afarntrog (changes requested):

src/strands_evals/detectors/constants.py
poshinchen: Personally I would prefer to allow users to configure it aside from agent models. Users can always use an advanced model for agent run and evaluations / detection with a bit cheaper model.

src/strands_evals/detectors/chunking.py
poshinchen: This is okay, the estimated token is used to decide whether to chunk the context or not.

src/strands_evals/detectors/chunking.py
```diff
@@ -0,0 +1,266 @@
...
+        try:
+            from strands.models.model import _get_encoding
+
+            _tiktoken_encoding = _get_encoding()
+            _tiktoken_available = True
+            logger.debug("tiktoken encoding loaded via strands SDK")
+        except (ImportError, Exception) as exc:
+            _tiktoken_available = False
+            logger.info("tiktoken unavailable, using char-based fallback: %s", exc)
+    return _tiktoken_encoding
```
afarntrog, line 44: Why are we being so defensive here? Is it not installed as a dependency and if not why not?

src/strands_evals/detectors/failure_detector.py
```diff
@@ -0,0 +1,314 @@
...
+    async def _stream() -> str:
+        chunks: list[str] = []
+        async for event in model.stream(messages):
+            if "contentBlockDelta" in event:
+                delta = event["contentBlockDelta"].get("delta", {})
+                if "text" in delta:
+                    chunks.append(delta["text"])
+        return "".join(chunks)
+
+    return run_async(_stream)
```
afarntrog, line 185: Is there data backing this up that combining the `system_prompt` together with the `user_prompt` performs better? If so, when would it ever be appropriate to use a system prompt?

src/strands_evals/detectors/failure_detector.py
```diff
@@ -0,0 +1,314 @@
...
+        trace_dict = {
+            "trace_id": trace_id,
+            "session_id": trace_spans[0].span_info.session_id,
+            "spans": [s.model_dump() for s in trace_spans],
+        }
+        traces.append(trace_dict)
+    return traces
+
+
+def _serialize_spans(spans: list[SpanUnion], session_id: str) -> str:
```
afarntrog: do we need the `session_id: str)`? It doesn't seem to be used

src/strands_evals/detectors/failure_detector.py
```diff
@@ -0,0 +1,314 @@
...
+import logging
+import re
+from collections import OrderedDict
+
+from pydantic import ValidationError
+from strands.models.bedrock import BedrockModel
+from strands.models.model import Model
+from strands.types.content import ContentBlock, Message, Messages
+from strands.types.exceptions import ContextWindowOverflowException
+from typing_extensions import Union
```
afarntrog: Let's use `|` instead of `Union`

afarntrog (changes requested):

src/strands_evals/detectors/chunking.py
poshinchen: correct, tiktoken is not installed by default in Strands since some model_providers has their own token_counter and strands counts the token in the order of: `model_provider > tiktoken > chars//3`.

Changing the token_counter will impact the performances, that's why we're checking tiktoken. However, we could force the tiktoken to be installed though in strands-evals.

afarntrog (approved):

PR #207, opened 2026-04-30
feat(simulator): structured_output for ActorSimulator

poshinchen:
## Summary

Adds structured stop-signalling and custom output schema support to `ActorSimulator`, replacing the `<stop/>` text sentinel with a proper `stop: bool` field on the structured response.

## Breaking Changes

- **`ActorResponse.message`** changed from `str` (required) to `str | None` (optional, default `None`)
- **Stop detection** uses `ActorResponse.stop == True` instead of parsing `<stop/>` from message text
- **`ActorResponse` fields added**: `stop: bool` and `stop_reason: str | None`

## Key Changes

- `__init__` accepts keyword-only `structured_output_model: type[BaseModel] | None`, validated at construction and used as the default for all `act()` calls
- `act()` accepts per-call `structured_output_model` override; any Pydantic `BaseModel` with `message` and `stop` fields.
- Simulator populates `stop_reason`: `"goal_completed"` or `"max_turns"`
- `system_prompt_template` defaults to `DEFAULT_USER_SIMULATOR_PROMPT_TEMPLATE` and supports pre-rendered strings (no `{actor_profile}` placeholder required)
- Prompt template instructs "set `stop=true`" instead of "generate `<stop/>`"

### `types/simulation/actor.py`
- `ActorResponse.message`: `str` → `str | None = None`
- `ActorResponse` gains `stop: bool = False` and `stop_reason: str | None = None`

### `actor_simulator.py`
- `structured_output_model` kwarg on `__init__`, validated via `_validate_output_model()` (checks `message` and `stop` fields exist)
- `act()` resolution order: per-call → init-level → `ActorResponse`
- Stop logic reads `stop` from structured output, sets `stop_reason`, enforces `max_turns`
- `has_next()` simplified to `return not self.stop`
- Removed `_last_message` tracking and `<stop/>` sentinel parsing

### `prompt_templates/actor_system_prompt.py`
- Exit conditions: "set `stop=true` in your structured response" replaces "generate `<stop/>`"
- Fixed typo: "Do no deviate" → "Do not deviate"

## Usage

```python
from strands_evals.simulation import ActorSimulator
from pydantic import BaseModel
from strands_evals.types.simulation import ActorProfile

# Any BaseModel with `message` and `stop` fields works
class AgentInput(BaseModel):
    reasoning: str = ""
    stop: bool = False
    message: str | None = None
    urgency: str = "normal"

profile = ActorProfile(
    traits={"expertise_level": "beginner", "communication_style": "casual"},
    context="A user trying to debug a production outage.",
    actor_goal="Get help identifying the root cause of a memory leak.",
)

simulator = ActorSimulator(
    actor_profile=profile,
    initial_query="Our service is running out of memory in prod",
    structured_output_model=AgentInput,
    max_turns=10,
)

agent = Agent(system_prompt="You are an SRE assistant.")

user_message = simulator.initial_query
while simulator.has_next():
    agent_response = agent(user_message)
    result = simulator.act(str(agent_response))
    if result.structured_output.stop:
        break
    user_message = result.structured_output.message
```

Default usage (no custom model):
```python
simulator = ActorSimulator(
    actor_profile=profile,
    initial_query="Help me debug this",
    max_turns=10,
)

while simulator.has_next():
    agent_response = agent(user_message)
    result = simulator.act(str(agent_response))
    user_message = str(result.structured_output.message)
```

## Testing

25 unit tests covering init-level and per-call `structured_output_model`, field validation (`message` and `stop` required), `stop_reason` values, custom model stop management, template selection, and default prompt `stop=true` instruction.

## Type of Change

Breaking change + new feature

## Checklist
- [x] I have added tests that prove my fix is effective or my feature works

----

### DIFF
--- a/src/strands_evals/simulation/actor_simulator.py
+++ b/src/strands_evals/simulation/actor_simulator.py
@@ -1,6 +1,7 @@
 import logging
 import random
 
+from pydantic import BaseModel, Field, create_model
 from strands import Agent
 from strands.agent.agent_result import AgentResult
 from strands.types.content import Message
@@ -11,7 +12,7 @@ from strands_evals.simulation.profiles.actor_profile import DEFAULT_USER_PROFILE
 from strands_evals.simulation.prompt_templates.actor_profile_extraction import ACTOR_PROFILE_PROMPT_TEMPLATE
 from strands_evals.simulation.prompt_templates.actor_system_prompt import DEFAULT_USER_SIMULATOR_PROMPT_TEMPLATE
 from strands_evals.simulation.tools.goal_completion import get_conversation_goal_completion
-from strands_evals.types.simulation import ActorProfile, ActorResponse
+from strands_evals.types.simulation import ActorProfile, ActorResponse, SimulatorResult
 
 logger = logging.getLogger(__name__)
 
@@ -138,10 +139,12 @@ class ActorSimulator:
         self,
         actor_profile: ActorProfile,
         initial_query: str,
-        system_prompt_template: str,
+        system_prompt_template: str = DEFAULT_USER_SIMULATOR_PROMPT_TEMPLATE,
         tools: list | None = None,
         model: str | None = None,
         max_turns: int = 10,
+        *,
+        input_type: type[BaseModel] | None = None,
     ):
         """
         Initialize an ActorSimulator with profile and goal.
@@ -152,10 +155,27 @@ class ActorSimulator:
         Args:
             actor_profile: ActorProfile object containing traits, context, and actor_goal.
             initial_query: The actor's first query or message.
-            system_prompt_template: Template string for system prompt. Must include {actor_profile} placeholder.
+            system_prompt_template: System prompt for the actor. Accepts two shapes:
+
+                - A template containing the `{actor_profile}` placeholder, which
+                  is rendered via `str.format(actor_profile=...)` against the
+                  actor's profile (legacy behavior, the default).
+                - An already-rendered system prompt string with no
+                  `{actor_profile}` placeholder, which is used verbatim.
+
+                Defaults to the built-in `DEFAULT_USER_SIMULATOR_PROMPT_TEMPLATE`.
             tools: Additional tools available to the actor. Defaults to goal completion tool only.
             model: Model identifier for the underlying agent. Uses Strands default if None.
             max_turns: Maximum number of conversation turns before stopping (default: 10).
+            input_type: Pydantic model class describing the agent-under-test's expected
+                input payload. Only affects :meth:`act_structured` — the LLM's
+                structured-output schema is narrowed so `message` is produced as an
+                `input_type` instance. :meth:`act` is unaffected and always uses the
+                legacy :class:`ActorResponse` schema.
+
+                Must describe what the agent under test accepts — not a simulator
+                type. Passing :class:`SimulatorResult` (or a subclass) raises
+                `ValueError` at construction time.
 
         Example:
             ```python
@@ -188,8 +208,13 @@ class ActorSimulator:
         self._turn_count = 0
         self._last_message = ""
         self._max_turns = max_turns
+        self._input_type = input_type
+        self._structured_model = self._build_structured_model(input_type)
 
-        system_prompt = system_prompt_template.format(actor_profile=actor_profile.model_dump())
+        if "{actor_profile}" in system_prompt_template:
+            system_prompt = system_prompt_template.format(actor_profile=actor_profile.model_dump())
+        else:
+            system_prompt = system_prompt_template  # already rendered
 
         # Combine tools
         all_tools = [get_conversation_goal_completion]
@@ -205,6 +230,46 @@ class ActorSimulator:
             tools=all_tools,
             model=self.model_id,
             callback_handler=None,
+        )
+
+    def _build_structured_model(self, input_type: type[BaseModel] | None) -> type[SimulatorResult]:
+        """Return the :class:`SimulatorResult` subclass used by :meth:`act_structured`.
+
+        When `input_type` is `None`, returns :class:`SimulatorResult` itself —
+        `message` stays typed as `Any` so the LLM is free to produce a string.
+        When `input_type` is set, returns a dynamic subclass that narrows
+        `message` to `input_type | None` so the LLM's tool-use schema enforces
+        the caller's agent-input shape.
+
+        `input_type` is rejected if it is :class:`SimulatorResult` or a subclass
+        of it — that's the one nesting case where the outer simulator envelope
+        (`reasoning`, `stop`, `message`) is duplicated inside the payload
+        and the LLM's schema becomes ambiguous.
+        """
+        if input_type is None:
+            return SimulatorResult
+
+        if isinstance(input_type, type) and issubclass(input_type, SimulatorResult):
+            raise ValueError(
+                "input_type must describe the agent-under-test's input schema, not "
+                "SimulatorResult (or a subclass). SimulatorResult is the simulator's "
+                "return envelope and cannot be used as an agent payload."
+            )
+
+        return create_model(
+            "SimulatorResult",
+            __base__=SimulatorResult,
+            message=(
+                input_type | None,
+                Field(
+                    None,
+                    description=(
+                        f"Structured message matching the agent's input schema "
+                        f"({input_type.__name__}). Provide when stop=false. Set "
+                        "to null when stop=true."
+                    ),
+                ),
+            ),
         )
 
     def _initialize_conversation(self):
@@ -232,13 +297,18 @@ class ActorSimulator:
         profile and goal. The response includes reasoning about the actor's thought
         process and the actual message to send.
 
+        This method uses the legacy :class:`ActorResponse` schema and returns the
+        raw Strands :class:`AgentResult`. It is preserved for backwards compatibility
+        with existing callers. The `input_type` kwarg on `__init__` does **not**
+        affect this method — use :meth:`act_structured` to consume `input_type`.
+
         Args:
             agent_message: The agent's response to react to (required).
 
         Returns:
             AgentResult containing the actor's structured response with:
                 - structured_output.reasoning: Actor's internal reasoning
-                - structured_output.message: Actor's response message
+                - structured_output.message: Actor's response message (str)
 
         Example:
             ```python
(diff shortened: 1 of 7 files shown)


src/strands_evals/types/simulation/actor.py
```diff
@@ -32,3 +37,53 @@ class ActorResponse(BaseModel):
 
     reasoning: str = Field(..., description="Reasoning for the actor's response")
     message: str = Field(..., description="Message from the actor")
+
+
+class SimulatorResult(BaseModel):
```
padmak30: nit: `ActorStructuredResponse` to keep it inline with ActorResponse

padmak30 (dismissed):

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -207,6 +257,46 @@ def __init__(
...
+            return ActorStructuredResponse
+
+        if isinstance(input_type, type) and issubclass(input_type, ActorStructuredResponse):
+            raise ValueError(
+                "input_type must describe the agent-under-test's input schema, not "
+                "ActorStructuredResponse (or a subclass). ActorStructuredResponse is the simulator's "
+                "return envelope and cannot be used as an agent payload."
+            )
+
+        return create_model(
```
jjbuck: There's a risk of class name collision here.

`create_model("ActorStructuredResponse", __base__=ActorStructuredResponse, ...)` names the subclass identically to its base. Pydantic JSON-schema generators key by __name__ in $defs, and collisions with the base in the same process can produce ambiguous schema references. I'd say let's name it something like `f"ActorStructuredResponse_{input_type.__name__}"` to disambiguate.

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -207,6 +257,46 @@ def __init__(
             callback_handler=None,
         )
 
+    def _build_structured_model(self, input_type: type[BaseModel] | None) -> type[ActorStructuredResponse]:
```
jjbuck: Nit: Since this doesn't use `self` anywhere in the business logic, should be a static method or a module-level helper.

tests/strands_evals/simulation/test_actor_simulator.py
```diff
@@ -211,3 +212,276 @@ def test_has_next_detects_stop_token(sample_actor_profile):
...
+    mock_response.structured_output = ActorStructuredResponse(reasoning="r", stop=False, message="hi")
+    simulator.agent = MagicMock(return_value=mock_response)
+
+    simulator.act_structured("agent reply")
+
+    assert simulator.agent.call_args[1]["structured_output_model"] is ActorStructuredResponse
+
+
+def test_init_with_input_type_narrows_message_schema(sample_actor_profile):
+    """With input_type, act_structured() hands a ActorStructuredResponse subclass whose message is typed."""
```
jjbuck: Nit: "*an* ActorStructuredResponse"

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -188,8 +220,26 @@ def __init__(
         self._turn_count = 0
         self._last_message = ""
         self._max_turns = max_turns
+        self._input_type = input_type
+        self._structured_model = self._build_structured_model(input_type)
```
jjbuck: `self._structured_model` shouldn't be instance state — the output model is a property of the return type, not the simulator. It only exists here because `create_model(__base__=...)` returns a new class on every call, so you have to cache it.

  Use Pydantic generics instead:

      class ActorStructuredResponse(BaseModel, Generic[MessageT]):
          reasoning: str
          stop: bool = False
          message: MessageT | None = None

  Then `ActorStructuredResponse[input_type]` is a cached, canonical parametrization. The whole `_build_structured_model` method and this attribute collapse to an inline expression at the call site:

      schema = ActorStructuredResponse[self._input_type] if self._input_type else ActorStructuredResponse
      response = self.agent(msg, structured_output_model=schema)

  Side benefits: no `__name__` collision between the base and the dynamic subclass, proper static typing for callers, and `isinstance(result, ActorStructuredResponse)` keeps working.

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -261,6 +358,61 @@ def act(self, agent_message: str) -> AgentResult:
         self._last_message = str(cast(ActorResponse, response.structured_output).message)
         return response
 
+    def act_structured(self, agent_message: str) -> ActorStructuredResponse:
```
jjbuck: I think there's a subtle bug  in `act_structured()`that silently breaks `has_next()`.

  `has_next()` only returns False via `"<stop/>" not in self._last_message` or the `max_turns` backstop. On the structured path the LLM sets `stop=true` on the response instead of emitting the `<stop/>` sentinel, so `has_next()` keeps returning True until `max_turns` trips. So concretely, the idiomatic loop
```
  while sim.has_next():
      result = sim.act_structured(str(agent(msg)))
      msg = result.message
```
  runs to `max_turns` every time, even when the actor signalled completion earlier. The example given in the PR text loops over `range(simulator._max_turns)` instead — not sure if that was intentional or not. I'd go for an approach that maintains the ergonomics of `has_next()` so that users don't have to reach into a private attribute (`_max_turns`).

Possible implementation - store the last result on the simulator and consult it from `has_next()`.
```
  # __init__
  self._last_result: ActorStructuredResponse | None = None

  # end of act_structured()
  self._last_result = result

  # has_next()
  if self._turn_count >= self._max_turns:
      return False
  if self._last_result is not None and self._last_result.stop:
      return False
  return "<stop/>" not in self._last_message
```
  Storing the full result (rather than a `_stopped: bool`) keeps `stop_reason` and future fields available to `has_next()` without another round of plumbing. 

In the longer term, converging `act()` and `act_structured()` into one method would be a good goal for v1.0.

jjbuck:
I want to pitch a slightly different shape for this feature. The issue is that `input_type` as currently designed conflates three decisions onto one kwarg: prompt template selection, response schema construction, and which method (`act` vs `act_structured`) the caller is supposed to use. None of the coordination is enforced, so several combinations can run but produce broken conversations — e.g. `input_type=AgentInput` + caller uses `act()` installs the structured prompt (which tells the LLM to set `stop=true`) but the `ActorResponse` schema has no `stop` field, so the actor can't terminate. 

Instead, if we wrote....
```
def __init__(
    self,
    ...,
    *,
    output_type: type[BaseModel] | None = None,
):
```
....then this would have the following behavior.
1. When `output_type` is None, the simulator behaves exactly as it does today. `ActorResponse` is the schema, the sentinel template is the prompt, `<stop/>` terminates the conversation, and `act()` returns the same `AgentResult` shape that existing callers already consume.
2. When `output_type` is set, `act()` internally uses `ActorStructuredResponse[output_type]` as the schema and the structured template as the prompt. The actor terminates by setting `stop=true` on the response. The return type is still `AgentResult`; what differs is the class the caller finds on `structured_output`.

This collapses the two methods into one because passing a Pydantic class is already an unambiguous request for structured output. There's nothing additional for a flag to disambiguate. A caller who doesn't want the new behavior doesn't pass the kwarg, and everything works as it does today. A caller who does want it passes the kwarg once at construction and uses `act()` normally from then on.

So concretely, this would mean
1. `__init__` gains `output_type` and drops `input_type`.
2. `act()` branches internally on `self._output_type` to pick the schema.
3. `ActorStructuredResponse` becomes a Pydantic generic parameterized on the message type (separate comment on `_structured_model` goes deeper).
4. `act_structured()` and `_build_structured_model()` are deleted.

jjbuck (changes requested):

src/strands_evals/simulation/actor_simulator.py
poshinchen: decided to merge the methods into act()!

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -138,67 +124,99 @@ def __init__(
         self,
         actor_profile: ActorProfile,
         initial_query: str,
```
jjbuck, line 126: We can defer this to a future CR (perhaps slate it for v1.0), but for future-proofing I'd like to propose that we change the type from `str` to `strands.types.agent.AgentInput`.

This is the real first-turn message handed to the agent under test. Typing it as AgentInput (imported from strands.types.agent) matches what Strands itself accepts and means we inherit future extensions (e.g., new content block types, interrupts, etc.) without a change here.

src/strands_evals/simulation/actor_simulator.py
```diff
@@ -138,67 +124,99 @@ def __init__(
...
+                Can still be overridden per-call via `act(structured_output_model=...)`.
 
         Example:
             ```python
             from strands_evals.simulation import ActorSimulator
+            from pydantic import BaseModel
             from strands_evals.types.simulation import ActorProfile
 
-            # Define custom actor profile
+            class AgentInput(BaseModel):
```
jjbuck: This isn't really the "input" is it? This is more like the output of the actor simulator, right? And also, see other comment above that I think touches on why I'm getting a little confused here. It has the has the same name as the Strands `AgentInput` abstraction but is a different type.

src/strands_evals/types/simulation/actor.py
```diff
@@ -22,13 +22,33 @@ class ActorProfile(BaseModel):
 
 
 class ActorResponse(BaseModel):
```
jjbuck, line 24: Building on the observations around the `AgentInput` type from Strands, I want to propose that we also rework ActorResponse to use
`message: Message` from Strands. 

This would also mean getting rid of the `reasoning` property on `ActorResponse`, because the Strands `Message` type provides reasoning in the `reasoningContent` blocks in `message.content`.

poshinchen: Good callout, it is returned in `act() -> AgentResult` so user can get the message already. I think we still need a structured_output which returns a response easily. I need to think if it's a breaking change though.

src/strands_evals/simulation/actor_simulator.py
poshinchen: This makes sense!

poshinchen: Will investigate this in the next PR before v1

jjbuck (approved):

padmak30 (approved):

PR #210, opened 2026-05-01
feat(detectors): integrated rca into evaluation workflow

poshinchen:
## Summary

Integrates the detection pipeline (`detect_failures` → `analyze_root_cause`) into the `Experiment` evaluation workflow so that failing cases are automatically diagnosed and fix recommendations surface directly in the report.

### What's new

- **`diagnose_session()`** — single-call pipeline that runs failure detection followed by root cause analysis on a `Session`. Returns a `DiagnosisResult` with failures, root causes, and deduplicated recommendations.
- **`DiagnosisResult`** — Pydantic model with a `recommendations` property that extracts and deduplicates fix strings from all root causes.
- **`DiagnosisConfig`** — groups all diagnosis-related parameters into a single config object:
  - `trigger: DiagnosisTrigger = "on_failure"` — when to run diagnosis (`"on_failure"` or `"always"`)
  - `model: Model | str | None = None` — override the LLM used for diagnosis
  - `confidence_threshold: ConfidenceLevel = "medium"` — minimum confidence for failure detection
- **`Experiment` integration** — accepts an optional `diagnosis_config: DiagnosisConfig` parameter. When `None` (the default), diagnosis is disabled. When provided, diagnosis runs according to the trigger setting.
- **`EvaluationReport`** — two new fields: `diagnoses: list[dict | None]` and `recommendations: list[str | None]`, populated per-case (shared across evaluator reports).
- Diagnosis runs once per failing case (not per evaluator) to avoid duplicate LLM calls, then results are shared across all evaluator reports for that case.

### Usage

**Standalone diagnosis:**
```python
from strands_evals.detectors import diagnose_session

result = diagnose_session(session, confidence_threshold="low")

for rc in result.root_causes:
    print(f"{rc.root_cause_explanation}")
    print(f"  Fix ({rc.fix_type}): {rc.fix_recommendation}")

for rec in result.recommendations:
    print(rec)
```

**Integrated with Experiment:**
```python
from strands_evals import DiagnosisConfig, Experiment

# Diagnose on failure (default trigger)
experiment = Experiment(
    cases=test_cases,
    evaluators=[evaluator],
    diagnosis_config=DiagnosisConfig(),
)

# Always diagnose, with custom model
experiment = Experiment(
    cases=test_cases,
    evaluators=[evaluator],
    diagnosis_config=DiagnosisConfig(trigger="always", model="claude-haiku"),
)

reports = experiment.run_evaluations(task_fn)
report = reports[0]

for i in range(len(report.scores)):
    if not report.test_passes[i]:
        print(f"Recommendation: {report.recommendations[i]}")
        print(f"Diagnosis: {report.diagnoses[i]}")
```

## Testing

- 5 unit tests for `diagnose_session()` (skips RCA when no failures, passes model/threshold, serialization round-trip)
- 2 unit tests for `DiagnosisResult.recommendations` (deduplication, empty case)
- 6 unit tests for `Experiment` integration (disabled when no config, calls diagnosis for failing cases, skips passing cases, handles non-Session trajectory, shared across evaluators, exception handling)

```
pytest tests/strands_evals/detectors/test_diagnosis.py tests/strands_evals/detectors/test_types.py tests/strands_evals/test_experiment.py::TestDiagnoseOnFailure -v
```

All 13 tests pass.

## Type of Change

New feature

## Checklist

### DIFF
--- a/src/strands_evals/experiment.py
+++ b/src/strands_evals/experiment.py
@@ -17,6 +17,7 @@ from tenacity import (
 from typing_extensions import Any, Generic
 
 from .case import Case
+from .detectors.diagnosis import diagnose_session
 from .evaluation_data_store import EvaluationDataStore
 from .evaluators.deterministic import Contains, Equals, StartsWith, StateEquals, ToolCalled
 from .evaluators.evaluator import Evaluator
@@ -25,8 +26,10 @@ from .evaluators.output_evaluator import OutputEvaluator
 from .evaluators.trajectory_evaluator import TrajectoryEvaluator
 from .telemetry import get_tracer, serialize
 from .telemetry._cloudwatch_logger import _send_to_cloudwatch
+from .types.detector import DiagnosisConfig
 from .types.evaluation import EvaluationData, InputT, OutputT
 from .types.evaluation_report import EvaluationReport
+from .types.trace import Session
 from .utils import is_throttling_error
 
 logger = logging.getLogger()
@@ -101,13 +104,14 @@ class Experiment(Generic[InputT, OutputT]):
         self,
         cases: list[Case[InputT, OutputT]] | None = None,
         evaluators: list[Evaluator[InputT, OutputT]] | None = None,
+        diagnosis_config: DiagnosisConfig | None = None,
     ):
         self._cases = cases or []
         self._evaluators = evaluators or [Evaluator()]
         self._tracer = get_tracer()
-        # self._logger = get_logger(__name__)
 
         self._config_id = os.environ.get("EVALUATION_RESULTS_LOG_GROUP", "default-strands-evals")
+        self._diagnosis_config = diagnosis_config
 
     @property
     def cases(self) -> list[Case[InputT, OutputT]]:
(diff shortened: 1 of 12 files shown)


src/strands_evals/experiment.py
```diff
@@ -101,13 +105,18 @@ def __init__(
         self,
         cases: list[Case[InputT, OutputT]] | None = None,
         evaluators: list[Evaluator[InputT, OutputT]] | None = None,
+        diagnose_on_failure: bool = False,
+        diagnosis_model: "Model | str | None" = None,
+        diagnosis_confidence_threshold: ConfidenceLevel = "low",
```
afarntrog: I really like this new devx. However, I think we're introducing too many new parameters for this one feature. Can we merge this into one? Something like, `diagnostic_config` or some other fancy name that we can come up with.

afarntrog (changes requested):

src/strands_evals/experiment.py
poshinchen: Moved to DiagnosisConfig!

src/strands_evals/types/evaluation_report.py
afarntrog, line 40: This method is ignoring the new diagnostic report. Let's update to include it.

afarntrog (changes requested):

afarntrog (dismissed):

afarntrog (approved):

PR #211, opened 2026-05-04
chore(detectors): included more fields to the RCAItem

poshinchen:
## Description

Adds `failure_detection_timing` and `completion_status` fields to `RCAItem`, the public API output type for root cause analysis.

These fields were already produced by the LLM (present in `RootCauseItem` / `RCAStructuredOutput`) but were being dropped in `_parse_structured_result` when mapping to the simplified `RCAItem`. Now they flow through to callers.

**Why:**
- `failure_detection_timing` (e.g. `SILENT_UNDETECTED`) signals stealth bugs vs. obvious errors — qualitatively useful for prioritizing fixes.
- `completion_status` per root cause tells you which failures were recovered vs. which sank the task — useful when multiple root causes exist and you need to triage.

**Changes:**
- `src/strands_evals/types/detector.py` — added two fields to `RCAItem`
- `src/strands_evals/detectors/root_cause_analyzer.py` — updated `_parse_structured_result` to pass the fields through
- `tests/strands_evals/detectors/test_types.py` — updated `RCAItem` constructions with new required fields
- `tests/strands_evals/detectors/test_diagnosis.py` — same

## Type of Change

New feature (non-breaking additive fields)

## Testing

- [x] All detector unit tests pass (`pytest tests/strands_evals/detectors/ -v` — 18 passed)

## Checklist

### DIFF
--- a/src/strands_evals/types/detector.py
+++ b/src/strands_evals/types/detector.py
@@ -61,6 +61,10 @@ class RCAItem(BaseModel):
     location: str = Field(description="Span where root cause originated")
     causality: str = Field(description="PRIMARY_FAILURE | SECONDARY_FAILURE | TERTIARY_FAILURE")
     propagation_impact: list[str] = Field(default_factory=list)
+    failure_detection_timing: str = Field(
+        description="IMMEDIATELY_AT_OCCURRENCE | SEVERAL_STEPS_LATER | ONLY_AT_TASK_END | SILENT_UNDETECTED"
+    )
+    completion_status: str = Field(description="COMPLETE_SUCCESS | PARTIAL_SUCCESS | COMPLETE_FAILURE")
     root_cause_explanation: str
     fix_type: str = Field(description="SYSTEM_PROMPT_FIX | TOOL_DESCRIPTION_FIX | OTHERS")
     fix_recommendation: str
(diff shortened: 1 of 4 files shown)


src/strands_evals/types/detector.py
```diff
@@ -61,6 +61,10 @@ class RCAItem(BaseModel):
     location: str = Field(description="Span where root cause originated")
     causality: str = Field(description="PRIMARY_FAILURE | SECONDARY_FAILURE | TERTIARY_FAILURE")
     propagation_impact: list[str] = Field(default_factory=list)
+    failure_detection_timing: str = Field(
+        description="IMMEDIATELY_AT_OCCURRENCE | SEVERAL_STEPS_LATER | ONLY_AT_TASK_END | SILENT_UNDETECTED"
```
afarntrog, line 65: nit: Is this coming from an enum? If so let's update the description to point to that enum so that we don't have to update this list each time.

tests/strands_evals/detectors/test_diagnosis.py
```diff
@@ -73,6 +73,8 @@ def test_with_failures_runs_rca(self, mock_detect, mock_rca):
                 location="span_1",
                 causality="PRIMARY_FAILURE",
                 propagation_impact=["QUALITY_DEGRADATION"],
+                failure_detection_timing="IMMEDIATELY_AT_OCCURRENCE",
```
afarntrog, line 76: nit: would like to see this as an enum ideally

afarntrog (approved):

Issue #114, opened 2026-02-02
[FEATURE] Chaos/Resiliency Evaluation of Agents


vinayakarannil:
### Problem Statement

When evaluating AI agents in production environments, developers currently lack systematic approaches to assess operational resilience under failure conditions. This creates several critical gaps.

- Untested failure modes: Agents are deployed without understanding how they handle tool failures, resource constraints, degraded inputs, or coordination issues
- Production incidents: Agents fail unpredictably when dependent services experience degradation, leading to customer experience issues
- Limited observability: No standardized way to measure agent robustness, retry strategies, fallback behaviors, or graceful degradation
- Evaluation blind spot: Traditional evaluation focuses on functional correctness under normal conditions but ignores operational resilience
.
I would like Strands to support chaos/resiliency testing as a core capability (may be as an additional library)

### Proposed Solution

Add a chaos testing library that enables systematic injection of failures during agent execution:

1. Failure injection framework: Tool-agnostic interception layer that wraps tool calls to inject controlled failures without modifying agent core logic
2. Failure categories: Support for tool failures, resource constraints, input degradation, output corruption, and multi-agent coordination failures
3. Declarative configuration: Define chaos experiments through configuration files specifying failure types, injection rates, and conditions
4. Resiliency metrics: Built-in evaluators to measure retry strategies, fallback effectiveness, error handling, and graceful degradation
5. Framework compatibility: Support agents built with any framework (LangChain, AutoGPT, CrewAI, custom implementations)

### Use Case

• Pre-production validation: Test agent resilience before deployment by simulating API timeouts, rate limits, and service degradation
• Resiliency benchmarking: Measure and compare agent robustness across different implementations or configurations
• Failure mode discovery: Identify weak points in agent error handling and recovery logic
• Production readiness: Validate that agents can handle real-world infrastructure failures gracefully

PR #230, opened 2026-05-15
feat(aggregators): add general aggregator for evaluation reports

venkatkrish543re:
## Description

Adds a general aggregator over EvaluationReport lists. Groups results by (case_key, evaluator_name),  computes descriptive stats. The three overridable hooks(_group_key, _filter_entry, _build_result) for project specific subclasses

LLM summary is optional via a configurable system_prompt. JSON serialization and terminal display included.
## Related Issues

https://github.com/strands-agents/evals/issues/114
## Type of Change

New feature

## Testing

How have you tested the change? Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agent-tools, agents-cli

## Checklist

----


src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,718 @@
...
+
+Inputs are ``EvaluationReport`` objects whose per-case metadata may carry:
+
+    metadata = {
+        "case_key": str,            # logical case identifier (optional;
+                                    # falls back to case_data["name"])
+        "condition_label": str,     # condition this trial belongs to;
+                                    # required for paired comparisons
+        "trial_idx": int,           # pairing key across conditions
+        "corrupted": bool,          # excluded from stats when True
```
ybdarrenwang: This PR is supposed to be only about general case aggregator. This "corrupt" concept and the `trajectory_pointers` don't belong to this PR.

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,718 @@
+"""General-purpose aggregator over per-trial evaluation reports.
+
+Groups reports by ``(case_key, evaluator_name)``, computes descriptive
+statistics, filters corrupted trials, rolls up efficiency metrics, and
+optionally computes paired statistics when exactly two conditions are
+present. An LLM summary is generated per group when a model is configured.
+
+Inputs are ``EvaluationReport`` objects whose per-case metadata may carry:
```
ybdarrenwang: Your proposed stats features are nice for large-scale CI/CD tests. But they belong to P1/P2. Besides, many of them depends on the existence of a certain `metadata` fields. afaik they are not yet in Strands Evals main; in fact we might deprecate the `metadata` field entirely one day.

Can we leave these to future waves:
- Paired statistical tests (P1)
- Bootstrap CI (P1)
- Efficiency rollup (P2)
- raw_values, trajectory_pointers, condition_label / trial_idx pairing

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,718 @@
+"""General-purpose aggregator over per-trial evaluation reports.
```
ybdarrenwang: What this PR actually expects are simply:
1. Aggregator base class — groups by evaluator, computes basic stats (mean/min/max/pass_rate), aggergate reasons
2. AggregationResult type — the simple one (group_key, evaluator_name, stats, summary)
3. AggregationReport wrapper — a nice addition you made

Please strip down all others as the other comments specified.

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,718 @@
...
+_DEFAULT_METRICS = ("pass_rate", "tokens_in", "tokens_out", "wall_clock_s", "cost_usd")
+
+# Test selection: "auto" picks paired_t / wilcoxon based on normality.
+_VALID_TESTS = {"auto", "wilcoxon", "paired_t", "mcnemar"}
+
+# Shapiro-Wilk alpha for "auto" — use paired-t when it fails to reject normality.
+_NORMALITY_ALPHA = 0.05
+
+# Bootstrap CI configuration.
+_BOOTSTRAP_RESAMPLES = 1000
```
sangminwoo: Skills will probably want 10,000 resamples for the win-rate CI. This being a module-level constant means we can't override it without monkey-patching. Would it be reasonable to expose through the constructor (1,000 stays the default)?

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,718 @@
...
+    # ------------------------------------------------------------------
+
+    def _build_aggregation(
+        self, case_key: str, evaluator_name: str, entries: list[dict]
+    ) -> AggregationResult:
+        """Build one ``AggregationResult`` from grouped trial entries."""
+        n_total = len(entries)
+        used_entries = [e for e in entries if not e["corrupted"]]
+        n_corrupted = n_total - len(used_entries)
+        n_used = len(used_entries)
```
sangminwoo: `n_total`, `n_corrupted`, and `n_used` on `AggregationResult` are aggregated across all conditions, so the per-condition breakdown is gone by the time the result is built. For skills, the per-condition counts matter because they change the cross-product size (5×5 vs. 5×4) and users need to see which side dropped trials.

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,718 @@
...
+        base_stats = self._compute_stats(
+            scores=[e["score"] for e in used_entries],
+            passes=[e["passed"] for e in used_entries],
+        )
+
+        # Efficiency rollup over used entries.
+        efficiency = self._aggregate_efficiency(used_entries)
+
+        # Raw per-trial values and trajectory pointers.
+        raw_values = self._collect_raw_values(used_entries)
```
sangminwoo: `_collect_raw_values` returns one flat list per metric across both conditions. For skills' downstream analysis we'd want to recover the per-condition arrays without re-parsing trajectories. Wondering if there's a clean way to keep the condition association at this step.

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,718 @@
...
+        trajectory_pointers = [e["session_id"] for e in used_entries if e["session_id"]]
```
sangminwoo: Same shape question as `raw_values`: this list mixes conditions. When skills hands a pair of session_ids to the judge, we'd need to know which one is with-skill and which is without-skill. Is there a way to keep the condition label alongside the pointer?

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,718 @@
...
+            if e["condition_label"] is not None and e["trial_idx"] is not None
+        ]
+        if not labeled:
+            return []
+
+        # Preserve first-seen order so the user has a stable baseline_label.
+        seen_order: list[str] = []
+        for e in labeled:
+            if e["condition_label"] not in seen_order:
+                seen_order.append(e["condition_label"])
```
sangminwoo: The split-by-condition logic is inlined here. Skills' subclass needs the same split to build cross-product pairs, so it'd help if this could be exposed (or factored out) so we don't end up reimplementing the labeling/ordering rules.

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,369 @@
...
+filtering, efficiency rollups, etc.) is intentionally out of scope here.
+Subclasses can layer those on top by overriding ``_group_key``,
+``_filter_entry``, or ``_build_result``.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections import defaultdict
+from typing import Any, Optional, cast
```
poshinchen: nit

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,369 @@
+"""General-purpose aggregator over evaluation reports.
+
+The default ``Aggregator`` groups per-trial results by
+``(case_key, evaluator_name)`` and computes descriptive statistics
```
ybdarrenwang: I think grouping by `evaluator_name` makes sense because, if you think of the 2 dimensions of case vs. evaluator in an experiment, since the existing experiment report basically groups by case (summarizing all evaluation results w.r.t. 1 case), it makes sense that this new aggregator summarizes from the other dimension which is `evaluator_name`. And that's why in `_group_results` there's a flatten (breaking down reports by case X evaluator) → regrouping (by evaluator) happening.

On the other hand though, despite I do see why you introduced `metadata["case_key"]` which tries to provide an additional grouping id besides case name, how's the user experience going to be? Your current implementation:
- prevents grouping different case name if `case_key` is not assigned (because  case_key = case name by default);
- when use wants to mix match aggregate different subsets of cases, they have to assign new `case_key` every time.

Can users not simply manage the grouping by case name? That is:
```
cases = [
    Case(name="foo", ...),
    Case(name="bar", ...),
    Case(name="baz", ...)
]

reports = Experiment(cases=cases, ...).run_evaluations(...)

aggregated_report_1 = Aggregator(reports, cases=["foo", "bar"])
aggregated_report_2 = Aggregator(reports, cases=["bar", "baz"])
```

src/strands_evals/aggregators/base.py
```diff
@@ -0,0 +1,369 @@
...
+        Default keeps all entries. Subclasses can override to drop entries
+        based on metadata (e.g. ``entry["metadata"].get("corrupted")``).
+        """
+        return True
+
+    # ------------------------------------------------------------------
+    # Per-group aggregation
+    # ------------------------------------------------------------------
+
+    def _build_result(
```
sangminwoo: What I want to verify: when a subclass overrides `_build_result`, does it receive the raw list of trial entries as its second argument, or just the pre-computed summary stats (mean, min, max, pass_rate)?

If it gets the raw entries, we're fine. Skill Eval needs the raw per-pair scores to compute a bootstrap 95% CI on the win-rate, and Chaos Testing likely needs the raw failure events to count specific failure modes. Neither of those can be reconstructed from the summary stats.

If `_build_result` only sees the summary stats, every project subclass would have to re-read the original EvaluationReport list themselves, which defeats the whole purpose of grouping. In that case I'd ask for the second argument to be the raw entries list (with the summary stats available as a third argument or as a method on `self`).

A code snippet showing what we'd want to write in our Skill Eval subclass:

```python
def _build_result(self, group_key, entries):
    result = super()._build_result(group_key, entries)
    scores = [e["score"] for e in entries]   # raw per-pair scores
    result.bootstrap_ci_low, result.bootstrap_ci_high = bootstrap_winrate(scores)
    return result
```

ybdarrenwang: +1 on the right track that the original complete reports shall be preserved.

src/strands_evals/types/evaluation_report.py
```diff
@@ -5,9 +5,9 @@
 
 from ..display.display_console import CollapsibleTableReportDisplay
 from ..types.evaluation import EvaluationOutput
+from ..aggregation import DataFrameMixin
 
-
-class EvaluationReport(BaseModel):
+class EvaluationReport(DataFrameMixin, BaseModel):
```
ybdarrenwang: Why no directly modify `class EvaluationReport` into owning the methods in DataFrameMixin? I thought per our discussion we've decided it's the right way to generalize EvaluationReport into a dataframe-like object.

src/strands_evals/aggregation/__init__.py
```diff
@@ -0,0 +1,39 @@
+"""Pandas-style aggregation for evaluation reports.
```
ybdarrenwang: There's slight naming collision with `class Evaluator`'s `self.aggregator`. Plus the 3 scripts are more about generalizing experiment report into dataframe. Can we dissolve these 3 into other places instead of creating a new module?

src/strands_evals/aggregation/frame.py
```diff
@@ -0,0 +1,211 @@
+"""Lightweight wrappers around ``pandas`` for evaluation results.
```
ybdarrenwang: Can we move `frame.py` to `types/`?

[edit] In fact, we can add these classes directly at the beginning of evaluation_report because they're all Report* specific classes, only used by the evaluation report.

src/strands_evals/aggregation/coerce.py
```diff
@@ -0,0 +1,103 @@
...
+        return pd.DataFrame(columns=_BASE_COLUMNS)
+
+    frame = pd.DataFrame(rows)
+    # Keep the base columns first, metadata columns after, for stable display.
+    meta_cols = [c for c in frame.columns if c.startswith(METADATA_PREFIX)]
+    ordered = [c for c in _BASE_COLUMNS if c in frame.columns] + sorted(meta_cols)
+    return frame[ordered]
+
+
+def flatten(reports: Iterable["EvaluationReport"]) -> pd.DataFrame:
```
ybdarrenwang: We can and should dissolve this coerce.py because:
- The name "coerce" is vague. It doesn't tell you what is being "coerced" or to what. A reader has to open the file to find out.
- The module is small — it's really just two functions: to_dataframe(report) and flatten(reports) (which is just pd.concat over to_dataframe).
- The flatten function collides semantically with EvaluationReport.flatten — the PR's __init__.py even calls this out: "EvaluationReport.flatten already exists upstream with different semantics." Having two flatten concepts in the same package is a maintenance hazard.

Can we not:
- Give to_dataframe to be an instance method on EvaluationReport, just like other common exporting pattern like to_json, to_file etc.
- Update EvaluationReport.flatten so it does both: 
```
@classmethod
def flatten(cls, reports: list["EvaluationReport"]) -> "EvaluationReport":
    """Flatten multiple reports into one, with a DataFrame view."""
    # concat the parallel lists (existing behavior)
    ...
    # also build the combined DataFrame internally
    ...
```

venkatkrish543re: Agree on dissolving this. `to_dataframe` becomes an instance method on `EvaluationReport `alongside `to_json/to_file`.
One thing to confirm on folding `flatten`: the current `flatten` merges into one report with `evaluator_name="Combined"`, so if `to_dataframe()` reads `self.evaluator_name`, a flattened multi-evaluator set tags every row `"Combined"` and `group_by("evaluator") `collapses to one group — that's the collision the `__init__` note flagged. If I build the combined frame inside `flatten` from the source reports (`pd.concat` of each report's `to_dataframe()`), every row keeps its real evaluator and we get back `EvaluationReport.flatten(reports).group_by("evaluator").agg(...)` from the previous discussion notes. Good with that approach?

src/strands_evals/aggregation/frame.py
venkatkrish543re: Will move `ReportFrame/ReportGroupBy` into `evaluation_report.py`  agreed they're report-only. If the file gets heavy I'll split them into a` types/report_frame.py` sibling, but defaulting to same-file

src/strands_evals/aggregation/__init__.py
venkatkrish543re: Good catch on the `Evaluator.aggregator` clash. Dropping the module entirely: `to_dataframe` → instance method, `ReportFrame`/`ReportGroupBy` → `evaluation_report.py`, `flatten` → folded into the existing classmethod. No new package.

src/strands_evals/types/evaluation_report.py
venkatkrish543re: Agreed. The mixin only existed to keep the new code isolated and unit-testable while it was a separate package  now that the methods live on `EvaluationReport` directly it's needless indirection. Removing it and defining the methods on the class

src/strands_evals/aggregation/coerce.py
ybdarrenwang: Yes. Building the combined DataFrame inside flatten by pd.concat-ing each source report's to_dataframe() output lgtm.

src/strands_evals/types/evaluation_report.py
```diff
@@ -1,11 +1,198 @@
...
+EVALUATOR = "evaluator"
+SCORE = "score"
+PASSED = "passed"
+REASON = "reason"
+METADATA_PREFIX = "metadata."
+
+_BASE_COLUMNS = [CASE, EVALUATOR, SCORE, PASSED, REASON]
+
+
+def _case_key(case: dict, index: int) -> str:
```
ybdarrenwang, line 24: Let's tuck these 2 private methods into the class where they're called.

tests/types/test_evaluation_report_dataframe.py
```diff
@@ -0,0 +1,370 @@
+"""Tests for the pandas-style surface on ``EvaluationReport``.
+
+Consolidates what were previously test_coerce / test_frame / test_mixin, now
```
ybdarrenwang, line 3: Remove this legacy slop

src/strands_evals/types/evaluation_report.py
```diff
@@ -1,11 +1,198 @@
...
+        """Return a copy of the underlying frame."""
+        return self._df.copy()
+
+    # -- core operations ------------------------------------------------
+
+    def group_by(self, by: str | list[str], **kwargs: Any) -> ReportGroupBy:
+        """Group rows by one or more columns. Mirrors ``DataFrame.groupby``."""
+        return ReportGroupBy(self._df.groupby(by, **kwargs))
+
+    def filter(
```
ybdarrenwang, line 111: All the helper methods — `filter`, `describe`, `display`, `to_file`/`from_file`, `to_pandas`, the `__getattr__` passthrough, `__len__`, `__repr__` — are nice to have that mirrors Pandas dataframe indeed. But these speculative API surfaces are not really consumed by EvaluationReport or any other class in the codebase. I worry that they only add dead weights to the codebase.

tests/types/test_evaluation_report_dataframe.py
```diff
@@ -0,0 +1,370 @@
+"""Tests for the pandas-style surface on ``EvaluationReport``.
```
ybdarrenwang, line 1: Add these tests to `test_evaluation_report.py` instead of creating this new test file.

src/strands_evals/types/evaluation_report.py
```diff
@@ -1,11 +1,198 @@
...
+            report.group_by("evaluator").agg(
+                mean_score=("score", "mean"),
+                pass_rate=("passed", "mean"),
+                n=("score", "count"),
+            )
+        """
+        result = self._groupby.agg(*args, **kwargs)
+        return ReportFrame(result.reset_index())
+
+    def __getattr__(self, name: str) -> Any:
```
ybdarrenwang, line 61: This method is not needed either. If a user genuinely needs raw pandas, they do `report.frame().df.groupby(...)`

src/strands_evals/types/evaluation_report.py
```diff
@@ -61,6 +255,86 @@ def flatten(cls, reports: list["EvaluationReport"]) -> "EvaluationReport":
...
+
+            report.group_by("evaluator").agg(mean_score=("score", "mean"))
+
+        Across reports, flatten first::
+
+            EvaluationReport.flatten(reports).group_by("evaluator").agg(...)
+        """
+        return self.frame().group_by(by, **kwargs)
+
+    def agg(self, *args: Any, **kwargs: Any) -> ReportFrame:
```
ybdarrenwang, line 324: Ideally, `.agg` should still return a `EvaluationReport`, not a ReportFrame, so the users can directly
```
aggregated = report.group_by("evaluator").agg(
    mean_score=("score", "mean"),
    pass_rate=("passed", "mean"),
)
aggregated.display()  # ← same CollapsibleTableReportDisplay experience
```
and so you don't need to implement `display` for `ReportFrame` either.

The intention is all the pandas df should be hiding beneath `EvaluationReport`, while its interface remains the same except newly supported `.group_by().agg()`.

Here's the shape (Delete ReportFrame entirely):
```
class EvaluationReport(BaseModel):

    def group_by(self, by: str | list[str]) -> "ReportGroupBy":
        """Group cases. Returns an intermediate that resolves back to EvaluationReport."""
        df = self._to_dataframe()  # private — hidden from users
        return ReportGroupBy(df.groupby(by))


class ReportGroupBy:
    """Internal. Users only see .agg() on this."""

    def __init__(self, groupby):
        self._groupby = groupby

    def agg(self, **kwargs) -> "EvaluationReport":
        """Aggregate and return a new EvaluationReport."""
        result_df = self._groupby.agg(**kwargs).reset_index()
        # Convert aggregated rows back into EvaluationReport's structure
        return _dataframe_to_report(result_df)

    def _dataframe_to_report(df: pd.DataFrame) -> EvaluationReport:
        """Convert an aggregated DataFrame back into an EvaluationReport."""
        cases = []
        scores = []
        test_passes = []
        reasons = []
    
        for _, row in df.iterrows():
            case = {"name": str(row.get("case", row.get("evaluator", "")))}
            # Stuff remaining columns into metadata
            case["metadata"] = {
                k: v for k, v in row.items()
                if k not in ("case", "evaluator", "score", "passed", "reason")
            }
            cases.append(case)
            scores.append(float(row.get("score", row.get("mean_score", 0.0))))
            test_passes.append(bool(row.get("passed", row.get("pass_rate", 0) >= 0.5)))
            reasons.append(str(row.get("reason", "")))
    
        overall = sum(scores) / len(scores) if scores else 0.0
        return EvaluationReport(
            evaluator_name="Aggregated",
            overall_score=overall,
            scores=scores,
            cases=cases,
            test_passes=test_passes,
            reasons=reasons,
        )

```

venkatkrish543re:
Closing per working group discussion. With #241 returning a single flattened report, grouping and aggregation are straightforward with pandas directly on the report, So a dedicated aggregator in the library is not needed

PR #232, opened 2026-05-22
chore: allow additional fields to EvaluationData and flexible experiment report type

poshinchen:
## Description

Two related extensibility changes to `Case` / `EvaluationData` and `Experiment` so subclasses can carry typed fields end-to-end and customize the report element type without fighting overrides.

**1. `EvaluationData` accepts extra fields from `Case` subclasses.**
- Added `model_config = ConfigDict(extra="allow")` to `EvaluationData`.
- `Experiment._run_task_async` now passes any subclass-only `Case` fields through via `getattr` (not `model_dump`, so nested `BaseModel`s reach the evaluator with their types intact).
- A `Case` subclass that adds a typed field (e.g. `config: SomeModel`) no longer needs to flatten that data into `metadata` — the typed object reaches the evaluator directly.

**2. `Experiment` is generic over its report element type.**
- New `ReportT = TypeVar("ReportT", bound=EvaluationReport, default=EvaluationReport)`.
- `Experiment[InputT, OutputT, ReportT]` with a `report_cls: type[ReportT] = EvaluationReport` class attribute.
- `run_evaluations` and `run_evaluations_async` return `list[ReportT]`.
- Subclasses bind `ReportT = MyReport` and set `report_cls = MyReport`; no `# type: ignore[override]` needed.

## Related Issues

n/a

## Documentation PR

n/a

## Type of Change

New feature (extensibility hooks; backwards-compatible)

## Testing

- `hatch test` — 1069 unit tests pass (1065 existing + 4 new)
- `mypy src/strands_evals/experiment.py tests/strands_evals/test_experiment.py` — clean
- New tests cover both extension points: typed subclass-field passthrough into `EvaluationData`, an evaluator reading that typed field end-to-end, the default `report_cls` behavior, and a subclass swapping `report_cls`.
- Default callers (`Experiment[str, str]`, `Experiment(cases=..., evaluators=...)`) unchanged at runtime and in type-checker output thanks to the PEP 696 TypeVar default.

## Backwards compatibility

Runtime: identical for default usage. Static typing: existing `Experiment[str, str]` annotations keep working because `ReportT` defaults via PEP 696 (already supported by the repo's pinned `typing_extensions>=4.13.2`). Modern mypy/pyright handle this; only callers on outdated type-checkers would see a "too few type arguments" warning.

## Checklist

----

### DIFF
--- a/src/strands_evals/types/evaluation.py
+++ b/src/strands_evals/types/evaluation.py
@@ -1,4 +1,4 @@
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict
 from typing_extensions import Any, Generic, TypedDict, TypeVar
 
 from .trace import Session
@@ -93,7 +93,14 @@ class EvaluationData(BaseModel, Generic[InputT, OutputT]):
         metadata: Additional information about the test case.
         actual_interactions: The actual interaction sequence given the input.
         expected_interactions: The expected interaction sequence given the input.
+
+    Extra fields from `Case` subclasses are preserved as model attributes (e.g. a typed
+    `config` field on a `Case` subclass passes through to the evaluator with its type intact).
+    Note: `extra="allow"` means misspelled field names will be silently accepted rather than
+    rejected — the tradeoff for supporting subclass passthrough.
     """
+
+    model_config = ConfigDict(extra="allow")
 
     input: InputT
     actual_output: OutputT | None = None
--- a/src/strands_evals/experiment.py
+++ b/src/strands_evals/experiment.py
@@ -14,7 +14,7 @@ from tenacity import (
     stop_after_attempt,
     wait_exponential,
 )
-from typing_extensions import Any, Generic
+from typing_extensions import Any, Generic, TypeVar
 
 from .case import Case
 from .detectors.diagnosis import diagnose_session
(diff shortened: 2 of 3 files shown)


src/strands_evals/types/evaluation.py
```diff
@@ -93,8 +93,15 @@ class EvaluationData(BaseModel, Generic[InputT, OutputT]):
...
         actual_interactions: The actual interaction sequence given the input.
         expected_interactions: The expected interaction sequence given the input.
+
+    Extra fields from `Case` subclasses are preserved as model attributes (e.g. a typed
+    `config` field on a `Case` subclass passes through to the evaluator with its type intact).
+    Note: `extra="allow"` means misspelled field names will be silently accepted rather than
+    rejected — the tradeoff for supporting subclass passthrough.
     """
 
+    model_config = ConfigDict(extra="allow")
```
jjbuck, line 103: Let's double-check that there's not a subtle gotcha lurking here.

The pydantic extras are preserved as live objects in memory, but they have no field annotation, so model_validate can't reconstruct their type. Any `EvaluationDataStore` round-trips via `model_dump/model_validate`, so on a cache hit a typed `config: SomeModel` comes back as a plain dict:
```
  FRESH path : config -> SomeModel   (config.label works)
  CACHED path: config -> dict        (config.label raises AttributeError)
```

So an evaluator written with `evaluation_case.config.label` passes on first run and raises an exception on the second once `evaluation_data_store=` is set. The new tests only cover the cache-miss path.



  Please either:
  1. add a cache-path test (save → load → assert the evaluator still reads the field), and document that extras survive as dict through a store; or
  2.  have subclasses declare the field on a typed `EvaluationData` subclass instead of relying on `extra="allow"`.

  Minimal repro:
```
  ed = EvaluationData(input="hi", config=SomeModel(label="x", weight=0.7))
  EvaluationData.model_validate(ed.model_dump()).config  # -> dict, not SomeModel
```

poshinchen: yes I think this is not a blocker as of now, but we definitely need to think about how to achieve the recovery when loading from the `EvaluationData`.

PR #239, opened 2026-06-02
fix: joined tool_results to prevent context lost

poshinchen:
## Description

Fixes a bug where multi-block `toolResult.content` was silently dropped to the first block, causing `FaithfulnessEvaluator` (and other evaluators) to flag values from `content[1+]` as hallucinated.

**Root cause:** Three sites in `StrandsInMemorySessionMapper` and two in `CloudWatchSessionMapper` read `content[0]` (or `response[0]`) only:
- `StrandsInMemorySessionMapper._process_tool_results` (legacy convention)
- `StrandsInMemorySessionMapper._convert_inference_messages` (latest convention — had a `## To-do` comment)
- `StrandsInMemorySessionMapper._convert_tool_execution_span` (latest convention)
- `CloudWatchSessionMapper._extract_tool_results` / `_process_tool_results` (via `_extract_tool_result_text`)

**Fix:** Two shared helpers in `mappers/utils.py`, used by both mappers:
- `serialize_tool_result_block(block)` — serializes a single Bedrock-style content block. Handles `text`, `json` (via `json.dumps`), and `image`/`document`/`video` (placeholder marker so the judge knows non-text data exists).
- `join_tool_result_content(content)` — joins all blocks with `\n`, filtering empties.

This addresses both the reported `text`-block bug **and** the latent same-shape bug for `json` blocks (common via Strands' `@tool` decorator returning dicts).

## Related Issues

Fixes https://github.com/strands-agents/evals/issues/235

Ref: https://github.com/aws/agentcore-cli/issues/1393

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

- 4 new regression tests in `test_strands_in_memory_mapper.py` covering: legacy multi-text, legacy text+json, latest-convention multi-text inference, latest-convention multi-text tool execution span.
- 1 new regression test in `test_cloudwatch_session_mapper.py` for multi-block tool result reaching the `ToolExecutionSpan`.
- All 172 existing mapper tests pass unchanged.

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #240, opened 2026-06-02
fix(mappers): join all toolResult.content blocks to fix faithfulness false negatives

### LINKED ISSUE #235: [BUG] FaithfulnessEvaluator drops multi-part toolResult.content → false negatives
### Checks

- [x] I have searched issues and there are no duplicates of my issue

### Strands Version

1.18.0

### Strands Evals Version

0.1.0

### Python Version

3.10

### Operating System

macOS / Linux (reproducible on both)

### Installation Method

pip

### Steps to Reproduce

1. Create an agent trace where a tool returns multi-part `toolResult.content` (e.g., `content[0]` = summary, `content[1]` = raw output containing specific values like "Output: 2")
2. The agent's final answer references values from `content[1]`
3. Run `FaithfulnessEvaluator` against this trace
4. Observe the evaluator claims those values are not present in tool outputs

The root cause is in `StrandsInMemorySessionMapper._process_tool_results()` (line 200):

```python
result_text = content[0].get("text", "") if isinstance(content, list) else str(content)
```

Same `[0]`-only pattern at `_convert_inference_messages()` (~line 334) and `_convert_tool_execution_span()` (~lines 386-388).

### Expected Behavior

All content blocks in `toolResult.content` should be included when building the context for the faithfulness judge. Values present in any content block should not be flagged as hallucinated.

### Actual Behavior

Only `content[0]` is read. Values in `content[1+]` are invisible to the evaluator, causing false negative faithfulness scores.

### Additional Context

There is a TODO at line 331 acknowledging this is incomplete:
```python
## To-do: Compare the differences for multiple toolResults
```

Downstream impact: false negatives triggered an automated self-improvement pipeline to generate unnecessary prompt changes to 4 production agents.

Ref: https://github.com/aws/agentcore-cli/issues/1393

### Possible Solution

```python
# Concatenate all content blocks instead of taking only [0]:
result_text = "\n".join(block.get("text", "") for block in content) if isinstance(content, list) else str(content)
```

Apply the same fix to the other two locations (~lines 334 and 386-388).

### Related Issues

https://github.com/aws/agentcore-cli/issues/1393

poshinchen:
Closes #235

## Problem

`FaithfulnessEvaluator` produced false negatives because both mappers read only `content[0]` from Bedrock-style `toolResult.content` lists, silently discarding blocks at index 1+.

## Solution

- Add `join_tool_result_content()` helper in `src/strands_evals/mappers/utils.py` that joins every block type: text → passthrough (with `None`/non-str coercion), json → `json.dumps(sort_keys=True)` (with serialization error guard), image/document/video → placeholder marker, unknown → silently skipped with a debug log. Empty-string text blocks are filtered from the join.
- Extract `_response_to_text()` module-level helper in `strands_in_memory_session_mapper.py` to eliminate the duplicated list-vs-string ternary.
- Replace all `content[0]`-only reads in `strands_in_memory_session_mapper.py` (`_process_tool_results`, `_convert_inference_messages`, `_convert_tool_execution_span`) and `cloudwatch_session_mapper.py` (`_extract_tool_result_text`).

## Testing

All 198 mapper tests pass. Regression tests added for: legacy multi-text, legacy text+json, latest-convention inference span multi-text, latest-convention tool-execution span multi-text, and CloudWatch multi-block tool result. Direct unit tests in `TestJoinToolResultContent` cover all block types, edge cases (None, empty list, falsy non-None, empty-string text block, non-serializable json, bare dict, unknown keys, multi-key precedence).

### DIFF
--- a/src/strands_evals/mappers/cloudwatch_session_mapper.py
+++ b/src/strands_evals/mappers/cloudwatch_session_mapper.py
@@ -17,7 +17,7 @@ from datetime import datetime, timezone
 from typing import Any
 
 from ..mappers.session_mapper import SessionMapper
-from ..mappers.utils import get_body
+from ..mappers.utils import get_body, join_tool_result_content
 from ..types.trace import (
     AgentInvocationSpan,
     AssistantMessage,
@@ -318,11 +318,7 @@ class CloudWatchSessionMapper(SessionMapper):
 
     def _extract_tool_result_text(self, content: Any) -> str:
         """Extract text from tool result content."""
-        if not content:
-            return ""
-        if isinstance(content, list) and content:
-            return content[0].get("text", "")
-        return str(content)
+        return join_tool_result_content(content)
 
     # --- Body-to-messages conversion ---
 
--- a/src/strands_evals/mappers/strands_in_memory_session_mapper.py
+++ b/src/strands_evals/mappers/strands_in_memory_session_mapper.py
@@ -24,6 +24,7 @@ from ..types.trace import (
     UserMessage,
 )
 from .session_mapper import SessionMapper
+from .utils import join_tool_result_content
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


poshinchen:
/strands which implementation would you prefer, "https://github.com/strands-agents/evals/pull/239" or this.

jjbuck (approved):

PR #241, opened 2026-06-03
chore(report): always return flattened report

poshinchen:
## Description

**Breaking change.** `Experiment.run_evaluations()` and `run_evaluations_async()` (plus the chaos and red-team subclasses) now return a single `EvaluationReport` instead of `list[EvaluationReport]`. Callers no longer need to write `reports[0]` or call `EvaluationReport.flatten(reports)` themselves.

Every case row carries an `evaluator` tag (`report.cases[i]["evaluator"]`), regardless of how many evaluators ran. With multiple evaluators, the report is flattened across `(case, evaluator)` pairs so callers can group/filter without an extra step.

### What changed

- `Experiment.run_evaluations[_async]` returns a single `EvaluationReport`.
- `EvaluationReport` no longer has an `evaluator_name` field — each row's evaluator is on the row itself.
- `EvaluationReport.flatten(reports)` is a pure concatenation now: it preserves whatever `evaluator` tag each row already has and does not stamp a new one.
- `RedTeamReport.from_evaluation_reports(list)` was renamed to `from_evaluation_report(report)` to match the new shape.

### Migration

```diff
- reports = experiment.run_evaluations(task)
- reports[0].run_display()
+ report = experiment.run_evaluations(task)
+ report.run_display()
```

For multi-evaluator runs, walk the single report and read the per-row tag:

```python
report = experiment.run_evaluations(task)
for row, score in zip(report.cases, report.scores, strict=True):
    print(row["evaluator"], row["name"], score)
```

If you previously stored per-evaluator reports separately (e.g., across multiple experiments), tag each row before merging:

```python
from strands_evals.types.evaluation_report import EvaluationReport

tagged = [
    EvaluationReport(**{**r.model_dump(), "cases": [{**c, "evaluator": name} for c in r.cases]})
    for name, r in zip(["Eval1", "Eval2"], [r1, r2], strict=True)
]
combined = EvaluationReport.flatten(tagged)
```

## Type of Change

Breaking change

## Testing

- Updated every test that previously indexed `reports[0]` or asserted `len(reports) == N`. Multi-evaluator tests look up rows by their `evaluator` tag.
- Full unit suite: 1192 passed.
- \`hatch fmt --linter --check\` and \`hatch fmt --formatter --check\` clean.
- Updated \`README.md\`, \`SKILL.md\`, and \`AGENTS.md\` to describe the new return shape.

- [x] I ran \`hatch run prepare\`

## Checklist

----

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -202,7 +202,7 @@ All evaluators subclass `strands_evals.evaluators.Evaluator[InputT, OutputT]`:
 
 - `Case[InputT, OutputT]`: one test scenario (`input`, `expected_output`, optional `trajectory`, `metadata`)
 - `Experiment[InputT, OutputT]`: collection of Cases plus evaluators
-- Entry point: `experiment.run_evaluations(task_function)` returns a list of reports
+- Entry point: `experiment.run_evaluations(task_function)` returns a single `EvaluationReport`. With multiple evaluators, results are flattened across (case, evaluator) pairs and each row is tagged via `cases[i]["evaluator"]`.
 
 ### Session / Trace Types
 
--- a/README.md
+++ b/README.md
@@ -93,8 +93,8 @@ def get_response(case: Case) -> str:
     return str(agent(case.input))
 
 # Run evaluations
-reports = experiment.run_evaluations(get_response)
-reports[0].run_display()
+report = experiment.run_evaluations(get_response)
+report.run_display()
 ```
 
 ## Installation
@@ -194,8 +194,8 @@ evaluators = [HelpfulnessEvaluator()]
 experiment = Experiment[str, str](cases=test_cases, evaluators=evaluators)
 
 # Run evaluations
-reports = experiment.run_evaluations(user_task_function)
-reports[0].run_display()
+report = experiment.run_evaluations(user_task_function)
+report.run_display()
 ```
 
 ### Multi-turn Conversation Simulation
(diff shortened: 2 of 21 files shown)


jjbuck (approved):
Non-blocking nit: there's the possibility for evaluator name collision in the flattened evaluation report unless we add a first class "name" field to the evaluators.

E.g., if we run 3 cases × 2 evaluators, using deterministic evaluators (Contains + Equals), we get the following
```
  return type:        EvaluationReport      ← PR delivers: one report, not list[…]
  # rows (cases):     6                     ← 3 cases × 2 evaluators, flattened
  Keys on each row:   actual_output, evaluator, expected_output, input, metadata, name, …
```
Concretely, this would produce the following output report.

```

  #  name             evaluator  score  pass  reason
  0  capital-france   Contains    1.00  True  actual_output contains 'capital'
  1  capital-japan    Contains    0.00 False  actual_output does not contain 'capital'
  2  capital-italy    Contains    1.00  True  actual_output contains 'capital'
  3  capital-france   Equals      1.00  True  actual_output matches expected value
  4  capital-japan    Equals      0.00 False  actual_output does not match expected value
  5  capital-italy    Equals      1.00  True  actual_output matches expected value
```

The evaluator tag comes from Evaluator.get_type_name(), which is just cls.__name__ (evaluator.py:265). Two instances of the same class (e.g., two OutputEvaluators with different rubrics, two Contains for different substrings) are indistinguishable.  

As a follow-up, we should give Evaluator an optional name/id (defaulting to the class name) and key evaluator_data + the row tag on that, so same-class instances stay distinct and the doubling disappears.

PR #243, opened 2026-06-04
feat(cli): added strands-evals cli

poshinchen:
## Description

Adds the `strands-evals` CLI: a five-subcommand console script for running, validating, and inspecting experiments without writing a Python entry point.

### Subcommands

- `run` — load an `Experiment` from JSON, resolve `--agent` (factory callable) or `--task`, execute `run_evaluations_async`, write a flattened `EvaluationReport` to stdout / `-o`.
- `validate` — schema-check an experiment file (case + evaluator counts).
- `report` — render a saved `EvaluationReport` as JSON or a Rich table.
- `diagnose` — run failure detection / RCA over a saved `Session`.
- `version` — print the installed `strands_evals` version.

### Notable design choices

- **`module:attr` entry points** (like `pytest --pyargs` / `gunicorn`). Supports dotted module names (`pkg.agent:build`) and path-style specs (`./agent.py:build`, `../sibling/agent:build`).
- **`--agent` accepts factory callables only** (zero-arg or one-arg taking a `Case`). Prebuilt `Agent` instances and `Agent` subclasses are rejected at resolution time, since each case needs a fresh agent to avoid sharing conversation state.
- **Per-case OTel context** carries `session.id`, `gen_ai.conversation.id`, and any `--trace-attributes` as W3C Baggage. A new `_BaggageSpanProcessor` stamps these onto every span, so the in-memory exporter no longer needs to be cleared between cases — `--max-workers > 1` works correctly.
- **Standard exit codes**: `0` ok, `1` evaluator failures (configurable via `--fail-on`), `2` bad input (file not found, schema invalid, entry-point spec invalid), `3` runtime error.
- **Auto format**: `--rich` when stdout is a TTY, `--json` otherwise. Both can be set explicitly.
- `Experiment.from_dict` registers all built-in evaluators (Coherence, Faithfulness, Helpfulness, multimodal variants, etc.), so serialized experiments round-trip without `custom_evaluators=`.

### Invocation

```
strands-evals run experiment.json --agent ./agent.py:build_agent
python -m strands_evals run experiment.json --agent pkg.agent:build_agent
```

### Next step
- Create the CLI for experiment generation
- Download and store traces from remote via CLI 

## Type of Change

New feature

## Testing

Tests cover entry-point resolution (dotted / path / cwd), each subcommand end-to-end, factory-only `--agent` classification with arity checks, baggage attached/detached around the agent invocation, the shared in-memory exporter not being cleared between cases, the baggage span processor stamping attributes on spans, and built-in evaluators round-tripping through `from_dict`.

## Checklist

----


agent-of-mkmeral:
## 🤖 Automated multi-agent review — `strands-evals` CLI

Hey @poshinchen 👋 — we ran a multi-agent review (CLI/UX, code, adversarial/security, and design) and re-verified everything live at HEAD `e28537` (`pip install -e .`, Python 3.13). This is a **genuinely well-architected PR** — clean `module:attr` entry points (pytest/gunicorn style), sensible 0/1/2/3 exit taxonomy, factory-only `--agent` contract that closes a real conversation-state footgun, and unusually good explanatory docstrings. **70 CLI tests pass.**

**TL;DR: 🟡 great direction, a few robustness + contract fixes worth landing before merge.** None of these are architectural — they're sharp edges that can produce *silently wrong eval results* or break the exit-code contract automation depends on. Details collapsed below. 🧵

<details>
<summary>🚦 <b>Must-fix (robustness + contract)</b></summary>

| # | Issue | Evidence (verified live @ `e28537`) |
|---|-------|-------------------------------------|
| 1 | 🟠 **`--max-workers <= 0` hangs forever** | `run … --max-workers 0` → **exit 124** (killed by timeout). `num_workers = min(max_workers, len(cases))` → 0 workers → `queue.join()` never returns. No validation on the arg. Self-DoS in CI on a typo/unset `$VAR`. **Fix:** reject/clamp `< 1` (argparse `type=` or `max(1, …)`). |
| 2 | 🟡 **Exit code `3` instead of `2` for bad file extension** | `run exp.txt …` → exit **3** (`Only .json format is supported`). `from_file` raises a bare `ValueError`, which isn't in `run_command`'s bad-input tuple, so it maps to `EXIT_RUNTIME_ERROR` instead of `EXIT_BAD_INPUT`. Breaks the documented 0/1/2/3 contract. |
| 3 | 🟡 **Exit code `3` instead of `2` for signatureless builtin `--agent`** | `--agent builtins:dict` → exit **3** (`no signature found for builtin type`). `_callable_arity` calls `inspect.signature()` unguarded; the bare `ValueError` escapes the `EntryPointError` discipline. Wrap & re-raise as `EntryPointError`. |
| 4 | 🟡 **Arity classifier mis-detects required args** | `def factory(*, model)` is accepted as a **zero-arg** factory and then never invoked correctly (`run` exits 0 silently) instead of a clean exit-2 resolution error. `def f(*args)` → arity 0; `len` accepted as 1-arg. `_callable_arity` ignores required keyword-only params and `*args`. This is the worst failure mode for an eval tool — *silently wrong/skipped* results. |
| 5 | 🟡 **`--fail-on` exit-1 path is untested** | The exit-code decision is `run`'s core value-add, but every failure test uses `--fail-on none` / `--exit-zero` (or an invalid-threshold *parse* error). No test asserts **exit 1** on a genuinely failing eval. Cheap unit test on `_decide_exit_code` (`any` + `threshold` above/below `overall_score`) closes it. |

</details>

<details>
<summary>🟢 <b>Strong recommendations (non-blocking)</b></summary>

- **Version the output JSON now.** `EvaluationReport.model_dump_json()` to stdout instantly becomes a de-facto scriptable public contract. Add a top-level `schema_version` field while external consumers = 0 — cheapest-now / expensive-later. Also worth documenting the load-bearing `cases[].evaluator` tag convention that `_print_summary` relies on.
- **Split the OTel baggage/concurrency change into its own telemetry PR.** `_BaggageSpanProcessor` + the `setup_in_memory_exporter` behavior change are **library-level** changes affecting all consumers, riding in a `feat(cli)` PR. They're correct & necessary (they're what make `--max-workers>1` sound), but they should be independently reviewable/bisectable — and the repo *already* ships an `@eval_task` + `TracedHandler` path that this re-implements with a different concurrency strategy. Unifying them removes a real divergence.
- **Restore a way to query the version.** The `version` subcommand was removed and no `--version`/`-V` flag was added — there's currently **no way to query the installed version**. Add `parser.add_argument("--version", action="version", …)` (matches git/pip/uv/gunicorn).
- **Cross-thread telemetry caveat.** Baggage rides on `contextvars`, which don't cross thread boundaries → spans emitted from child threads get `session.id=None`. `opentelemetry-instrumentation-threading` is a declared dep but `ThreadingInstrumentor().instrument()` is never called. Combined with the mapper's include-all fallback, this could theoretically cross-contaminate trajectories under `--max-workers>1`. Either instrument threading or document the single-thread-emit assumption + tighten the mapper fallback.

</details>

<details>
<summary>✅ <b>What held up well</b></summary>

- **No security holes.** Arbitrary code execution via `--agent`/`--task`/`--custom-evaluator` is **by-design** (pytest/gunicorn/inspect-ai convention) and gated on local CLI argv — the caller already controls the process. No trust boundary is crossed. (One doc line worth adding: specs are *trusted code, not data* — pipelines that template `--agent` from untrusted metadata would turn this into real code-exec.)
- **`module:attr` resolution is the strongest part of the CLI** — dotted-vs-path disambiguation, Windows drive-letter `rsplit(":",1)`, cwd injection without leaking `sys.path`, and actionable error messages (file-not-found / no-attr / not-callable). The factory-only `--agent` rejection explaining *why* (shared conversation state) is 👌.
- **OTel context detach on error** (no baggage leak post-exception), **same-thread baggage isolation** across concurrent workers, and **valid JSON on partial failure** (exit 1, no truncation) all verified.
- Standard bad-input paths (missing file / invalid JSON / missing attr) correctly → exit 2.

</details>

---

Nice work overall — the bones are solid and idiomatic. The five must-fixes are small, surgical changes; happy to open issues or send a patch for any of them if useful. 🙏

<sub>🤖 Automated review by an autonomous Strands agent. Findings reproduced with real commands at `e28537` under Python 3.13.</sub>

jjbuck:
Nit: does the `version` command actually exist?

src/strands_evals/cli/_entrypoint.py
```diff
@@ -0,0 +1,299 @@
...
+        arity = _callable_arity(obj)
+        if arity == 0:
+            return ResolvedEntryPoint(kind="callable_no_arg", obj=obj, spec=spec)
+        if arity == 1:
+            return ResolvedEntryPoint(kind="callable_one_arg", obj=obj, spec=spec)
+        raise EntryPointError(
+            f"--agent target '{spec}' is callable but requires {arity} positional "
+            f"arguments; expected 0 (factory) or 1 (factory taking a Case)"
+        )
+    raise EntryPointError(f"--agent target '{spec}' is not callable (got {type(obj).__name__}). {_FACTORY_HINT}")
```
jjbuck, line 252: `classify_agent` only rejects strands.Agent instances/subclasses (lines 232, 237), but the `callable(obj)` branch here accepts any prebuilt callable object as a factory. A prebuilt non-Strands agent instance (e.g. a custom class with `__call__(self, prompt))` classifies as `callable_one_arg`, and `_build_agent` then invokes it as a factory, `obj(case)` passes the Case where the prompt should go, and the return value is treated as the agent and called again with `case.input`. That's either a confusing TypeError or, if the call returns something callable, the silent cross-case state sharing this classifier is meant to prevent.

I think there are two options to mitigate:
1. broaden the heuristic, i.e., reject a resolved object that is an instance (not a function/lambda/functools.partial) unless it's a recognized factory; or
2. at minimum, document in the `--agent` help and this docstring that the freshness guarantee only holds for strands.Agent, and let other shapes fail with a clear message rather than factory-misuse semantics.

jjbuck (approved):

poshinchen:
> Nit: does the `version` command actually exist?

I removed it since it is the same as `pip show strands-evals`.

src/strands_evals/cli/_entrypoint.py
poshinchen: Will take it as follow-up!

PR #249, opened 2026-06-10
chore(evaluator): added evaluator name and evaluator_type for report

poshinchen:
## Description

Two fixes that make `EvaluationReport` rows uniquely identifiable when the same evaluator class is instantiated multiple times in one experiment.

**1. Evaluator instance names**:  
- `Evaluator.__init__` now accepts an optional `name=` kwarg. 
- `get_name()` returns the instance name when set, otherwise falls back to the class name. 
- The per-row `evaluator` tag in `EvaluationReport.cases` (and the `gen_ai.evaluation.name` OTel attribute) keys on `get_name()` so two `Contains(value="x")` / `Contains(value="y")` instances no longer collide.
- `Experiment._validate_evaluator_names` rejects duplicates up front with a clear message. 
- Store and retrieved through `Evaluator.to_dict` / `from_dict`.

**2. `evaluator_type` row tag**: 
- Every report row now also carries `evaluator_type` (= `Evaluator.get_type_name()`, the class name) next to the existing `evaluator` instance tag. 
  - This lets downstream consumers group/aggregate across all instances of the same class (e.g., "average score across every `Contains` instance") without re-deriving the type from the instance name. 
  - `EvaluationReport.flatten` already uses `dict(case)`, so the field flows through to JSON via `to_dict` / `to_file` automatically.

Surface area is intentionally narrow: no changes to the Rich display, the CLI summary line, or OTel span attributes — those still key on the existing `gen_ai.evaluation.name`.

### Results
Screenshot 2026-06-10 at 10 56 38 AM

## Related Issues

Follow-up to the non-blocking nit on #241: https://github.com/strands-agents/evals/pull/241#pullrequestreview-4437956135

## Documentation PR

N/A

## Type of Change

Other: small enhancement to evaluator identification + report row schema. Backwards-compatible — existing consumers that read `cases[i]["evaluator"]` see the same values; `evaluator_type` is purely additive.

## Testing

- New `test_run_evaluations_two_same_class_evaluators_with_distinct_names` covers the doubling scenario from the PR-241 review and now also asserts `evaluator_type == "Contains"` across both instances.
- New `test_run_evaluations_rejects_duplicate_evaluator_names` covers the validation path.
- New `test_evaluator_name_round_trips_through_to_dict_from_dict` and `test_evaluator_name_default_omitted_from_to_dict` cover serialization.
- All 111 tests in `tests/strands_evals/test_experiment.py` pass locally.

## Checklist

----

### DIFF
--- a/src/strands_evals/cli/commands/validate.py
+++ b/src/strands_evals/cli/commands/validate.py
@@ -16,7 +16,7 @@ def _run(args: argparse.Namespace) -> int:
     experiment: Experiment = Experiment.from_file(args.experiment_file, custom_evaluators=custom_evaluators)
 
     case_count = len(experiment.cases)
-    evaluator_names = [evaluator.get_type_name() for evaluator in experiment.evaluators]
+    evaluator_names = [evaluator.get_name() for evaluator in experiment.evaluators]
 
     fmt = resolve_format(args)
     if fmt == "json":
--- a/src/strands_evals/evaluators/chaos/failure_communication_evaluator.py
+++ b/src/strands_evals/evaluators/chaos/failure_communication_evaluator.py
@@ -46,8 +46,9 @@ class FailureCommunicationEvaluator(Evaluator[InputT, OutputT]):
         version: str = "v0",
         model: Model | str | None = None,
         system_prompt: str | None = None,
+        name: str | None = None,
     ):
-        super().__init__()
+        super().__init__(name=name)
         self.version = version
         default_prompt = get_template(version).SYSTEM_PROMPT
         self.system_prompt = system_prompt if system_prompt is not None else default_prompt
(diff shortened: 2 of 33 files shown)


src/strands_evals/cli/commands/validate.py
```diff
@@ -16,7 +16,7 @@ def _run(args: argparse.Namespace) -> int:
     experiment: Experiment = Experiment.from_file(args.experiment_file, custom_evaluators=custom_evaluators)
 
     case_count = len(experiment.cases)
-    evaluator_names = [evaluator.get_type_name() for evaluator in experiment.evaluators]
+    evaluator_names = [evaluator.get_name() for evaluator in experiment.evaluators]
```
zastrowm, line 19: I presume this isn't a breaking change?

zastrowm (approved):

src/strands_evals/cli/commands/validate.py
poshinchen: it isn't a breaking change, evaluator type_name is default so it's compatible.

mehtarac (approved):

jjbuck (approved):

PR #251, opened 2026-06-10
chore(redteam): redteam multi-agent session

poshinchen:
## Description

Adds a `StrandsMultiAgentSession` so red-team strategies can target a `MultiAgentBase` (Graph, Swarm, nested orchestrator) in addition to a single `Agent`, and routes such targets through the task builder.

The session walks the orchestrator tree once at init into two path indexes — leaf agents and orchestrators — and uses them to drive a composite snapshot/restore:

- `snapshot()` → one `Agent.take_snapshot(preset="session")` per leaf + one `serialize_state()` per orchestrator, wrapped in `_MultiAgentSnapshot` and stored opaquely inside `TargetCheckpoint.agent_snapshot`.
- `restore()` → orchestrators first via `deserialize_state`, leaves last via `load_snapshot`, then trace truncation. The order matters: `Graph.deserialize_state` / `Swarm.deserialize_state` reset every node's executor state to `GraphBuilder`-build-time when the payload has no `next_nodes_to_execute` (the normal between-turn state for a `PENDING` / `COMPLETED` orchestrator). Restoring orchestrators first lets the per-leaf snapshots be the final writers; otherwise every Crescendo backtrack against a Graph or Swarm target would silently restart the conversation from build-time and break the "escalate from accumulated context" property.
- `reset()` → replays the baseline composite (or, with no baseline, best-effort clears each leaf's `messages`).
- `invoke()` → diffs each leaf's `messages` tail across `root(message)` and runs `_tool_uses_in` over each diff, so tool uses anywhere in the tree are captured.
- `_multi_agent_result_text` flattens `MultiAgentResult.results` via `NodeResult.get_agent_results` to a single string for the strategy; falls back to `str(result)` rather than raising into the per-case try/except.

`task.py` and `_build_session` now accept `Agent | MultiAgentBase | TargetSession`. The build-time baseline for a `MultiAgentBase` target is captured via a throwaway `StrandsMultiAgentSession(agent).snapshot().agent_snapshot` so this layer never has to import `_MultiAgentSnapshot` directly.

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

New feature (multi-agent target support for the experimental red-team slice).

## Testing

- New `tests/strands_evals/experimental/redteam/test_multi_agent_session.py` (16 tests) covering tree indexing (top-level, nested, unsupported-executor skip), `invoke` + tail-diff trace capture, snapshot/restore round trips, wrong-payload `TypeError`, and `reset()` with and without a baseline. Uses real `Agent` instances so the SDK `take_snapshot`/`load_snapshot` paths run for real.
- A dedicated `TestRestoreOrderVsGraphReset` class with a `_GraphLikeOrchestrator` whose `deserialize_state` mimics the real `Graph` / `Swarm` reset path (calls executor-state reset on every node when `next_nodes_to_execute` is empty). Both regression tests fail if the loops in `_restore` are reverted to leaf-first ordering and pass with orchestrators-first.
- Existing `tests/strands_evals/experimental/redteam/test_task.py` extended to verify a `MultiAgentBase` target is routed to `StrandsMultiAgentSession` and the baseline is captured once at build time.
- Full red-team suite green: `hatch test tests/strands_evals/experimental/redteam/` → 133 passed.
- `hatch fmt --linter` clean on the touched files.

## Checklist

----

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -240,7 +240,7 @@ Shared helpers live in `detectors/utils.py`. Types live in `types/detector.py`. 
 
 ### Docstrings
 
-Use Google-style docstrings for public functions, classes, and modules.
+Use Google-style docstrings for public functions, classes, and modules. Use single backticks (`` `foo` ``) for inline code references, not RST-style double backticks (`` ``foo`` ``).
 
 ### Logging Style
 
--- a/src/strands_evals/experimental/redteam/strategies/__init__.py
+++ b/src/strands_evals/experimental/redteam/strategies/__init__.py
@@ -2,7 +2,13 @@ from .base import AttackRunResult, AttackStrategy
 from .crescendo import CrescendoStrategy
 from .prompt_strategy import PromptStrategy
 from .prompt_strategy.gradual_escalation import get_template as _gradual_escalation_template
-from .target_session import StrandsAgentSession, TargetCheckpoint, TargetSession, ToolUseEntry
+from .target_session import (
+    StrandsAgentSession,
+    StrandsMultiAgentSession,
+    TargetCheckpoint,
+    TargetSession,
+    ToolUseEntry,
+)
 
 # Ready-made strategy instances users can pass to RedTeamExperiment(attack_strategies=[...]).
 # Strategy instances are shared across cases, so each must keep `__init__` for static
@@ -19,6 +25,7 @@ __all__ = [
     "CrescendoStrategy",
     "PromptStrategy",
     "StrandsAgentSession",
+    "StrandsMultiAgentSession",
     "TargetCheckpoint",
     "TargetSession",
     "ToolUseEntry",
--- a/src/strands_evals/experimental/redteam/strategies/target_session.py
+++ b/src/strands_evals/experimental/redteam/strategies/target_session.py
@@ -22,7 +22,7 @@ from dataclasses import dataclass
 from typing import Any, Protocol, TypedDict
 
 from strands import Agent, Snapshot
-from strands.multiagent.base import MultiAgentBase
+from strands.multiagent.base import MultiAgentBase, Status
 from strands.types.content import Message
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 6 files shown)


jjbuck (changes requested):
I think there's a silent issue where Swarm cases crash but are silently scored "defended".

The relevant locations are `target_session.py:391-397` (_restore), triggered via `task.py:75` and `crescendo:270`

The assumption in this PR's `_restore` function is that both Graph and Swarm `deserialize_state` reset leaves "when the payload has no next_nodes_to_execute., but I think this is true only for Graph.


  - `Graph.deserialize_state` uses a truthiness check (`if not payload.get("next_nodes_to_execute")`). Concretely, an empty list takes the reset path. 
  - `Swarm.deserialize_state` (`swarm.py:1020`) uses a membership check (`"next_nodes_to_execute" in payload`), and `Swarm.serialize_state` (`swarm.py:991`) always emits that key (empty list for a
  settled swarm). So a round-tripped Swarm always takes the resume branch → sets `_resume_from_session=True`, and `_from_dict` leaves `current_node=None`. The next invoke dereferences `current_node.node_id at swarm.py:416.`

So  after `reset: _resume_from_session = True | current_node = None`, we run into an exception where `AttributeError - 'NoneType' object has no attribute 'node_id'`  at `swarm.py:416`

Because `task.py:75` calls `session.reset()` before every case, the first invoke crashes, the base Experiment's per-case try/except records score=0 / "defended," and every Swarm case is silently mislabeled safe. (Secondarily, a settled-checkpoint restore also leaves a stale state.task, so Crescendo's next attacker message is dropped.)

I think the fix is to not treat `deserialize_state` as an idempotent reset across all `MultiAgentBase`. After `orch.deserialize_state(...)`, force `_resume_from_session=False` when the restored status is
  PENDING/COMPLETED/FAILED so the next invoke re-initializes cleanly (leaf rollback still comes from load_snapshot); or explicitly scope the feature to Graph until Swarm is handled.

I think this tricky issue may have snuck because most of the tests are tailored to Graph agents, not swarms.

jjbuck (approved):

PR #255, opened 2026-06-11
feat(issue-labeler): add LLM issue labeler for area and type

yonib05:
## Summary

Adds the reusable [`strands-agents/devtools/issue-labeler`](https://github.com/strands-agents/devtools/tree/main/issue-labeler) action to this repo, mirroring the harness-sdk setup but tuned for Evals.

## What it does

Two parallel jobs run on opened issues and PRs and apply labels from a hardcoded allowlist (the LLM has no tools/shell/API access; worst case is a mislabel):

- **Area** (`max_labels: 2`) — ten `area-*` labels mapped to `src/strands_evals/` subsystems: evaluators, multimodal, simulation, detectors, chaos, redteam, generators, tracing, cli, core.
- **Type** (`max_labels: 1`) — bug / enhancement / question / chore.

## Design notes

- **Overlap handling.** chaos, redteam, and multimodal each have their own evaluators/generators/sessions, so `area.yml` includes precedence rules: the specialized area wins over the general one (e.g. the recovery-strategy evaluator is `area-chaos`, not `area-evaluators`).
- **No catch-all.** `area-core` is explicitly scoped to shared primitives, and the classifier is told to assign no area when none clearly applies.
- **Conventional-commit prefix is authoritative** in `type.yml`, so a `feat:` PR is an enhancement even if it touches CI.
- **Python-only.** No language job; the `python` label is applied elsewhere (issue templates / dependabot).

The 11 labels (10 `area-*` + `chore`) have already been created on the repo. Existing issues/PRs will be backfilled separately.

## Testing

- [ ] After merge, open a test issue and confirm area + type labels are applied correctly.

### DIFF
--- a/.github/labelers/area.yml
+++ b/.github/labelers/area.yml
@@ -16,7 +16,7 @@ instructions: |
   - Failure detection and root cause analysis of a session belong to area-detectors.
   - General experiment generation and topic planning belong to area-generators.
   - Ingesting trace or session data that already exists elsewhere (CloudWatch, Langfuse, OpenSearch providers; session mappers; trace/tool/graph/swarm extractors; OTEL telemetry) belongs to area-tracing. This is about reading external data IN, as opposed to area-simulation which generates new conversations.
-  - CLI commands (run, report, validate, diagnose) and console output belong to area-cli. Use this only when the issue is about the command/console layer itself, not the underlying subsystem it invokes.
+  - CLI commands (run, report, validate, diagnose) and terminal/console output belong to area-cli. area-cli is strictly the command-line and console layer. Do NOT use it for web-based UIs, GUIs, dashboards, hosting, or any non-terminal interface — those have no dedicated area, so assign no area label unless they clearly concern area-core primitives.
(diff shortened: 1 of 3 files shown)


yonib05:
Thanks for the review. On the three optional suggestions, I'm going to decline all three for now and keep the config as-is:

- **Matrix over the two jobs**: with only two configs the duplication is minimal, and separate named jobs (`Label: Area` / `Label: Type`) read more clearly in the Actions UI and let us tune each independently. A matrix would save a few lines at the cost of that clarity.
- **Relabel on title edits**: deliberately `opened`-only. Re-running on every edit means repeated model calls and the risk of fighting a maintainer's manual relabeling. The type label is a starting point for triage, not a binding classification.
- **Pin `@main`**: intentionally consistent with the repo's other `strands-agents/devtools` action usages, which all track `@main`. Pinning just this one would be inconsistent; if we want SHA-pinning it should be a repo-wide decision.

All three are reasonable and we can revisit if the labeler grows.

poshinchen (approved):

PR #256, opened 2026-06-11
fix(templates): correct doc link and clean up issue/PR templates

yonib05:
## Summary

Fixes the broken documentation link and tidies the issue/PR templates.

## Changes

- **`config.yml`** — the documentation contact link pointed at `github.com/strands-agents/docs`. Updated to the Evals quickstart (`https://strandsagents.com/docs/user-guide/evals-sdk/quickstart/`) and added a Discussions support link (Discussions is enabled on this repo).
- **`bug_report.yml` / `feature_request.yml`** — removed the hardcoded type label (`bug` / `enhancement`). Blank issues are disabled, so users sometimes file through the wrong form, which baked in the wrong type. The issue-labeler classifies type from the issue content, so it handles this more reliably. Kept `triage` and added `python` (Evals is Python-only).
- **`bug_report.yml`** — fixed "lastest" typo and dropped the `binary` install-method option (Evals ships via pip only).
- **PR template** — removed the stale `consuming repositories: agents-docs, agents-tools, agents-cli` line (copied from another repo; not applicable here) and fixed the "agent-docs repo" reference.

## Notes

- The `triage` label did not exist on the repo and has been created (the templates referenced it but it was never present).

## Testing

- [ ] Open a new issue via each template and confirm the contact links resolve and labels apply as expected.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,7 +1,7 @@
 name: Bug Report
 description: Report a bug in the Strands Agents Evals
 title: "[BUG] "
-labels: ["bug", "triage"]
+labels: ["triage", "python"]
 assignees: []
 body:
   - type: markdown
@@ -13,7 +13,7 @@ body:
     attributes:
       label: "Checks"
       options:
-        - label: "I have updated to the lastest minor and patch version of Strands and evals"
+        - label: "I have updated to the latest minor and patch version of Strands and evals"
           required: true
         - label: "I have checked the documentation and this is not expected behavior"
           required: true
@@ -59,7 +59,6 @@ body:
       options:
         - pip
         - git clone
-        - binary
         - other
   - type: textarea
     id: steps-to-reproduce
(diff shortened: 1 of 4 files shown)


yonib05:
Thanks — the labeler concern is valid but already resolved: the `issue-labeler` workflow is introduced in companion PR #255 (`.github/workflows/issue-labeler.yml`), which classifies and applies the `bug`/`enhancement`/etc. type labels from issue content. This PR drops the hardcoded template type labels precisely because that automation handles it more reliably (people sometimes file through the wrong form since blank issues are disabled).

**Merge order matters**: #255 should merge before (or together with) this one, so there's no window where new issues get no type label. The CONTRIBUTING.md `label:bug` / `label:enhancement` filter links you flagged will populate correctly once #255 is in.

On the `python` label semantics — fair point that its description is PR-oriented. Since evals is Python-only, applying it to issues is intentional (a cheap, always-correct language signal), but I'm open to tightening the label description in a follow-up if it reads oddly.

mehtarac (approved):

PR #263, opened 2026-06-12
chore: added RedTeamExperiment round-trips

poshinchen:
## Description

`RedTeamExperiment` now round-trips through `to_file` / `from_file` (and `to_dict` / `from_dict`). 
Currently Live targets (agent) are not JSON-serializable, so we'll need *load-then-plug-in*:

```python
exp.to_file("rt.json")
loaded = RedTeamExperiment.from_file("rt.json")
loaded.agent = my_agent
report = loaded.run_evaluations()
```

What's persisted: cases (validated as `RedTeamCase` so the typed `config` survives reload), evaluators, attack strategies, and the experiment-level `model` id. What's deliberately omitted: the live `agent` (`Agent` / `MultiAgentBase` / `TargetSession`) and per-run state (`_run_meta`).

### Changes

- **`AttackStrategy.to_dict` / `from_dict`** (base): emits `strategy_type` + `label`; `from_dict` resolves against a built-in registry (`CrescendoStrategy`, `PromptStrategy`) plus a `custom_strategies=[...]` escape hatch.
- **`CrescendoStrategy.to_dict`**: persists `max_turns`, `max_backtracks`, `success_threshold`, and `model` (when set, as a string id).
- **`PromptStrategy.to_dict`**: persists `strategy_name`, `system_prompt_template`, `max_turns`.
- **`RedTeamExperiment`**:
  - `agent` property + setter (and `attack_strategies` read-only view).
  - `to_dict` override adds `attack_strategies` and `model`; never persists `agent`.
  - `from_dict` / `from_file` overrides accept `custom_evaluators` and `custom_strategies`, auto-register `AttackSuccessEvaluator`, and validate cases as `RedTeamCase`.
- **`_serialize_model` helper** in `strategies/base.py`: coerces `Model | str | None` to a JSON-safe id; preserves `None` (so a strategy's "defer to experiment-level model" semantic survives a round-trip).

### Why setter, not constructor injection at load time

A loaded experiment is a complete config object containing case set, evaluators, strategies, model, and is valid even without an agent. Only `_default_task` reads the agent, so deferring attachment until just before `run_evaluations` mirrors how `cases` and `evaluators` are already exposed as settable properties on the base `Experiment`. Calling `run_evaluations()` with no agent and no explicit `task=` still raises the existing message.

## Type of Change

New feature

## Testing

5 new tests in `tests/strands_evals/experimental/redteam/test_experiment.py`:
- `test_agent_setter_round_trip` — setter accepts a target after construction.
- `test_to_dict_persists_strategies_and_model` — exact serialized shape; verifies `agent` is omitted.
- `test_from_dict_round_trip_runs_after_setting_agent` — full file round-trip, raise-without-agent, then run after attaching.
- `test_from_dict_accepts_custom_strategies` — custom strategy subclass round-trips via the registry hook.
- `test_from_dict_unknown_strategy_raises` — clear error when a custom subclass isn't registered.

- [x] I ran `hatch run prepare` (mypy + ruff clean; full redteam suite, 121 tests, passes).

## Checklist

----

### DIFF
--- a/src/strands_evals/experimental/redteam/utils.py
+++ b/src/strands_evals/experimental/redteam/utils.py
@@ -18,7 +18,7 @@ def _serialize_model(model: Model | str | None) -> str | None:
     `model=None` defers to the experiment-level model, and that semantic
     must survive a round-trip.
 
-    When a non-`None` `Model` instance does not expose a dict `config` with
+    When a non-`None` `Model` instance does not expose a dict config with
     a `model_id`, a warning is logged and `None` is returned. Returning `None`
     means the caller's `to_dict` will drop the field, so reload will fall
     back to the experiment-level model -- a visible behavior change, but the
@@ -26,10 +26,12 @@ def _serialize_model(model: Model | str | None) -> str | None:
     """
     if model is None or isinstance(model, str):
         return model
-    if isinstance(model, Model) and hasattr(model, "config") and isinstance(model.config, dict):
-        model_id = model.config.get("model_id")
-        if model_id is not None:
-            return str(model_id)
+    if isinstance(model, Model):
+        config = model.get_config()
+        if isinstance(config, dict):
+            model_id = config.get("model_id")
+            if model_id is not None:
+                return str(model_id)
     logger.warning(
         "type=<%s> | non-coercible Model: missing dict config with 'model_id'; dropping field",
         type(model).__name__,
--- a/src/strands_evals/experimental/redteam/utils.py
+++ b/src/strands_evals/experimental/redteam/utils.py
@@ -1,0 +1,37 @@
+"""Shared helpers for the red team module."""
+
+from __future__ import annotations
+
+import logging
+
+from strands.models.model import Model
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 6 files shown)


src/strands_evals/experimental/redteam/utils.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import annotations
+
+import logging
+
+from strands.models.model import Model
+
+logger = logging.getLogger(__name__)
+
+
+def _serialize_model(model: Model | str | None) -> str | None:
```
jjbuck, line 12: Nit: `_serialize_model` reads `model.config` but the SDK's contract is `model.get_config()`. The Model ABC exposes `get_config()`. `config` is an implementation detail that happens to exist on `BedrockModel`. A provider that stores config under a different attribute (or as a property) would hit the "non-coercible" warning and silently drop the model id, falling back to the  experiment-level model on reload. Prefer `model.get_config()` (with the same `isinstance(..., dict)` guard) for forward-compatibility across providers.

jjbuck (dismissed):

jjbuck (approved):

PR #265, opened 2026-06-13
refactor(redteam): per-call judge lifecycle + per-risk-category judge rubric + minor strategy improvements

yeomjiwonyeom:
## What

Three related changes to the red-team module, landed together:

1. **Agent lifecycle: judges built fresh per call, attacker fresh per case; strategies stateless across cases.** The four multi-turn strategies (bad_likert_judge, goat, pair, crescendo) used to cache their judge/attacker `Agent`s on `self` and rely on a `reset()` override between cases. They now hold only static config on `self`, build the attacker once per `run_attack`, and build judges fresh per scoring call. This removes the strategy-state half of the `max_workers=1` constraint.

   **Crescendo P5a (bundled with this commit):** `is_refusal(make_judge=None)` now returns `False` — refusal markers alone never decide "refused"; a positive verdict requires a judge. The live path always passes a judge factory, so production behavior is unchanged; only the exported-helper path (PAIR/TAP reuse) is affected.

2. **Judge prompt v1: per-risk-category success criteria (new default).** `AttackSuccessEvaluator`'s v0 rubric defined success generically ("harmful content / policy violation"), which systematically mis-scored some categories — most visibly reading a verbatim system-prompt leak wrapped in "I can't share this" as only0-*partial*. v1 adds a "what counts as success, by risk category" block for the five risk categories. Same four score anchors, same severity labels, same schema — v0 is preserved unchanged.

3. **Bad Likert Judge: per-category score-3 anchor.** The top-score example anchor was a single generic line; it now describes, per risk category, the *shape* of a complete top-score example (no payloads — see Safety).

## "Always build a new agent" — how it landed

This addresses the request to stop reusing cached agents. It splits by agent role:

- **Judge / refusal judge → fresh per call.** A judge that is cleared (`messages.clear()`) and reused still carries other accreting state (conversation-manager counters, agent state) that a fresh agent does not. Building fresh per call makes cross-call carryover structurally impossible rather than relying on a partial scrub, and matches the existing `AttackSuccessEvaluator.evaluate()` idiom (which already builds a fresh `Agent` per call).
- **Attacker → fresh per case, not per call.** The attacker's multi-turn history *is* the strategy (Crescendo escalation, PAIR refinement, GOAT's O/T/S/R chain) — rebuilding it every turn would erase the accumulated context the attack depends on. So it is built once per `run_attack` and never cached across cases.

Net: no agent state survives across cases, and the `reset()`-between-cases dependency is gone — without breaking within-case attacker continuity.

## Why

- **(1)** See above; also drops the `reset()`-between-cases dependency.
- **(2)** Reading the risk category is the principled judge behavior — the judge should know what the targeted violation *is* before scoring it. The category text describes the threat type generally (keyed off the `risk_category` taxonomy, not any specific target), so it applies to any case of a category.
- **(3)** Sharpens what the role-played "judge" should elicit, per category.

## Behavior / compatibility

- **v0 is byte-for-byte unchanged** (anchors + system prompt verbatim; pinned by test). Select it with `version="v0"`.
- The shared judge-prompt builder now emits a `# Risk Category` line from case metadata (already present on `RedTeamCase`); v0's rubric ignores it, so **v0 scoring is unchanged**.
- `reset()` is now a base no-op that strategies don't override; the task runner still calls it harmlessly (kept as a hook for any future strategy that needs per-case state). No base files changed.

## Safety

The Bad Likert Judge anchors are *shape descriptors* of a top-score example (e.g. "reproduces the configuration verbatim, not a paraphrase") — they contain no harmful payloads or instructions. The target still generates any content; nothing harmful is committed to the repo.

## Validation

- Full red-team test suite green; ruff + mypy clean.
- New tests: cross-case isolation without `reset()` (per strategy), per-category score-3 anchor rendering + fallback, risk-category prompt injection, v1 anchors verbatim-match v0.
- A quantitative v0-vs-v1 judge comparison is a held-out evaluation (new targets/goals or human-labeled ground truth) owned by the Science track; this PR supplies the rubric and the lifecycle refactor.

## Note

SequentialBreak gets the same lifecycle transform once it merges (not on this base).


src/strands_evals/experimental/redteam/evaluators/prompt_templates/attack_success/__init__.py
```diff
@@ -1,10 +1,14 @@
-from . import attack_success_v0
+from . import attack_success_v0, attack_success_v1
 
 VERSIONS = {
     "v0": attack_success_v0,
+    "v1": attack_success_v1,
```
poshinchen: Why do we keep v0? we haven't actually announced it, can we make v1 -> v0?

yeomjiwonyeom: Done in 597c0ca. v0 was never released, so there was no reproducibility reason to keep a separate v1. I folded the per-risk-category rubric into `v0` as the single template, deleted `attack_success_v1.py`, and set `DEFAULT_VERSION` back to `v0`. The RUBRIC body is byte-identical to the old v1 (all five category definitions preserved), so judge behavior is unchanged -- this just removes the version slot and the attack_success-only `v1`-default asymmetry, matching every other evaluator in the package (single version, default v0).

src/strands_evals/experimental/redteam/strategies/bad_likert_judge/__init__.py
```diff
@@ -180,11 +181,13 @@ def run_attack(
...
-        # resolving a model that could be a typo) it never uses is wasteful. _success_score
-        # short-circuits on no-criteria before touching the judge, so None is safe on that path.
-        judge = self._judge_agent(self._model or model) if goal.success_criteria else None
+        # A judge FACTORY (not a judge): _success_score builds a fresh judge per scoring call so no
+        # agent state can carry between turns. Pass the factory ONLY when the case has
+        # success_criteria -- a no-criteria case never scores, so we never resolve a model (which
+        # could be a typo) for a judge it would not use. None on that path; _success_score
+        # short-circuits before calling it.
+        resolved_model = self._model or model
+        make_judge = (lambda: self._build_judge(resolved_model)) if goal.success_criteria else None
```
poshinchen: This is the same as `partial(self._build_judge, resolved_model)`? You can drop the checks since it'll be defended later on too.

poshinchen:
Should [Crescendo.is_refusal](https://github.com/strands-agents/evals/blob/56118d4bcb955bbe45a82d2c20dea8900f3a5886/src/strands_evals/experimental/redteam/strategies/crescendo/__init__.py#L82) be independent calls too?

src/strands_evals/experimental/redteam/strategies/bad_likert_judge/__init__.py
yeomjiwonyeom: Done in 96a6075. Switched to `partial(self._build_judge, resolved_model)` unconditionally and dropped the `make_judge is None` clause (plus the `None` from `_success_score`'s signature) — BLJ was the only strategy with a conditional `make_judge`, so this matches crescendo/goat/pair now.

The no-criteria safety is preserved: `partial` is lazy and `_success_score` returns at `if not success_criteria` *before* ever calling `make_judge()`, so a no-criteria case still never resolves a model / builds a judge. That invariant is now guaranteed by call-ordering rather than the None-check; I documented it in the `_success_score` docstring, and the existing `test_no_criteria_does_not_build_judge` (real `run_attack`, asserts the judge builder is never called) covers it end-to-end. Removed the redundant unit test that pinned `_success_score(None, ...) == 0.0` since `None` is no longer a valid input.

yeomjiwonyeom:
@poshinchen on [`is_refusal`](https://github.com/strands-agents/evals/blob/56118d4bcb955bbe45a82d2c20dea8900f3a5886/src/strands_evals/experimental/redteam/strategies/crescendo/__init__.py#L82) — it already is on this branch; that's one of the things this PR does. The link points at `56118d4` (main, the merge base), which still has the old shape: `is_refusal(response, judge: Agent)` reusing one judge via `judge.messages.clear()`, plus cached `self._judge` / `self._refusal_judge`. Refusal checks are independent calls, same as the success scorer. Let me know if you meant something beyond the per-call freshness.

poshinchen (approved):
LGTM

PR #272, opened 2026-06-16
feat(redteam): support async cases execution

poshinchen:
## Description

Lifts the `max_workers=1` restriction in `RedTeamExperiment.run_evaluations_async` so users can run case-strategy cross-products in parallel, and widens the public API to make multi-agent (`Graph`/`Swarm`) targets discoverable.

The previous code pinned `max_workers=1` because the entire experiment shared one `TargetSession` wrapping one target agent — concurrent cases would interleave on `invoke()`/`snapshot()`/`restore()` of that single agent. Built-in strategies are stateless across cases (their `reset()` is a no-op), so the strategy instances themselves are not the blocker. The fix is to give each parallel case its own target.

### Parallelism (the main change)

- New ctor arg `agent_factory: Callable[[], Agent | MultiAgentBase | TargetSession]`. Called once per case to produce a fresh, independent target. Required for parallel runs against a custom `TargetSession` (a Protocol implementer cannot be safely deep-copied).
- When `agent` is a `strands.Agent` / `MultiAgentBase` and no factory is set, parallel runs `copy.deepcopy(agent)` per case so concurrent workers never share mutable state.
- `run_evaluations_async` now accepts any `max_workers >= 1` (default still `1`, fully backward-compatible). Sequential runs keep the existing snapshot-based per-case reset against a single shared target.
- New helper `_build_per_case_task_fn` in `task.py` handles parallel + factory paths; sequential snapshot path unchanged.

Concurrent writes to `self._run_meta` are safe because the cross-product expansion guarantees unique `case.name` per work item, so each thread targets a distinct dict key.

### Multi-agent target discoverability (follow-up to #251)

`StrandsMultiAgentSession` was already wired through `task._build_session`, but the public surface of `RedTeamExperiment` didn't surface it:

- `RedTeamExperiment.__init__` and the `agent` / `agent_factory` getters/setters now type the target as `Agent | MultiAgentBase | TargetSession | None`, matching what the runtime actually accepts.
- `StrandsAgentSession` and `StrandsMultiAgentSession` are re-exported from `strands_evals.experimental.redteam` so users find them at the top-level package, not under `.strategies`.
- Class docstring shows a `Graph` / `Swarm` example alongside the single-`Agent` example.

### Usage

```python
# Sequential (unchanged)
exp = RedTeamExperiment(agent=agent, attack_strategies=[CrescendoStrategy()])
report = await exp.run_evaluations_async()

# Parallel via factory
exp = RedTeamExperiment(
    agent_factory=lambda: Agent(model=..., tools=...),
    attack_strategies=[CrescendoStrategy()],
)
report = await exp.run_evaluations_async(max_workers=8)

# Parallel via deep-copy of a single Agent / MultiAgentBase target
graph = Graph(...)
exp = RedTeamExperiment(agent=graph, attack_strategies=[CrescendoStrategy()])
report = await exp.run_evaluations_async(max_workers=4)
```

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

New feature.

## Testing

- `tests/strands_evals/experimental/redteam/test_experiment.py` — replaced `test_max_workers_must_be_one` with `test_max_workers_must_be_positive`, plus new `test_parallel_requires_clonable_agent_or_factory` and `test_parallel_uses_agent_factory_per_case`.
- `tests/strands_evals/experimental/redteam/test_task.py` — new tests covering the factory-per-case path, the deep-copy-per-case path, the unclonable `TargetSession` rejection, and that the factory path skips the build-time `take_snapshot` call.
- Full red-team suite green: `hatch test tests/strands_evals/experimental/redteam/` → 318 passed.
- `hatch fmt --linter` clean.

## Checklist

----

### DIFF
--- a/src/strands_evals/experimental/redteam/__init__.py
+++ b/src/strands_evals/experimental/redteam/__init__.py
@@ -12,6 +12,8 @@ from .strategies import (
     PairStrategy,
     PromptStrategy,
     SequentialBreakStrategy,
+    StrandsAgentSession,
+    StrandsMultiAgentSession,
     TargetCheckpoint,
     TargetSession,
 )
@@ -36,6 +38,8 @@ __all__ = [
     "RedTeamExperiment",
     "RedTeamReport",
     "SequentialBreakStrategy",
+    "StrandsAgentSession",
+    "StrandsMultiAgentSession",
     "TargetCheckpoint",
     "TargetSession",
     "TargetSpec",
--- a/src/strands_evals/experimental/redteam/experiment.py
+++ b/src/strands_evals/experimental/redteam/experiment.py
@@ -11,6 +11,7 @@ from typing import Any, cast
 
 from strands import Agent
 from strands.models.model import Model
+from strands.multiagent.base import MultiAgentBase
 
 from ...case import Case
 from ...evaluation_data_store import EvaluationDataStore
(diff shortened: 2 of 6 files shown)


poshinchen:
Need to cover the async execution with docs.

jjbuck (approved):

PR #308, opened 2026-07-13
fix: route smolagents OpenInference spans to OpenInferenceSessionMapper

ybdarrenwang:
### Problem
SmolagentsInstrumentor sets its OTel scope to "openinference.instrumentation.smolagents", but OpenInferenceSessionMapper only accepted "openinference.instrumentation.langchain". Smolagents spans were silently filtered out, producing empty Sessions with no tool data.

### Fix
Add SCOPES_OPENINFERENCE_FAMILY — a set containing both the langchain and smolagents scope variants. detect_otel_mapper() and OpenInferenceSessionMapper's span filter now check membership in this set. The mapper's existing attribute parsing already handles smolagents' bare-string output.value format correctly.

## Type of Change
Bug fix

## Testing

## Checklist

----


poshinchen (dismissed):

src/strands_evals/mappers/openinference_session_mapper.py
```diff
@@ -296,13 +296,19 @@ def _convert_tool_execution_span(self, span: dict, session_id: str) -> ToolExecu
...
+                    if isinstance(parsed, dict):
+                        tool_output_content = parsed.get("content", str(parsed))
+                        tool_call_id = parsed.get("tool_call_id")
+                        tool_status = parsed.get("status", "success")
+                    else:
+                        tool_output_content = output_value
                 elif isinstance(output_value, dict):
                     tool_output_content = output_value.get("content", str(output_value))
-            except json.JSONDecodeError:
+            except Exception:
```
jjbuck: Could we remove this outer try/except rather than broaden it? JSON decoding failures are now handled by the inner except json.JSONDecodeError, while the remaining branches explicitly distinguish strings and dictionaries. Any exception after that point is therefore more likely to indicate a mapper defect than malformed telemetry.
 
Catching Exception converts such defects into apparently valid tool output and prevents _build_trace() from logging the conversion failure. That risks silently corrupting evaluation inputs. If another input type is expected, could we handle it explicitly and add a focused test? Otherwise, I suggest removing this catch or narrowing it to the specific anticipated exception.

tests/strands_evals/mappers/test_openinference_session_mapper.py
```diff
@@ -1349,3 +1353,144 @@ def test_no_empty_response_inference_spans(self, adot_session):
...
+
+        assert len(session.traces[0].spans) == 1
+        tool = session.traces[0].spans[0]
+        assert isinstance(tool, ToolExecutionSpan)
+        assert tool.tool_call.name == "web_search"
+        assert tool.tool_result.content == "Temperature: 15C, cloudy with rain"
+
+    def test_smolagents_llm_span(self):
+        """LLM span with smolagents scope produces InferenceSpan."""
+        span = make_llm_span(
```
jjbuck, line 1400: Can we sanity check that this is doing the right thing for smolagents? This test changes the scope but still uses `make_llm_span()`, which produces the existing/older `llm.*.message.content` shape. It therefore verifies mapper routing, not compatibility with current smolagents telemetry.

I saw your related experiments (https://code.amazon.com/packages/AgentCoreMace/blobs/mainline/--/experiments/results/2026-06-24__21-52-35_phoenix_iter04/reflection/trials/trial_008.json#L2630-L2646).... if I'm reading that correctly, it looks like a captured `SmolagentsInstrumentor` run with instrumentation scope version 0.1.32 emits content-block attributes such as `llm.input_messages.0.message.contents.0.message_content.text`, while older 0.1.6 traces use the singular `.message.content` form.

The current mapper only reads the singular representation, so a current LLM span is dropped with no user content. Which instrumentor versions do we intend to support? Could we add a versioned fixture captured from the instrumentor and normalize both representations behind the mapper’s existing interface?

src/strands_evals/mappers/openinference_session_mapper.py
```diff
@@ -27,7 +27,7 @@
     Trace,
     UserMessage,
 )
-from .constants import SCOPE_OPENINFERENCE
+from .constants import SCOPES_OPENINFERENCE_FAMILY
```
jjbuck: Since this changes the mapper from LangChain-specific to multi-producer, could we update the module and class documentation accordingly? They still describe a LangChain-only mapper, and the class documentation claims `AGENT` spans are supported even though `_is_agent_invocation_span()` rejects them. These comments define the module’s contract, so disagreement with the implementation makes the supported behavior difficult to reason about.

src/strands_evals/mappers/openinference_session_mapper.py
```diff
@@ -77,8 +77,8 @@ def map_to_session(self, data: Any, session_id: str) -> Session:
         # Normalize input to flat spans
         spans = self._normalize_to_flat_spans(data)
 
-        # Filter to only spans from this scope
-        openinference_spans = [s for s in spans if self._get_scope_name(s) == SCOPE_OPENINFERENCE]
+        # Filter to only spans from this scope (including smolagents variant)
+        openinference_spans = [s for s in spans if self._get_scope_name(s) in SCOPES_OPENINFERENCE_FAMILY]
```
jjbuck, line 96: This broadens the mapper’s input contract from OpenInference LangChain telemetry to smolagents telemetry. But a shared OpenInference scope prefix and semantic attributes do not guarantee identical payload encoding, i.e., the producers represent agent invocations, message content, and tool arguments differently. Could we use the scope to select a producer-specific normalization path, then feed a canonical representation into the existing conversion logic? That keeps format complexity inside the mapper instead of allowing partial or malformed evaluation data to reach downstream consumers.

tests/strands_evals/mappers/test_openinference_session_mapper.py
```diff
@@ -1349,3 +1353,144 @@ def test_no_empty_response_inference_spans(self, adot_session):
...
+            tool_output="Temperature: 15C, cloudy",
+            tool_call_id="tc-smol",
+            scope_name=SMOLAGENTS_SCOPE_NAME,
+        )
+        session = self.mapper.map_to_session([span], "sess-1")
+
+        tool = session.traces[0].spans[0]
+        assert isinstance(tool, ToolExecutionSpan)
+        assert tool.tool_call.name == "web_search"
+        assert tool.tool_call.arguments == {"query": "weather london"}
```
jjbuck, line 1395: Can we sanity check the materialized representation of the smolagents telemetry? This assertion represents tool arguments as `{"query": ...}`, but if I understand correctly, smolagents instrumentation captures calls in a wrapper like this: `{"args": [4], "kwargs": {}, "sanitize_inputs_outputs": false}`. Passing that wrapper directly into `ToolCall.arguments` exposes instrumentation details to parameter-accuracy evaluators rather than the logical tool arguments.

Assuming I'm interpreting the smolagents format correctly (and I might not be!), can we test the actual smolagents representation and normalize it by merging kwargs, mapping positional args using `tool.parameters`, and excluding instrumentation-only fields?

tests/strands_evals/mappers/test_openinference_session_mapper.py
```diff
@@ -1349,3 +1353,144 @@ def test_no_empty_response_inference_spans(self, adot_session):
...
+            user_content="What is 2+2?",
+            assistant_content="4",
+            scope_name=SMOLAGENTS_SCOPE_NAME,
+        )
+        session = self.mapper.map_to_session([span], "sess-1")
+
+        assert len(session.traces[0].spans) == 1
+        assert isinstance(session.traces[0].spans[0], InferenceSpan)
+
+    def test_smolagents_full_trace(self):
```
jjbuck: Can we sanity check whether this is genuinely a "full" smolagents trace? Doesn't that require the top-level agent invocation? E.g., in smolagents, CodeAgent is an agent implementation and CodeAgent.run(task) executes the agent loop. `SmolagentsInstrumentor` records that call as an OpenTelemetry span named `CodeAgent.run` with `openinference.span.kind="AGENT"`. I.e., its input carries the user task and its output carries the final response.

As written,  `_is_agent_invocation_span()` currently rejects every `AGENT` span, so the mapper produces no `AgentInvocationSpan`. Consequently, CloudWatch returns an empty task output, trace-level extraction finds no turns, and session-level extraction raises. Could this test include the root `AGENT` span and verify that it becomes an `AgentInvocationSpan`, while preserving the existing LangChain-specific rejection behavior?

jjbuck (changes requested):

tests/strands_evals/mappers/test_openinference_session_mapper.py
ybdarrenwang: Added an actual smolagents trace as test material. _is_agent_invocation_span() now accepts smolagents AGENT spans (guarded by requiring both input.value and output.value to distinguish from LangChain routing spans). The fixture test verifies the root CodeAgent.run becomes an AgentInvocationSpan.

src/strands_evals/mappers/openinference_session_mapper.py
ybdarrenwang: Updated module and class documentation

src/strands_evals/mappers/openinference_session_mapper.py
ybdarrenwang: Removed the outer try/except entirely. The inner json.JSONDecodeError catch plus the isinstance(parsed, dict) guard should cover all paths without swallowing unexpected errors.

tests/strands_evals/mappers/test_openinference_session_mapper.py
ybdarrenwang: Added _normalize_smolagents_llm_attrs() normalization for the plural path, plus integration tests against a real fixture that uses that format.

src/strands_evals/mappers/openinference_session_mapper.py
ybdarrenwang: Added per-producer normalization: after scope filtering, smolagents spans pass through _normalize_smolagents_span() which canonicalizes LLM content paths and tool argument wrappers before the shared conversion logic runs.

tests/strands_evals/mappers/test_openinference_session_mapper.py
ybdarrenwang: Added _normalize_smolagents_tool_attrs() to extract kwargs as logical arguments, and an integration test against the real fixture verifying instrumentation fields are excluded.

jjbuck: Thanks, the kwargs-only path is fixed, but positional arguments are still not normalized. The checked-in v0.1.31 fixture’s FinalAnswerTool (https://github.com/strands-agents/evals/pull/308/files#diff-03cdbc88cc6cb7f23cf1639a5e234219d7154b8a6d7358e07b1a017d2a473c35R139) records the following

  > "args": ["The population of Tokyo is ..."],
  > "kwargs": {}
  
Its `tool.parameters` names that positional parameter answer. However, `_normalize_smolagents_tool_attrs()`  (https://github.com/strands-agents/evals/pull/308/files#diff-ef2ce58111d4bef87b94b697df75e4215bff8ccfb46996f1879bcb3ab9c83473R210) currently produces:

  > ToolCall.arguments == {"args": ["The population of Tokyo is ..."]}

rather than the logical representation:

  > ToolCall.arguments == {"answer": "The population of Tokyo is ..."}

The new test only checks `web_search`, whose arguments are entirely in kwargs, so it does not catch this fixture case. Could we use the names in `tool.parameters` to map positional values, then merge kwargs? If that makes sense, please add an exact assertion for `final_answer`, ideally `{"answer": final_answer.tool_result.content}`, plus a mixed args/kwargs case so positional values are not discarded when kwargs are present.

tests/strands_evals/mappers/test_openinference_session_mapper.py
jjbuck: Thanks, the fixture now produces an AgentInvocationSpan, but its user prompt is not yet normalized. At `smolagents_live_spans.json:185`, `CodeAgent.run.input.value` is a JSON object containing task, stream, reset, and other invocation settings.

At `openinference_session_mapper.py:511-512`, that complete JSON string is assigned to `user_prompt`, so evaluators receive the instrumentation wrapper instead of just "What is the population of Tokyo?" The assertion at `test_openinference_session_mapper.py:1631` still passes because the wrapper happens to contain the words “population” and “Tokyo.”

Could we parse this smolagents input and extract its task field, then assert the exact expected prompt?

src/strands_evals/mappers/openinference_session_mapper.py
```diff
@@ -189,17 +308,29 @@ def _is_agent_invocation_span(self, span: dict) -> bool:
...
         span_name = span.get("name", "")
+
+        # LangGraph: CHAIN span named "LangGraph"
         if span_kind == "CHAIN" and span_name == "LangGraph":
             return True
 
+        # smolagents: AGENT span (e.g. CodeAgent.run) — must have both
+        # input.value (user task) and output.value (final answer) to distinguish
+        # from LangChain's mislabeled routing spans that only carry input.
+        if span_kind == "AGENT":
```
jjbuck, line 346: Does requiring both `input.value` and `output.value` preserve the existing LangChain rejection behavior? The checked-in LangChain fixture at openinference_live_spans.json:64-77 contains a `route_to_agent` span with kind `AGENT`, both input and output values, and scope `openinference.instrumentation.langchain`. This new branch therefore classifies it as an `AgentInvocationSpan`.

The current rejection test only constructs an `AGENT` span with input and no output, so it doesn't cover the actual fixture. Deduplication hides the "psuedo" span when a root LangGraph span follows it, but that depends on span ordering and completeness.

Could we require `SCOPE_OPENINFERENCE_SMOLAGENTS` in this branch and add a LangChain case containing both input and output?

jjbuck (changes requested):

src/strands_evals/mappers/openinference_session_mapper.py
ybdarrenwang: Great catch. Will add fork in revision by `SCOPE_OPENINFERENCE_SMOLAGENTS` for cases a) smolagents, b) langchain but has kind AGENT, both input and output values.

ybdarrenwang: Searched the web and found
- https://arize.com/docs/ax/concepts/otel-openinference/span-kinds#agent
- https://docs.inference.net/integrations/traces/langgraph

, confirming LangGraph also emits AGENT span with input and output values

tests/strands_evals/mappers/test_openinference_session_mapper.py
ybdarrenwang: Updated code and added targeted test cases in `test_mixed_positional_and_kwargs_merged`.

tests/strands_evals/mappers/test_openinference_session_mapper.py
ybdarrenwang: Done and verified by `test_agent_span_extracts_task_from_json_wrapper` test.

jjbuck (approved):

poshinchen (approved):

PR #309, opened 2026-07-13
fix: add GenericGenAISessionMapper for dict spans with unrecognized scope

ybdarrenwang:
### Problem
detect_otel_mapper() routes dict-format spans (from readable_spans_to_dicts()) to StrandsInMemorySessionMapper when the scope is unrecognized. But StrandsInMemorySessionMapper only accepts raw ReadableSpan objects — causing an AttributeError crash.

This happens when agents are manually instrumented with a custom OTel tracer (e.g. get_tracer("my-app")) and use gen_ai.* attributes for tool/inference data.

### Fix
Introduce GenericGenAISessionMapper — a new mapper that accepts dict-format spans and parses gen_ai.operation.name, gen_ai.tool.*, and span events. detect_otel_mapper() now routes unrecognized scopes to this mapper instead of StrandsInMemorySessionMapper.

StrandsInMemorySessionMapper remains unchanged and only handles ReadableSpan input for Strands-scoped traces.

## Type of Change
Bug fix

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/mappers/generic_gen_ai_session_mapper.py
+++ b/src/strands_evals/mappers/generic_gen_ai_session_mapper.py
@@ -11,7 +11,7 @@ This is the fallback mapper for spans that don't match any known instrumentor sc
 import json
 import logging
 from collections import defaultdict
-from datetime import datetime, timezone
+from datetime import datetime
 from typing import Any
 
 from ..types.trace import (
@@ -31,6 +31,7 @@ from ..types.trace import (
     UserMessage,
 )
 from .session_mapper import SessionMapper
+from .utils import join_tool_result_content
 
 logger = logging.getLogger(__name__)
 
@@ -62,39 +63,37 @@ class GenericGenAISessionMapper(SessionMapper):
 
         spans = self._normalize_to_flat_spans(data)
 
-        # Group by trace_id, optionally filtering by session.id
-        traces_by_id: dict[str, list[dict]] = defaultdict(list)
+        # Group by trace_id, optionally filtering by session.id.
+        # Single pass: bucket spans by trace and track which traces match session_id.
+        all_traces: dict[str, list[dict]] = defaultdict(list)
+        traces_with_session: set[str] = set()
+        any_span_has_session_id = False
 
-        any_span_has_session_id = any(
-            "session.id" in span.get("attributes", {}) or "gen_ai.conversation.id" in span.get("attributes", {})
-            for span in spans
-        )
-
-        if not any_span_has_session_id:
-            # No spans have session IDs — include all spans
-            for span in spans:
-                trace_id = span.get("trace_id", "unknown")
-                traces_by_id[trace_id].append(span)
-        else:
-            # Some spans have session IDs — use trace-level filtering.
-            # First, identify which trace_ids contain at least one span matching session_id.
-            all_traces: dict[str, list[dict]] = defaultdict(list)
-            traces_with_session: set[str] = set()
-
-            for span in spans:
-                trace_id = span.get("trace_id", "unknown")
-                all_traces[trace_id].append(span)
-                attrs = span.get("attributes", {})
-                span_session_id = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
-                if span_session_id and str(span_session_id) == session_id:
+        for span in spans:
+            trace_id = span.get("trace_id", "unknown")
+            all_traces[trace_id].append(span)
+            attrs = span.get("attributes", {})
+            span_session_id = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
+            if span_session_id:
+                any_span_has_session_id = True
+                if str(span_session_id) == session_id:
                     traces_with_session.add(trace_id)
 
-            # Include all spans from traces that contain at least one matching span.
-            # Exclude entire traces where no span matches the session_id (even if some
-            # spans in that trace have no session_id tag).
+        traces_by_id: dict[str, list[dict]] = defaultdict(list)
+        if not any_span_has_session_id:
+            # No spans carry session tags — include everything.
+            traces_by_id = all_traces
+        else:
+            # Include spans from matching traces; own tag wins over inheritance.
             for trace_id, trace_spans in all_traces.items():
-                if trace_id in traces_with_session:
-                    traces_by_id[trace_id] = trace_spans
+                if trace_id not in traces_with_session:
+                    continue
+                for span in trace_spans:
+                    attrs = span.get("attributes", {})
+                    own = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
+                    if own and str(own) != session_id:
+                        continue
+                    traces_by_id[trace_id].append(span)
 
         traces: list[Trace] = []
         for trace_id, trace_spans in traces_by_id.items():
@@ -147,20 +146,11 @@ class GenericGenAISessionMapper(SessionMapper):
     def _parse_timestamp(self, value: Any) -> datetime:
         """Parse timestamp from various formats.
 
-        readable_spans_to_dicts() preserves OTel's nanosecond epoch timestamps.
-        Values > 1e12 are assumed nanoseconds and divided by 1e9.
+        Delegates to SessionMapper.parse_timestamp, which additionally handles
+        ISO-8601 and string-encoded nanosecond epochs (the OTLP/JSON encoding
+        for int64 fields, as produced on the CloudWatch/ADOT path).
         """
-        if value is None:
-            return datetime.now(timezone.utc)
-        if isinstance(value, datetime):
-            return value
-        if isinstance(value, (int, float)):
-            # OTel timestamps from ReadableSpan are nanoseconds (> 1e18).
-            # Seconds-based timestamps are < 1e10. Threshold at 1e12 to distinguish.
-            if value > 1e12:
-                value = value / 1e9
-            return datetime.fromtimestamp(value, tz=timezone.utc)
-        return datetime.now(timezone.utc)
+        return self.parse_timestamp(value)
 
     def _parse_json_attr(self, attributes: Any, key: str, default: str = "[]") -> Any:
         """Parse a JSON-encoded attribute value."""
@@ -668,9 +658,7 @@ class GenericGenAISessionMapper(SessionMapper):
                 tool_result = item["toolResult"]
                 result_text = tool_result.get("content", "")
                 if isinstance(result_text, list):
-                    result_text = "\n".join(
-                        block.get("text", "") for block in result_text if isinstance(block, dict) and "text" in block
-                    )
+                    result_text = join_tool_result_content(result_text)
                 result.append(
                     ToolResultContent(
                         content=str(result_text),
--- a/src/strands_evals/mappers/generic_gen_ai_session_mapper.py
+++ b/src/strands_evals/mappers/generic_gen_ai_session_mapper.py
@@ -70,18 +70,31 @@ class GenericGenAISessionMapper(SessionMapper):
             for span in spans
         )
 
-        for span in spans:
-            attrs = span.get("attributes", {})
-
-            if not any_span_has_session_id:
-                should_include = True
-            else:
-                span_session_id = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
-                should_include = str(span_session_id) == session_id if span_session_id else True
-
-            if should_include:
+        if not any_span_has_session_id:
+            # No spans have session IDs — include all spans
+            for span in spans:
                 trace_id = span.get("trace_id", "unknown")
                 traces_by_id[trace_id].append(span)
+        else:
+            # Some spans have session IDs — use trace-level filtering.
+            # First, identify which trace_ids contain at least one span matching session_id.
+            all_traces: dict[str, list[dict]] = defaultdict(list)
+            traces_with_session: set[str] = set()
+
+            for span in spans:
+                trace_id = span.get("trace_id", "unknown")
+                all_traces[trace_id].append(span)
+                attrs = span.get("attributes", {})
+                span_session_id = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
+                if span_session_id and str(span_session_id) == session_id:
+                    traces_with_session.add(trace_id)
+
+            # Include all spans from traces that contain at least one matching span.
+            # Exclude entire traces where no span matches the session_id (even if some
+            # spans in that trace have no session_id tag).
+            for trace_id, trace_spans in all_traces.items():
+                if trace_id in traces_with_session:
+                    traces_by_id[trace_id] = trace_spans
 
         traces: list[Trace] = []
         for trace_id, trace_spans in traces_by_id.items():
@@ -124,7 +137,9 @@ class GenericGenAISessionMapper(SessionMapper):
             except (KeyError, TypeError, ValueError, json.JSONDecodeError) as e:
                 logger.warning(
                     "trace_id=%s span_id=%s | Failed to convert span: %s",
-                    span.get("trace_id", "?"), span.get("span_id", "?"), e,
+                    span.get("trace_id", "?"),
+                    span.get("span_id", "?"),
+                    e,
                 )
 
         return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
@@ -194,10 +209,19 @@ class GenericGenAISessionMapper(SessionMapper):
                     assistant_content = self._process_assistant_content(message_list)
                     if assistant_content:
                         messages.append(AssistantMessage(content=assistant_content))
+
+                elif event_name == "gen_ai.client.inference.operation.details":
+                    # Current OTel GenAI convention: single event with input/output messages
+                    # https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-events.md
+                    details_messages = self._parse_operation_details_event(event_attrs)
+                    messages.extend(details_messages)
+
             except (KeyError, TypeError, ValueError, json.JSONDecodeError) as e:
                 logger.warning(
                     "span_id=%s event=%s | Failed to process event: %s",
-                    span.get("span_id", "?"), event.get("event_name", "?"), e,
+                    span.get("span_id", "?"),
+                    event.get("event_name", "?"),
+                    e,
                 )
 
         # Fallback: gen_ai.input.messages / gen_ai.output.messages as span attributes
@@ -214,8 +238,25 @@ class GenericGenAISessionMapper(SessionMapper):
 
         tool_name = str(attrs.get("gen_ai.tool.name", ""))
         tool_call_id = str(attrs.get("gen_ai.tool.call.id", ""))
+
+        # Prefer error.type (current OTel GenAI convention for failed operations),
+        # fall back to gen_ai.tool.status / tool.status for legacy compatibility.
+        # https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/spans.yaml#L522-L562
+        error_type = attrs.get("error.type", "")
+        span_status = span.get("status", {}).get("code", "")
         tool_status = attrs.get("gen_ai.tool.status", attrs.get("tool.status", ""))
-        tool_error = None if tool_status == "success" else (str(tool_status) if tool_status else None)
+
+        if error_type:
+            # OTel convention: error.type indicates the error class (e.g., "TimeoutError")
+            tool_error = str(error_type)
+        elif span_status == "ERROR":
+            # Span status set to ERROR without error.type — still a failure
+            tool_error = "ERROR"
+        elif tool_status and tool_status != "success":
+            # Legacy fallback: gen_ai.tool.status or tool.status
+            tool_error = str(tool_status)
+        else:
+            tool_error = None
 
         tool_arguments: dict = {}
         tool_result_content: str = ""
@@ -234,7 +275,8 @@ class GenericGenAISessionMapper(SessionMapper):
             except (KeyError, TypeError, ValueError, json.JSONDecodeError) as e:
                 logger.warning(
                     "span_id=%s | Failed to process tool event: %s",
-                    span.get("span_id", "?"), e,
+                    span.get("span_id", "?"),
+                    e,
                 )
 
         # Fallback: gen_ai.tool.input/output attributes (manual instrumentation)
@@ -296,7 +338,8 @@ class GenericGenAISessionMapper(SessionMapper):
             except (KeyError, TypeError, ValueError, json.JSONDecodeError) as e:
                 logger.warning(
                     "span_id=%s | Failed to process agent event: %s",
-                    span.get("span_id", "?"), e,
+                    span.get("span_id", "?"),
+                    e,
                 )
 
         # Fallback: extract from attributes (PydanticAI / native gen_ai format)
@@ -320,6 +363,29 @@ class GenericGenAISessionMapper(SessionMapper):
                     pass
 
         if not agent_response:
+            # Try gen_ai.output.messages (current OTel GenAI convention for invoke_agent)
+            # https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md
+            output_msgs_raw = attrs.get("gen_ai.output.messages", "")
+            if output_msgs_raw:
+                try:
+                    output_msgs = json.loads(output_msgs_raw) if isinstance(output_msgs_raw, str) else output_msgs_raw
+                    if isinstance(output_msgs, list):
+                        for msg in output_msgs:
+                            if msg.get("role") == "assistant":
+                                parts = msg.get("parts", [])
+                                for part in parts:
+                                    if part.get("type") == "text" and part.get("content"):
+                                        agent_response = part["content"]
+                                        break
+                                # Also handle flat content string
+                                if not agent_response and msg.get("content"):
+                                    agent_response = str(msg["content"])
+                            if agent_response:
+                                break
+                except (json.JSONDecodeError, TypeError):
+                    pass
+
+        if not agent_response:
             agent_response = str(attrs.get("final_result", ""))
 
         return AgentInvocationSpan(
@@ -338,8 +404,9 @@ class GenericGenAISessionMapper(SessionMapper):
         """Parse gen_ai.input.messages / gen_ai.output.messages from span attributes.
 
         PydanticAI and other native gen_ai semconv frameworks store messages as
-        JSON arrays in span attributes rather than span_events. Format:
-            [{"role": "user", "parts": [{"type": "text", "content": "..."}]}, ...]
+        JSON arrays in span attributes rather than span_events. Supports:
+        - PydanticAI format: [{"role": "user", "parts": [{"type": "text", "content": "..."}]}]
+        - OTel GenAI convention: role "tool" with response field for tool results
         """
         messages: list[UserMessage | AssistantMessage] = []
 
@@ -367,6 +434,54 @@ class GenericGenAISessionMapper(SessionMapper):
                                     )
                             if user_content:
                                 messages.append(UserMessage(content=user_content))
+                        elif role == "tool":
+                            # OTel GenAI convention: role "tool" carries tool results.
+                            # Two formats:
+                            #   1. parts: [{"type": "tool_call_response", "id": "...", "response": "..."}]
+                            #   2. flat: {"role": "tool", "id": "...", "response": "..."}
+                            # https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md
+                            if parts:
+                                for part in parts:
+                                    if isinstance(part, dict) and part.get("type") == "tool_call_response":
+                                        tc_id = part.get("id") or part.get("tool_call_id")
+                                        resp = part.get("response", "")
+                                        if isinstance(resp, dict):
+                                            resp = json.dumps(resp)
+                                        messages.append(
+                                            UserMessage(
+                                                content=[
+                                                    ToolResultContent(
+                                                        content=str(resp) if resp else "",
+                                                        error=None,
+                                                        tool_call_id=tc_id,
+                                                    )
+                                                ]
+                                            )
+                                        )
+                            else:
+                                tool_call_id = msg.get("id") or msg.get("tool_call_id")
+                                response = msg.get("response", msg.get("content", ""))
+                                if isinstance(response, dict):
+                                    response = json.dumps(response)
+                                elif isinstance(response, list):
+                                    parts_text = []
+                                    for block in response:
+                                        if isinstance(block, dict) and block.get("text"):
+                                            parts_text.append(block["text"])
+                                        elif isinstance(block, str):
+                                            parts_text.append(block)
+                                    response = "\n".join(parts_text) if parts_text else str(response)
+                                messages.append(
+                                    UserMessage(
+                                        content=[
+                                            ToolResultContent(
+                                                content=str(response) if response else "",
+                                                error=None,
+                                                tool_call_id=tool_call_id,
+                                            )
+                                        ]
+                                    )
+                                )
                         elif role == "assistant":
                             assistant_content: list[TextContent | ToolCallContent] = []
                             for part in parts:
@@ -411,6 +526,118 @@ class GenericGenAISessionMapper(SessionMapper):
                                 messages.append(AssistantMessage(content=assistant_content))
             except (json.JSONDecodeError, TypeError, KeyError):
                 pass
+
+        return messages
+
+    def _parse_operation_details_event(self, event_attrs: dict) -> list[UserMessage | AssistantMessage]:
+        """Parse gen_ai.client.inference.operation.details event (current OTel GenAI convention).
+
+        This event carries gen_ai.input.messages and gen_ai.output.messages as structured
+        lists or JSON-encoded strings within a single span event.
+        https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-events.md
+        """
+        messages: list[UserMessage | AssistantMessage] = []
+
+        for key in ("gen_ai.input.messages", "gen_ai.output.messages"):
+            raw = event_attrs.get(key, "")
+            if not raw:
+                continue
+            try:
+                msg_list = json.loads(raw) if isinstance(raw, str) else raw
+            except (json.JSONDecodeError, TypeError):
+                continue
+
+            if not isinstance(msg_list, list):
+                continue
+
+            for msg in msg_list:
+                role = msg.get("role", "")
+                parts = msg.get("parts", [])
+
+                if role == "user":
+                    user_content: list[TextContent | ToolResultContent] = []
+                    for part in parts:
+                        if isinstance(part, dict):
+                            if part.get("type") == "text" and part.get("content"):
+                                user_content.append(TextContent(text=part["content"]))
+                            elif part.get("type") == "tool_call_response":
+                                user_content.append(
+                                    ToolResultContent(
+                                        content=str(part.get("result", "")),
+                                        error=None,
+                                        tool_call_id=part.get("id"),
+                                    )
+                                )
+                    if user_content:
+                        messages.append(UserMessage(content=user_content))
+
+                elif role == "tool":
+                    # OTel convention: role "tool" carries tool results.
+                    # Two formats:
+                    #   1. parts: [{"type": "tool_call_response", "id": "...", "response": "..."}]
+                    #   2. flat: {"role": "tool", "id": "...", "response": "..."}
+                    if parts:
+                        for part in parts:
+                            if isinstance(part, dict) and part.get("type") == "tool_call_response":
+                                tc_id = part.get("id") or part.get("tool_call_id")
+                                resp = part.get("response", "")
+                                if isinstance(resp, dict):
+                                    resp = json.dumps(resp)
+                                messages.append(
+                                    UserMessage(
+                                        content=[
+                                            ToolResultContent(
+                                                content=str(resp) if resp else "",
+                                                error=None,
+                                                tool_call_id=tc_id,
+                                            )
+                                        ]
+                                    )
+                                )
+                    else:
+                        tool_call_id = msg.get("id") or msg.get("tool_call_id")
+                        response = msg.get("response", msg.get("content", ""))
+                        if isinstance(response, dict):
+                            response = json.dumps(response)
+                        elif isinstance(response, list):
+                            parts_text = []
+                            for block in response:
+                                if isinstance(block, dict) and block.get("text"):
+                                    parts_text.append(block["text"])
+                                elif isinstance(block, str):
+                                    parts_text.append(block)
+                            response = "\n".join(parts_text) if parts_text else str(response)
+                        messages.append(
+                            UserMessage(
+                                content=[
+                                    ToolResultContent(
+                                        content=str(response) if response else "",
+                                        error=None,
+                                        tool_call_id=tool_call_id,
+                                    )
+                                ]
+                            )
+                        )
+
+                elif role == "assistant":
+                    assistant_content: list[TextContent | ToolCallContent] = []
+                    for part in parts:
+                        if isinstance(part, dict):
+                            if part.get("type") == "text" and part.get("content"):
+                                assistant_content.append(TextContent(text=part["content"]))
+                            elif part.get("type") == "tool_call":
+                                assistant_content.append(
+                                    ToolCallContent(
+                                        name=part.get("name", ""),
+                                        arguments=part.get("arguments", {}),
+                                        tool_call_id=part.get("id"),
+                                    )
+                                )
+                    # Also handle flat content string (no parts)
+                    if not assistant_content and msg.get("content"):
+                        assistant_content.append(TextContent(text=str(msg["content"])))
+                    if assistant_content:
+                        messages.append(AssistantMessage(content=assistant_content))
 
         return messages
 
--- a/src/strands_evals/mappers/generic_gen_ai_session_mapper.py
+++ b/src/strands_evals/mappers/generic_gen_ai_session_mapper.py
@@ -1,0 +1,340 @@
+"""
+GenericGenAISessionMapper - Maps dict-format spans with gen_ai.* attributes to Session format.
+
+Handles spans from any framework that uses OpenTelemetry GenAI semantic conventions
+(gen_ai.operation.name, gen_ai.tool.name, etc.) but has an unrecognized scope name.
+
+This is the fallback mapper for spans that don't match any known instrumentor scope
+(Strands, OpenInference, LangChain) but still carry structured gen_ai.* attributes.
+"""
+
+import json
+import logging
+from collections import defaultdict
+from datetime import datetime, timezone
+from typing import Any
+
+from ..types.trace import (
+    AgentInvocationSpan,
+    AssistantMessage,
+    InferenceSpan,
+    Session,
+    SpanInfo,
+    TextContent,
+    ToolCall,
+    ToolCallContent,
+    ToolConfig,
+    ToolExecutionSpan,
+    ToolResult,
+    ToolResultContent,
+    Trace,
+    UserMessage,
+)
+from .session_mapper import SessionMapper
+
+logger = logging.getLogger(__name__)
+
+
+class GenericGenAISessionMapper(SessionMapper):
+    """Maps dict-format spans with gen_ai.* attributes to Session format.
+
+    Classifies spans by gen_ai.operation.name:
+    - "chat" → InferenceSpan
+    - "execute_tool" → ToolExecutionSpan
+    - "invoke_agent" → AgentInvocationSpan
+
+    Extracts data from both span attributes (gen_ai.tool.input, gen_ai.tool.output)
+    and span_events (gen_ai.user.message, gen_ai.choice, etc.).
+    """
+
+    def map_to_session(self, data: Any, session_id: str) -> Session:
+        """Map dict-format spans to Session.
+
+        Args:
+            data: List of span dicts (from readable_spans_to_dicts or similar)
+            session_id: Session identifier for filtering
+
+        Returns:
+            Session object ready for evaluation
+        """
+        if not data:
+            return Session(traces=[], session_id=session_id)
+
+        spans = self._normalize_to_flat_spans(data)
+
+        # Group by trace_id, optionally filtering by session.id
+        traces_by_id: dict[str, list[dict]] = defaultdict(list)
+
+        any_span_has_session_id = any(
+            "session.id" in span.get("attributes", {}) or "gen_ai.conversation.id" in span.get("attributes", {})
+            for span in spans
+        )
+
+        for span in spans:
+            attrs = span.get("attributes", {})
+
+            if not any_span_has_session_id:
+                should_include = True
+            else:
+                span_session_id = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
+                should_include = str(span_session_id) == session_id if span_session_id else True
+
+            if should_include:
+                trace_id = span.get("trace_id", "unknown")
+                traces_by_id[trace_id].append(span)
+
+        traces: list[Trace] = []
+        for trace_id, trace_spans in traces_by_id.items():
+            trace = self._convert_trace(trace_id, trace_spans, session_id)
+            if trace.spans:
+                traces.append(trace)
+
+        return Session(traces=traces, session_id=session_id)
+
+    def _convert_trace(self, trace_id: str, spans: list[dict], session_id: str) -> Trace:
+        """Convert a list of dict spans (same trace_id) to a Trace."""
+        converted_spans: list[InferenceSpan | ToolExecutionSpan | AgentInvocationSpan] = []
+
+        for span in spans:
+            try:
+                attrs = span.get("attributes", {})
+                operation_name = attrs.get("gen_ai.operation.name", "")
+
+                span_info = SpanInfo(
+                    trace_id=span.get("trace_id"),
+                    span_id=span.get("span_id"),
+                    session_id=session_id,
+                    parent_span_id=span.get("parent_span_id"),
+                    start_time=self._parse_timestamp(span.get("start_time")),
+                    end_time=self._parse_timestamp(span.get("end_time")),
+                )
+
+                if operation_name == "chat":
+                    inference_span = self._convert_inference_span(span, span_info)
+                    if inference_span and inference_span.messages:
+                        converted_spans.append(inference_span)
+                elif operation_name == "execute_tool":
+                    tool_span = self._convert_tool_execution_span(span, span_info)
+                    if tool_span:
+                        converted_spans.append(tool_span)
+                elif operation_name == "invoke_agent":
+                    agent_span = self._convert_agent_invocation_span(span, span_info)
+                    if agent_span:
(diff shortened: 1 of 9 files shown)


jjbuck (dismissed):

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,454 @@
...
+
+        return InferenceSpan(span_info=span_info, messages=messages, metadata={})
+
+    def _convert_tool_execution_span(self, span: dict, span_info: SpanInfo) -> ToolExecutionSpan | None:
+        """Convert a span with gen_ai.operation.name='execute_tool' to ToolExecutionSpan."""
+        attrs = span.get("attributes", {})
+
+        tool_name = str(attrs.get("gen_ai.tool.name", ""))
+        tool_call_id = str(attrs.get("gen_ai.tool.call.id", ""))
+        tool_status = attrs.get("gen_ai.tool.status", attrs.get("tool.status", ""))
```
jjbuck, line 238: These status attributes do not cover the current generic execute_tool error contract. OpenTelemetry  records a failed operation using `error.type` (https://github.com/open-telemetry/semantic-conventions-genai/blob/2e994c6d59a93bb4fc1752c5378eedb9b8e14d6b/model/gen-ai/spans.yaml#L522-L562  together with an error span status.

I reproduced a tool span with error.type="TimeoutError" and status ERROR, and the mapper emits
1ToolResult(error=None)`, making the failed call appear successful to downstream failure and recovery evaluators.
 
Could we prefer `error.type`, retain these existing keys as compatibility fallbacks, and add a failed-tool regression test?

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,454 @@
...
+
+    # =========================================================================
+    # Span Conversion
+    # =========================================================================
+
+    def _convert_inference_span(self, span: dict, span_info: SpanInfo) -> InferenceSpan | None:
+        """Convert a span with gen_ai.operation.name='chat' to InferenceSpan."""
+        messages: list[UserMessage | AssistantMessage] = []
+
+        # Try span_events first (Strands telemetry / manual instrumentation format)
```
jjbuck, line 171: This loop recognizes only the legacy per-message events. The current OpenTelemetry GenAI convention defines a single `gen_ai.client.inference.operation.details` event (https://github.com/open-telemetry/semantic-conventions-genai/blob/2e994c6d59a93bb4fc1752c5378eedb9b8e14d6b/docs/gen-ai/gen-ai-events.md#event-gen_aiclientinferenceoperationdetails), carrying `gen_ai.input.messages` and `gen_ai.output.messages`. Those values may also be structured lists rather than JSON strings.

 A minimal example is:
 ```
span["attributes"] = {"gen_ai.operation.name": "chat"}
span["span_events"] = [{
    "event_name": "gen_ai.client.inference.operation.details",
    "attributes": {
        "gen_ai.input.messages": [
            {"role": "user", "parts": [{"type": "text", "content": "Hello"}]}
        ],
        "gen_ai.output.messages": [
            {"role": "assistant", "parts": [{"type": "text", "content": "Hi"}]}
        ],
    },
}]
```
I reproduced a custom-scope chat span using that event. The detection logic here selects this mapper, but mapping returns an empty session because messages remains empty and the inference span is discarded. Could we support this event, normalize both structured and JSON-encoded values, and add a generic-scope regression test?

File: src/strands_evals/mappers/generic_gen_ai_session_mapper.py [hunk 3]
```diff
@@ -0,0 +1,454 @@
...
+        )
+
+        for span in spans:
+            attrs = span.get("attributes", {})
+
+            if not any_span_has_session_id:
+                should_include = True
+            else:
+                span_session_id = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
+                should_include = str(span_session_id) == session_id if span_session_id else True
```
jjbuck: When at least one span has a session ID, this expression still includes every span without one. For example, mapping [session-A trace, untagged unrelated trace, session-B trace] for session A returns both the session-A and untagged traces. That silently mixes unrelated exporter data into the requested session.

Could we filter at trace level instead, i.e., include an untagged span only when another span in the same trace  identifies the requested session, while excluding entirely untagged traces once session-tagged data is present?

A mixed tagged/untagged regression test would pin down that boundary.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,454 @@
...
+                                for part in parts:
+                                    if part.get("type") == "text" and part.get("content"):
+                                        user_prompt = part["content"]
+                                        break
+                            if user_prompt:
+                                break
+                except (json.JSONDecodeError, TypeError):
+                    pass
+
+        if not agent_response:
```
jjbuck, line 414: This fallback reads the agent response only from PydanticAI’s final_result, but current OpenTelemetry `invoke_agent` spans expose their canonical response through `gen_ai.output.messages` (https://github.com/open-telemetry/semantic-conventions-genai/blob/2e994c6d59a93bb4fc1752c5378eedb9b8e14d6b/docs/gen-ai/gen-ai-agent-spans.md#span-invoke_agent-client).
 
I reproduced a standard span containing both input and output messages such that the mapper extracts `user_prompt` but returns an empty `agent_response`. Because the resulting `AgentInvocationSpan` is otherwise valid, downstream evaluators see an apparently successful invocation with no answer. Could we parse the assistant response from `gen_ai.output.messages` and add that case to the fixture coverage?

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,454 @@
...
+        # Parse input messages
+        input_msgs_raw = attrs.get("gen_ai.input.messages", "")
+        if input_msgs_raw:
+            try:
+                input_msgs = json.loads(input_msgs_raw) if isinstance(input_msgs_raw, str) else input_msgs_raw
+                if isinstance(input_msgs, list):
+                    for msg in input_msgs:
+                        role = msg.get("role", "")
+                        parts = msg.get("parts", [])
+                        if role == "user":
```
jjbuck: This role == "user" branch supports the captured PydanticAI representation, where tool results are user messages. Current OpenTelemetry represents them as role: "tool" with the value under response  (see the gen_ai.input.messages example in https://github.com/open-telemetry/semantic-conventions-genai/blob/2e994c6d59a93bb4fc1752c5378eedb9b8e14d6b/docs/registry/attributes/gen-ai.md#gen-ai-input-messages).

 Because there is no tool role branch, those messages are silently skipped. Could we accept both representations and normalize them into ToolResultContent?

jjbuck (changes requested):

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: fixed in latest commit

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: Added `_parse_operation_details_event` to handle `elif event_name == "gen_ai.client.inference.operation.details"` branch. Also added test with synthetic fixture.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: Implemented the section
```
# Some spans have session IDs — use trace-level filtering.
......
```
plus `TestSessionFilteringTraceLevel`

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: Added before this fallback
```
        if not agent_response:
            # Try gen_ai.output.messages (current OTel GenAI convention for invoke_agent)
            # https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md
            output_msgs_raw = attrs.get("gen_ai.output.messages", "")
            ......
```
and `TestAgentResponseFromOutputMessages`

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: Added `elif role == "tool":` below to capture it as ToolResultContent. Tested with `TestToolRoleMessages`

poshinchen:
@strandly-the-agent Review this pull requests and tell me if there's any potential issues. You should check the otel semantic conventions to make sure that the mapper works as expected (including having the spans locally, and pulling the traces from CloudWatch (ADOT log format I believe)). Also, does it mean that we'll then deprecate the StrandsInMemoryMapper since the behavior is almost the same, or an additional StrandsMapper?

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,681 @@
...
+            # Legacy fallback: gen_ai.tool.status or tool.status
+            tool_error = str(tool_status)
+        else:
+            tool_error = None
+
+        tool_arguments: dict = {}
+        tool_result_content: str = ""
+
+        # Try span_events first (Strands telemetry format)
+        for event in span.get("span_events", []):
```
strandly-the-agent, line 256: ## 🔴 `execute_tool` and `invoke_agent` never read the unified `operation.details` event — `chat` does

This event loop handles only the **legacy** events (`gen_ai.tool.message`, `gen_ai.choice`). `_convert_inference_span` got the `gen_ai.client.inference.operation.details` branch at `:213-217`, but it was never replicated here or in `_convert_agent_invocation_span` (`:327-343`). And the attribute fallback below (`:282-295`) reads only `gen_ai.tool.input` / `gen_ai.tool.call.arguments` / `gen_ai.tool.output` / `gen_ai.tool.call.result` — never `gen_ai.input.messages` / `gen_ai.output.messages`. So tool spans have **two** ways to come back empty, not one.

**Who hits this, and how.** A real Strands `Agent` under `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental`, captured through the documented `InMemorySpanExporter` → `readable_spans_to_dicts()` → `detect_otel_mapper()` path. Same code both runs, only the env var differs:

```
legacy (default):            ToolExecutionSpan result='Seattle: 62F...'  AgentInvocationSpan response='REAL STRANDS ANSWER...'
gen_ai_latest_experimental:  ToolExecutionSpan result=''                 AgentInvocationSpan prompt='' response=''
```

In that mode Strands emits **only** the unified event — no legacy events at all (`strands/telemetry/tracer.py`: `end_tool_call_span:579`, `end_agent_span:774`, gated on `:122`). It also isn't Strands-specific or mixed-batch-specific: a plain custom-scope producer emitting the spec's canonical event, through ordinary auto-detection, gives `chat` → correct `InferenceSpan` but `execute_tool` → `args={} result=''` and `invoke_agent` → `prompt='' response=''`.

**Why this is worth blocking on.** The consequence isn't an error — it's a wrong number. The empty `AgentInvocationSpan` extracts cleanly and reaches the judge (`evaluators/evaluator.py:191,242`), whose `structured_output_model` has no abstain category (`Perfectly` / `Partially` / `Incorrect`), so it must return a confident score for an empty response. At `cc0a3916` all of these inputs raised `AttributeError` — a loud crash. This trades that for silent wrong data, which in an eval framework is the worse failure.

For context, this is the unfixed remainder of `jjbuck`'s `operation.details` thread rather than a new ask — that fix landed for `chat` spans only. `CloudWatchSessionMapper._create_agent_invocation_span` returns `None` when fields are missing and lets `trace_extractor.py:203-204` raise loudly; that's the better precedent, and it already exists in this family.

**Suggested fix** — hoist the `operation.details` branch out of `_convert_inference_span` into a shared helper and call it from all three converters, plus add the message attributes to this method's fallbacks. Roughly:

```python
# in this loop, and in _convert_agent_invocation_span's loop at :327-343:
elif event_name == "gen_ai.client.inference.operation.details":
    detail_msgs = self._parse_operation_details_event(event_attrs)
    # tool span:  pull ToolCallContent / ToolResultContent out of detail_msgs
    # agent span: first user TextContent -> user_prompt,
    #             first assistant TextContent -> agent_response
```

I've deliberately not written this as a `suggestion` block — unlike the three 🟡s below I haven't validated an implementation, and the right factoring (shared helper vs. duplicated `elif`) is your call. The class docstring at `:46-47` also reads as uniform across all three operations and should either be narrowed or updated once this is fixed.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,681 @@
...
+            return datetime.now(timezone.utc)
+        if isinstance(value, datetime):
+            return value
+        if isinstance(value, (int, float)):
+            # OTel timestamps from ReadableSpan are nanoseconds (> 1e18).
+            # Seconds-based timestamps are < 1e10. Threshold at 1e12 to distinguish.
+            if value > 1e12:
+                value = value / 1e9
+            return datetime.fromtimestamp(value, tz=timezone.utc)
+        return datetime.now(timezone.utc)
```
strandly-the-agent: ## 🟡 This re-implements the inherited `parse_timestamp`, less capably — and misses the CloudWatch/OTLP-JSON case

`SessionMapper.parse_timestamp` (`session_mapper.py:77-111`) already exists and additionally handles ISO-8601 and **string-encoded** nanosecond epochs. All three sibling dict mappers call it; this is the only one with a private copy.

That matters because OTLP/JSON encodes int64 fields as decimal **strings**, and `CloudWatchLogsParser._normalize_span` (`cloudwatch_parser.py:173-174`) passes `startTimeUnixNano`/`endTimeUnixNano` through untouched. A string falls to `:163` and becomes `datetime.now()` — differently wrong per span, so anything ordering-sensitive scrambles too rather than failing:

```python
m._parse_timestamp("1700000000000000000")  # -> now()      (wrong)
m.parse_timestamp("1700000000000000000")   # -> 2023-11-14 (correct, inherited)
```

🟡 rather than 🔴 only because I couldn't confirm from AWS docs whether real `aws/spans` records land as strings or numbers — three fetches returned landing pages. One Logs Insights query would settle it, but the fix costs nothing either way. Also note the docstring documents the ns-vs-seconds threshold while staying silent on the type restriction, which is the more consequential limit.

Validated: applying this keeps the full suite at **310 passed**.

```suggestion
    def _parse_timestamp(self, value: Any) -> datetime:
        """Parse timestamp from various formats.

        Delegates to SessionMapper.parse_timestamp, which additionally handles
        ISO-8601 and string-encoded nanosecond epochs (the OTLP/JSON encoding
        for int64 fields, as produced on the CloudWatch/ADOT path).
        """
        return self.parse_timestamp(value)
```

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,681 @@
...
+                span_session_id = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
+                if span_session_id and str(span_session_id) == session_id:
+                    traces_with_session.add(trace_id)
+
+            # Include all spans from traces that contain at least one matching span.
+            # Exclude entire traces where no span matches the session_id (even if some
+            # spans in that trace have no session_id tag).
+            for trace_id, trace_spans in all_traces.items():
+                if trace_id in traces_with_session:
+                    traces_by_id[trace_id] = trace_spans
```
strandly-the-agent: ## 🟡 Trace-level filtering leaks across sessions when one trace carries two different session tags

A trace is included when *any* one span matches, and then **every** span in it is kept — including spans explicitly tagged with a *different* session. Asked for `session-123`:

```
two spans sharing trace 'tSHARED', tagged session-123 and session-456
  -> got content: 'msg for session-123'
  -> got content: 'msg for session-456'   # <- another session's data

control: same two sessions in separate traces -> only session-123   (correct)
```

`StrandsInMemorySessionMapper` pins the opposite behaviour for exactly this shape (`test_strands_in_memory_mapper.py:434-461`), and this PR's own tests only place the second session in a *different* trace, so nothing currently catches it.

The rule that satisfies both — *own tag wins; an untagged span inherits its trace's match* — also keeps all four of this PR's trace-level filtering tests passing. Validated: full suite stays at **310 passed**.

```suggestion
            for trace_id, trace_spans in all_traces.items():
                if trace_id not in traces_with_session:
                    continue
                for span in trace_spans:
                    attrs = span.get("attributes", {})
                    own = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
                    # Own tag wins; an untagged span inherits its trace's match.
                    if own and str(own) != session_id:
                        continue
                    traces_by_id[trace_id].append(span)
```

This is separate from the bigger question of whether trace-level filtering is the right policy at all (raised in the review body) — this fix doesn't need to wait on that decision.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,681 @@
...
+        """Process tool result content blocks."""
+        result: list[TextContent | ToolResultContent] = []
+        for item in content_list:
+            if "toolResult" in item:
+                tool_result = item["toolResult"]
+                result_text = tool_result.get("content", "")
+                if isinstance(result_text, list):
+                    result_text = "\n".join(
+                        block.get("text", "") for block in result_text if isinstance(block, dict) and "text" in block
+                    )
```
strandly-the-agent: ## 🟡 This drops `json`/`image`/`document` blocks that the shared helper it duplicates handles

This inlines a text-only version of `utils.join_tool_result_content` (`utils.py:38-83`), which `strands_in_memory_session_mapper.py` already imports and which is this repo's contract for Bedrock-style `toolResult.content`:

| `toolResult.content` | `join_tool_result_content` | this method |
|---|---|---|
| `[{"text": "a"}]` | `'a'` | `'a'` |
| `[{"json": {"result": 42}}]` | `'{"result": 42}'` | `''` ❌ |
| `[{"image": …}]` | `'[image]'` | `''` ❌ |
| `[{"text": "a"}, {"json": {"k": "v"}}]` | `'a\n{"k": "v"}'` | `'a'` ❌ |

This isn't a judgment call — `test_strands_in_memory_mapper.py:613` already pins the json-block behaviour as correct, so the same input maps two different ways depending on which mapper handled it. A tool that returned structured JSON reaches the judge as an empty string.

Validated: with the change below plus `from .utils import join_tool_result_content` at the top of the file, the full suite stays at **310 passed** and there's no circular import (`utils` doesn't import the mappers).

```suggestion
                if isinstance(result_text, list):
                    result_text = join_tool_result_content(result_text)
```

strandly-the-agent (commented):
**TL;DR — one real blocker, otherwise this is close.** `chat` spans handle the current OTel GenAI unified event correctly, but `execute_tool` and `invoke_agent` don't — so a spec-conformant producer (including a real Strands agent under `gen_ai_latest_experimental`) maps to a **silently empty** session, i.e. a confidently-wrong eval score rather than an error. Everything `jjbuck` asked for is genuinely fixed — I checked by mutation-testing each fix rather than trusting the "fixed" replies. Separately: **#341** is open and edits the same `detect_otel_mapper` with an overlapping design; worth reconciling before either lands.

@poshinchen — your two questions:

**(a) Does it work per OTel semconv, locally and from CloudWatch/ADOT?** Yes for a spec-conformant producer, on both paths, with one gap. I built a real OTel-SDK local capture of the spec's own canonical examples (`chat` via attributes and via the unified event, `execute_tool` success/failure, `invoke_agent`, a nested 3-level trace) and ran a synthesized ADOT record set end to end through the real `CloudWatchLogsParser` → `detect_otel_mapper` → `map_to_session`. All round-tripped faithfully, and camelCase/snake_case handling on the CloudWatch path is clean — zero unaccounted keys. The gap is the 🔴 below. One CloudWatch-specific caveat worth fixing regardless: OTLP/JSON encodes int64 as decimal **strings**, and `_parse_timestamp` doesn't handle strings (🟡 below). I could not confirm from AWS docs whether real `aws/spans` records land as strings or numbers — one Logs Insights query against a live log group would settle it, and the fix is a one-line deletion either way.

**(b) Deprecate `StrandsInMemoryMapper`, or add a `StrandsMapper`?** **Neither yet.** The "they take different input types" argument for keeping two mappers does collapse — `readable_spans_to_dicts` is lossless for everything either mapper reads. So rather than argue it, I built the deprecation shim (`readable_spans_to_dicts` → `GenericGenAISessionMapper`) and ran the repo's own 21-test `StrandsInMemory` corpus through it: **12 fail — 3 test-rewrite noise, 3 the filtering divergence below, and 6 real data loss** (the same unified-event gap as the 🔴, hitting from the other direction). Neither mapper is a superset of the other: on four `tool_call_response` shapes each gets 3 of 4 right, and it's a *different* 3.

> Recommendation: **extract the shared semconv parsing into common helpers taking the union of both behaviours, and keep both classes public.** Revisit deprecation once `StrandsInMemorySessionMapper` has nothing Strands-specific left — using the `CloudWatchSessionMapper` precedent (`.. deprecated::` + `warnings.warn(..., DeprecationWarning)`, `cloudwatch_session_mapper.py:6-9,58-66`). A third `StrandsMapper` adds surface without removing any. This is a follow-up, **not a blocker for #309** — the PR adds one class and one routing branch, removes nothing, and forecloses none of these options.

<details>
<summary>✅ Verification ledger — what I actually ran</summary>

- **Reviewed:** head `7581b108f9e250f6f47dd61e61826e7fa79cdd27` vs merge base `cc0a3916`. `reviewDecision: CHANGES_REQUESTED`; all 11 review threads currently show resolved.
- **`python3 -m pytest tests/strands_evals/mappers/ -q` → 310 passed.** The `test_detects_strands_in_memory_format` failure from an earlier round is genuinely fixed — not re-raising it. `ruff check` clean; `mypy` clean on 12 source files.
- **🔴 reproduced first-hand** with a real Strands `Agent` (scripted model, one tool), same code both runs, only `OTEL_SEMCONV_STABILITY_OPT_IN` differing:
  ```
  legacy (default):            ToolExecutionSpan result='Seattle: 62F...'  AgentInvocationSpan response='REAL STRANDS ANSWER...'
  gen_ai_latest_experimental:  ToolExecutionSpan result=''                 AgentInvocationSpan prompt='' response=''
  ```
  Corroborated in the SDK: `strands/telemetry/tracer.py` emits `gen_ai.client.inference.operation.details` in `end_tool_call_span` (`:579`) and `end_agent_span` (`:774`), both gated on `use_latest_genai_conventions` (`:122`).
- **Reachability confirmed without Strands and without a mixed batch** — a plain custom-scope producer emitting the spec's canonical event, through ordinary auto-detection: `chat` → correct `InferenceSpan`; `execute_tool` → `args={} result=''`; `invoke_agent` → `prompt='' response=''`.
- **All three 🟡 suggestion blocks below were applied and validated** — full suite stays at **310 passed**, no circular import, and each repro flips to correct output. They're safe to accept as-is.
- **`jjbuck`'s 5 items verified by mutation testing** (revert the fix → confirm a test actually fails): `error.type` precedence → 3 tests fail; `operation.details` (chat) → 8; trace-level filtering → 5; agent response from `gen_ai.output.messages` → 3; `role:"tool"` → 2. All genuinely fixed. **One nuance for `jjbuck`:** item 2 landed for `chat` spans only — the 🔴 is the unfixed remainder of that same thread.
- **Not re-raised** (all confirmed fixed): ambiguous `== 0 or == 0` asserts, broad `except Exception`, the `1e12` heuristic documentation, `test_create_agent_span_skipped`.
- Fixture provenance spot-checked: the two "live" fixtures read as genuine captures, and `otel_genai_convention_spans.json` is labelled hand-crafted in its own docstring. No concern.

</details>

### Questions

**Worth an answer before merge:**

1. **Is evaluating a Strands agent under `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental` a supported combination?** This sets much of the 🔴's weight. If yes, it's squarely blocking. If that's still experimental territory nobody runs evals against, it drops to 🟡 — but then it's worth saying so explicitly in the mapper's docstring rather than leaving it implied.
2. **Trace-level vs span-level session filtering.** This mapper filters at trace level (`:78-97`), which contradicts `StrandsInMemorySessionMapper`'s deliberate span-level design from #51 (*"only add the traces with matched `session.id` in the span attributes"*). Two mappers now answer "what belongs to this session" differently, in the same release. Is trace-level the intended new family policy? Either answer is fine — it's shipping the contradiction silently that gets expensive to unpick.
3. **#341.** It adds a second dict-span mapper for the same root problem (`StrandsInMemorySessionMapper` can't read dicts) and edits the same `detect_otel_mapper`, with an overlapping routing condition. Neither thread mentions the other. Which one lands, or how do they coexist?

**Non-blocking:**

4. **Would `OtelGenAISessionMapper` read better than `GenericGenAISessionMapper`?** Every sibling is named for its producer, but `OpenInferenceSessionMapper` is precedent for naming after a *convention*. Since `gen_ai.operation.name` is the spec's own `Required` field, this mapper arguably handles the most-specified input in the family rather than the least — "Generic" may have the polarity backwards. It's unreleased, so renaming is free right now.
5. **Should the routing predicate be "is dict-shaped" rather than "has `gen_ai.operation.name`"?** (`utils.py:159`) The real `StrandsInMemorySessionMapper` incompatibility is the dict-vs-`ReadableSpan` shape, which is why a dict span with an unrecognized scope and *no* `gen_ai.operation.name` still crashes identically (pre-existing — this PR narrows the class rather than closing it). Is narrowing an acceptable interim step here?

<details>
<summary>Appendix — non-blocking (14)</summary>

**⚪ Nits (new, low impact):**
- `:424-425` iterates `parts` without the `isinstance(part, dict)` guard its near-identical sibling has at `:445`. An `AttributeError` here is in none of the file's narrowed `except` tuples, so it escapes the per-span guard at `:137` and aborts the whole `map_to_session`, taking unrelated traces with it. Reach is contrived (nothing emits a bare string in `parts`) — worth the one-line guard for symmetry.
- `:500-501` and `:527-528` are bare `except: pass` with no logging, unlike the four narrowed-and-logged sites in this same file (`:137,219,275,338`). Malformed JSON in `gen_ai.input/output.messages` yields an empty session with no diagnostic at all. A `logger.warning` at both would match the file's own pattern.
- Explicit `trace_id: None` (as opposed to a missing key) bypasses the `"unknown"` default at `:76,85` → `Trace(trace_id=None)` → pydantic `ValidationError`. Neither real ingestion path produces this.
- `role:"tool"` is implemented twice, near-identically (`:445-475` and `:581-611`), each separately tested — drift risk rather than a coverage gap. Reverting one didn't fail the other's tests.
- Only 3 of ~18 spec `gen_ai.operation.name` values are handled (`:125,129,133`), with no `else`/log — `create_agent`, `embeddings`, `generate_content` etc. are silently dropped. An `else: logger.debug(...)` is cheap.
- `available_tools` (`:311-323`) keeps tool *names* only, dropping the `description`/`parameters` that `ToolConfig` and `gen_ai.tool.definitions` both carry — a judge asked "did the agent pick the right tool?" can't assess selection from bare names.
- `role:"user"` + `tool_call_response` reads `part.get("result")` (`:430`); the spec field is `response`. Nothing is broken today — PydanticAI's actual shape really does use `result`, and the mapper handles both keyed on role — but it's a forward risk if PydanticAI converges on the spec.
- `known_scopes` is rebuilt inline at `utils.py:154` rather than living in `constants.py` beside the `SCOPE_*` constants it's derived from. Currently correct, but this is precisely the seam where the Strands-routing regression was introduced (twice) earlier in this PR's history.
- `utils.py:98`'s new docstring clause *"…or ReadableSpan objects"* is false — real `ReadableSpan`s crash at `utils.py:148`→`:173`. The crash is pre-existing; the *claim* is new. Suggest reverting to the base wording. Relatedly, `test_raw_readable_spans_route_to_strands_mapper` (`test_utils.py:236-240`) passes a bare `MagicMock()`, which survives only because `MagicMock().get()` fakes a dict — it proves "a non-dict that doesn't error on `.get()` falls through", not what its name says. Rename it, or use a real `ReadableSpan` and mark it `xfail`.
- No CloudWatch/ADOT fixture anywhere in the new tests, though `cloudwatch_provider.py:162` is a real caller. I covered the path empirically (see answer (a)); the suite doesn't.
- `SKILL.md:140-145`'s mapper list and `AGENTS.md`'s Directory Structure (which `AGENTS.md:167` says to update) aren't touched, though the PR checklist has "I have updated the documentation accordingly" checked.

**Pre-existing — reproduces identically on `cc0a3916`, not this PR's problem (one line each):**
- The empty-string (rather than `None`-guarded) `AgentInvocationSpan` policy is shared with `StrandsInMemorySessionMapper`; `ADKOtelSessionMapper` and `CloudWatchSessionMapper` guard with `None` instead. This PR copied the weaker of two coexisting policies — worth one follow-up issue across the family: *"Unify empty-field policy for AgentInvocationSpan across mappers"*.
- `ToolResult.error` is populated correctly by every mapper, but no evaluator formatter ever renders `.error` — only `.content` (`evaluator.py:190,210,235`, untouched here). So `jjbuck`'s error-contract fix has **zero visible effect on any judge today**. Worth a separate issue: *"Evaluator prompts never surface ToolResult.error to the judge"*.
- No time-based sort in this mapper — spans/traces come out in raw input order, which a real exporter doesn't guarantee is chronological (verified: `get_finished_spans()` returns children before parents). Same gap in `StrandsInMemorySessionMapper`, but `ADKOtelSessionMapper`/`CloudWatchSessionMapper` already sort — a precedent worth following in a follow-up.

**Checked and clean:** all 5 semconv URLs in the diff resolve and match their adjacent claims · legacy per-message event support is a deliberate compatibility shim, not dead debt (Strands' default mode still emits those) · parallel tool-call correlation is correct (each `execute_tool` span self-contains call+result) · no mutable default args · duplicate `span_id`s handled benignly · the structured-vs-JSON-string concern is moot in Python, since `opentelemetry-sdk` drops nested attribute values on spans and events alike.

</details>

---

Three rounds in and every substantive item has been addressed with a real regression test each time — the convention coverage here is genuinely thorough. This is solid work; the one blocker is a gap in the same feature, not a new direction. Flagging as usual that I'm an experimental agent — a human should confirm the 🔴 before it gates anything.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: I think this is worth fixing because it can affect other agents that opt into the unified event.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: The base session mapper's `parse_timestamp` was added in https://github.com/strands-agents/evals/pull/326. You can call it directly in your mapper instead of writing/calling a helper `_parse_timestamp`.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,681 @@
...
+
+        # Group by trace_id, optionally filtering by session.id
+        traces_by_id: dict[str, list[dict]] = defaultdict(list)
+
+        any_span_has_session_id = any(
+            "session.id" in span.get("attributes", {}) or "gen_ai.conversation.id" in span.get("attributes", {})
+            for span in spans
+        )
+
+        if not any_span_has_session_id:
```
liramon2, line 92: I feel like this branch can be simplified. If you proceed with the existing code, you can do something like:
```
all_traces: dict[str, list[dict]] = defaultdict(list)
traces_with_session: set[str] = set()

for span in spans:
    trace_id = span.get("trace_id", "unknown")
    all_traces[trace_id].append(span)
    if any_span_has_session_id:
        attrs = span.get("attributes", {})
        span_session_id = attrs.get("gen_ai.conversation.id") or attrs.get("session.id")
        if span_session_id and str(span_session_id) == session_id:
            traces_with_session.add(trace_id)

if any_span_has_session_id:
    traces_by_id = {tid: s for tid, s in all_traces.items() if tid in traces_with_session}
else:
    traces_by_id = dict(all_traces)
```
This removes one of the loops and keeps the conditional logic small. If you follow the strandly suggestion, this branch becomes dead code.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,681 @@
...
+        messages: list[UserMessage | AssistantMessage] = []
+
+        # Parse input messages
+        input_msgs_raw = attrs.get("gen_ai.input.messages", "")
+        if input_msgs_raw:
+            try:
+                input_msgs = json.loads(input_msgs_raw) if isinstance(input_msgs_raw, str) else input_msgs_raw
+                if isinstance(input_msgs, list):
+                    for msg in input_msgs:
+                        role = msg.get("role", "")
```
liramon2: I see a lot of overlap between the message list conversion logic in `_parse_attribute_messages` and `_parse_operation_details_event`. Could be worth extracting into a helper.

ybdarrenwang: done in the latest revision

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: done in the latest revision

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: done in the latest revision

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: done in the latest revision

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: done in the latest revision

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
ybdarrenwang: fixed and added unit tests in the latest revision

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,617 @@
...
+    def _parse_timestamp(self, value: Any) -> datetime:
+        """Parse timestamp from various formats.
+
+        Delegates to SessionMapper.parse_timestamp, which additionally handles
+        ISO-8601 and string-encoded nanosecond epochs (the OTLP/JSON encoding
+        for int64 fields, as produced on the CloudWatch/ADOT path).
+        """
+        return self.parse_timestamp(value)
+
+    def _parse_json_attr(self, attributes: Any, key: str, default: str = "[]") -> Any:
```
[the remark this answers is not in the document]
ybdarrenwang: This can wait til we consolidate strands mapper and this generic mapper.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,617 @@
...
+                logger.warning(
+                    "trace_id=%s span_id=%s | Failed to convert span: %s",
+                    span.get("trace_id", "?"),
+                    span.get("span_id", "?"),
+                    e,
+                )
+
+        return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
+
+    def _parse_timestamp(self, value: Any) -> datetime:
```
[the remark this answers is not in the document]
ybdarrenwang: fixed in latest revision

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,617 @@
...
+            agent_response=agent_response,
+            available_tools=available_tools,
+            metadata={},
+        )
+
+    # =========================================================================
+    # Content Processing Helpers
+    # =========================================================================
+
+    def _convert_message_list_from_event(self, event_attrs: dict) -> list:
```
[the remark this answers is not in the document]
ybdarrenwang: fixed in latest revision

liramon2:
@strandly-the-agent Check if this is safe to merge.

strandly-the-agent:
**Short answer: yes, the code is in good shape to merge — but two things need a human, and I'd land one 6-line patch first.** All four findings from my last review are genuinely fixed *and* pinned by tests that fail when I revert the fix (mutation-tested, not taken on trust). CI is green across all 11 unit jobs, it merges cleanly into current `main`, and the #341 collision I flagged is closed. What's left is one root cause in `_convert_message_list` that puts Python-`repr` noise into judge prompts — **not a regression**, so it's your call whether it blocks.

<details>
<summary>✅ Verification ledger — head <code>d61aee8</code></summary>

**Mechanics**
- Rebased onto current `main` (`21d3f73`); merge base *is* `main`. Trial merge: **clean**. `mergeable: MERGEABLE`.
- Diff vs `main` is only **9 files / +2308 −19**, and just 3 source files (`generic_gen_ai_session_mapper.py` 608 lines, `utils.py` +14, `__init__.py` +2). The rest is tests/fixtures.
- CI: **11/11 unit jobs SUCCESS** (Py3.10–3.14 × linux/windows, macOS 3.13), Lint SUCCESS, `validate-pr-title` SUCCESS, CI Gate SUCCESS. `Run integration tests` is **WAITING** on maintainer approval — that's the only reason the rollup shows PENDING.
- Local: `tests/strands_evals/mappers/` → **325 passed**; mappers+extractors+types → **457 passed**; `ruff check` clean.
- `mypy` reports 2 errors (missing `langfuse` stubs) and the broader `evaluators/providers/telemetry/cli` dirs give `9 failed, 519 passed, 49 errors` — **byte-identical on `origin/main`**, so environmental/pre-existing, not this PR. I checked rather than assumed.

**My four findings — all fixed, all mutation-pinned**
| Finding | Fixed | Test actually pins it? |
|---|---|---|
| 🔴 `operation.details` unread on `execute_tool`/`invoke_agent` | ✅ now in all 3 converters (`:203`/`:266`/`:349`); agent prompt+response correct on a real Strands agent | ✅ each branch mutated separately → 2 killing tests each |
| 🟡 `_parse_timestamp` re-implementation | ✅ override deleted, uses inherited `self.parse_timestamp` (`:129-130`) | ✅ `test_string_timestamp_in_span_produces_correct_time` (mutation → `assert 2026 == 2023`) |
| 🟡 cross-session filtering leak | ✅ fixed, and **no over-correction** — untagged siblings still inherit their trace | ✅ 2 killing tests |
| 🟡 `join_tool_result_content` dropped json/image blocks | ✅ in `_process_tool_results` (`:600`) | ⚠️ 2 of 3 (see nit below) |

Also confirmed fixed: the unresolved-thread `status: None` crash — `:236-237` now guards with `span.get("status") or {}` + `isinstance`. `None` / `[]` / bare-string no longer crash, and precedence (`error.type` > span `ERROR` > legacy `gen_ai.tool.status`) is intact, with `"success"` still yielding `error=None`. That thread can be marked resolved.

</details>

### The one thing worth fixing: tool results reach the judge as a Python `repr`

`_convert_message_list`'s two `role:"tool"` branches disagree. The **flat** branch (`:507-514`) joins list-valued responses correctly; the **parts** branch does `content=str(resp)` at **`:495`**. Strands emits the parts form, with `response` as a Bedrock content-block list — so, on a real agent under `OTEL_SEMCONV_STABILITY_OPT_IN=gen_ai_latest_experimental`:

```
StrandsInMemory (ground truth):  result='Seattle: 62F, cloudy with light rain'
GenericGenAI    (this PR):       result="[{'text': 'Seattle: 62F, cloudy with light rain'}]"
```

That string goes straight into judge prompts (`evaluators/evaluator.py:190,220,245` interpolate `tool_result.content`). **Two honest caveats on severity:** the text is still *present*, just wrapped in repr noise — not data loss; and it is **not a regression** — the head I reviewed on 2026-08-04 produced the JSON-string `'[{"text": ...}]'` via a different path, so this is a lateral move, and legacy-convention mode is clean either way. That's why I'm not calling it a blocker.

Same root cause, two siblings: `:471` (`role:"user"`) reads `part.get("result")` where the spec and Strands emit `response`, so `InferenceSpan` tool results come out `''` — **latent only**, because `InferenceSpan` is never consumed downstream (`trace_extractor.py` reads only `AgentInvocationSpan` and `ToolExecutionSpan`). And a non-dict entry in a message list raises an `AttributeError` that escapes every `except` and aborts the whole `map_to_session`, taking healthy traces with it — contrived input, but the blast radius *grew* vs the previous head, where it was harmless.

<details>
<summary>The 6-line patch — I applied and validated it (325 pass, ruff clean, output matches ground truth exactly)</summary>

```python
# (a) :468  role:"user" tool_call_response — accept the spec's `response`, join lists
    elif part.get("type") == "tool_call_response":
        _r = part.get("response", part.get("result", ""))
        if isinstance(_r, list):
            _r = join_tool_result_content(_r)
        elif isinstance(_r, dict):
            _r = json.dumps(_r)
        user_content.append(ToolResultContent(content=str(_r) if _r else "", ...))

# (b) :488  role:"tool" parts branch — join lists like the flat branch already does
    resp = part.get("response", part.get("result", ""))
    if isinstance(resp, dict):
        resp = json.dumps(resp)
    elif isinstance(resp, list):
        resp = join_tool_result_content(resp)

# (c) :457  don't let one malformed message kill the whole session
    for msg in msg_list:
        if not isinstance(msg, dict):
            continue
```

After applying, both the tool span and the inference span match `StrandsInMemorySessionMapper` exactly:
```
GROUND TRUTH: result='Seattle: 62F, cloudy with light rain'
PATCHED:      result='Seattle: 62F, cloudy with light rain'
325 passed · ruff: All checks passed!
```
I reverted it — the tree is untouched. Worth a quick amend if you're still at the keyboard; otherwise a fast follow-up issue.

**Why the tests missed it:** coverage is 84%, and lines **`508-514` — the only branch that joins a list-valued response — are never executed by any test.** No test or fixture in the repo passes a list-valued `response`/`result`; they're all strings or dicts. One test to add per path would close this permanently.

**⚪ Related nit:** `test_tool_result_json_block_preserved` (`test_generic_gen_ai_session_mapper.py:1041`) passes against the *broken* code — it asserts `"42" in content` and `"result" in content`, both satisfied by the Python repr too. Its two siblings (`"[image]" in content`, `"\n" in content`) are genuinely strong. Assert the exact serialized string instead.

</details>

### Two things only a human can clear

1. **`reviewDecision` is still `CHANGES_REQUESTED`** from `jjbuck`'s review. All five of his threads are resolved and I re-verified each fix by mutation testing, so the review looks stale rather than outstanding — but it needs an explicit dismissal or re-approval from him, which I can't and shouldn't do.
2. **`Run integration tests` is awaiting approval.** Given the mapper is what changed and the integ suite is the only thing that exercises the CloudWatch/provider path end to end, I'd approve that run before merging rather than after.

Good cleanup on this round — the three commits fixed everything raised and the new tests genuinely hold the fixes in place, which is the part that usually gets skipped. Usual caveat that I'm an experimental agent: the ledger above is what I actually ran, and the merge call is yours.

cc `liramon2`

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,608 @@
...
+                        )
+                # Also handle flat content string (no parts)
+                if not assistant_content and msg.get("content"):
+                    assistant_content.append(TextContent(text=str(msg["content"])))
+                if assistant_content:
+                    messages.append(AssistantMessage(content=assistant_content))
+
+        return messages
+
+    def _parse_attribute_messages(self, attrs: dict) -> list[UserMessage | AssistantMessage]:
```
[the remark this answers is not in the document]
ybdarrenwang: removed `_parse_attribute_messages` in latest revision

tests/strands_evals/mappers/test_generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,1244 @@
...
+        "status": {"code": "OK"},
+        "span_events": span_events or [],
+    }
+
+
+class TestToolExecutionSpan:
+    def setup_method(self):
+        self.mapper = GenericGenAISessionMapper()
+
+    def test_tool_span_from_attributes(self):
```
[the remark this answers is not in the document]
ybdarrenwang: nit; and other existing unit tests also do field-by-field assertion

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,607 @@
...
+        attrs = span.get("attributes", {})
+
+        tool_name = str(attrs.get("gen_ai.tool.name", ""))
+        tool_call_id = str(attrs.get("gen_ai.tool.call.id", ""))
+
+        # Prefer error.type (current OTel GenAI convention for failed operations),
+        # fall back to gen_ai.tool.status / tool.status for legacy compatibility.
+        # https://github.com/open-telemetry/semantic-conventions-genai/blob/main/model/gen-ai/spans.yaml#L522-L562
+        error_type = attrs.get("error.type", "")
+        span_status = span.get("status", {}).get("code", "")
```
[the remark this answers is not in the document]
ybdarrenwang: fixed in latest revision

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,594 @@
...
+                    if isinstance(response, dict):
+                        response = json.dumps(response)
+                    elif isinstance(response, list):
+                        parts_text = []
+                        for block in response:
+                            if isinstance(block, dict) and block.get("text"):
+                                parts_text.append(block["text"])
+                            elif isinstance(block, str):
+                                parts_text.append(block)
+                        response = "\n".join(parts_text) if parts_text else str(response)
```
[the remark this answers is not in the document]
ybdarrenwang: fixed in latest revision

liramon2:
@strandly-the-agent Check if this is safe to merge.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,588 @@
...
+                elif operation_name == "execute_tool":
+                    tool_span = self._convert_tool_execution_span(span, span_info)
+                    if tool_span:
+                        converted_spans.append(tool_span)
+                elif operation_name == "invoke_agent":
+                    agent_span = self._convert_agent_invocation_span(span, span_info)
+                    if agent_span:
+                        converted_spans.append(agent_span)
+            except (KeyError, TypeError, ValueError, json.JSONDecodeError) as e:
+                logger.warning(
```
liramon2, line 146: Nit: the sentence after '|' should be lowercase to match conventions (eg. Failed -> failed). This applies to all warnings in the file.

strandly-the-agent:
**Yes — safe to merge on the substance. `9ac9405` closes the last thing I had open, and I verified the fix rather than assuming it.** The only remaining gates are the two human ones from my last comment, both unchanged.

The tool-result fix landed in full, and it went further than I suggested — the flat branch now uses the shared helper too, and the false `utils.py:98` "or ReadableSpan objects" docstring clause got reverted (that was an appendix nit from my first review). On a real Strands agent, the generic mapper is now **byte-identical to `StrandsInMemorySessionMapper` in both convention modes**:

```
legacy:                      result='Seattle: 62F, cloudy with light rain'   ✅ matches ground truth
gen_ai_latest_experimental:  result='Seattle: 62F, cloudy with light rain'   ✅ matches (was "[{'text': ...}]")
```

<details>
<summary>✅ Verification ledger — head <code>9ac9405</code></summary>

**Attacked the fix** (the same defect has survived consecutive fixes on this PR, so I probed the new code rather than re-reading it). Every tool-response shape now agrees with `join_tool_result_content` — across **both** branches (`parts` and flat), **both** keys (`response` and legacy `result`), and **both** roles (`user` and `tool`):

| response shape | mapped content |
|---|---|
| `"plain"` | `'plain'` |
| `[{"text": "hello"}]` | `'hello'` ✅ |
| `[{"json": {"k": 1}}]` | `'{"k": 1}'` ✅ |
| `[{"image": …}]` | `'[image]'` ✅ |
| `["a", "b"]` | `'a\nb'` ✅ |
| `{"k": 1}` (dict) | `'{"k": 1}'` |
| `[]` / `[{"unknown": 1}]` | `''` (clean, no repr) |

Also confirmed: a non-dict entry in a message list no longer aborts the whole `map_to_session` — the sibling message still converts. And the `InferenceSpan` tool-result path (the latent `result`-vs-`response` key divergence) now matches ground truth too.

**The one change I hadn't asked for, checked closely.** `_parse_attribute_messages` was deleted and the attribute fallback now calls `_convert_message_list_from_event(attrs)`. That's the kind of swap that quietly regresses a path, so I diffed the behaviour: same two keys, same order, same JSON-or-list handling, same guard, same delegation — **pure de-duplication, zero behaviour change**. Confirmed against the real fixtures through the attribute path: PydanticAI → 2 inference spans / 6 content blocks, `otel_genai_convention` → 2 / 11, AutoGen → 0 (expected; that capture genuinely carries no message content). Only nit: the name still says `_from_event` though it now also serves span attributes — cosmetic.

**Gates.** Merge base *is* `main` (`21d3f73`); trial merge **clean**; `mergeable: MERGEABLE`. CI **17 SUCCESS / 1 skipped / 1 waiting**. Local: `mappers` **332 passed** (was 325 — 7 new tests), mappers+extractors+types **464 passed**, `ruff` clean. The `evaluators`/`providers`/`telemetry`/`cli` dirs give `9 failed / 519 passed / 49 errors` — **exactly the baseline I measured on `origin/main`**, i.e. missing optional `langfuse` and absent AWS creds, not this PR.

**Remaining 3 unresolved threads are non-blocking style/typing suggestions** on the new code: `-> X | None` annotations that never return `None`; per-field vs whole-object test assertions; and `hasattr(c, "text")` duck-typing where the concrete types are already imported. The last one is the one with real substance — it's the shape I probed hardest above, and it behaves correctly today, so it's a readability/robustness cleanup rather than a defect.

</details>

### Still only these two, and neither is mine to do

1. **`reviewDecision` is `CHANGES_REQUESTED`** from `jjbuck`. All his threads are resolved and I've now mutation-verified every fix across two rounds, so it reads as stale — but it needs his dismissal or re-approval.
2. **`Run integration tests` is still awaiting approval.** Worth running before merge rather than after, since the mapper is what changed and integ is the only thing exercising the CloudWatch/provider path end to end.

Nice turnaround on that patch — and thanks for pushing back on the repr issue rather than deferring it; the mapper now agrees with the Strands mapper exactly on real traces, which is a much better place to merge from.

cc `liramon2`

jjbuck (approved):

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,588 @@
...
+                elif event_name == "gen_ai.client.inference.operation.details":
+                    # Current OTel GenAI convention: single event with input/output messages
+                    # https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-events.md
+                    details_messages = self._convert_message_list_from_event(event_attrs)
+                    messages.extend(details_messages)
+
+            except (KeyError, TypeError, ValueError, json.JSONDecodeError) as e:
+                logger.warning(
+                    "span_id=%s event=%s | Failed to process event: %s",
+                    span.get("span_id", "?"),
```
poshinchen, line 212: nit: hmm should we do "?" or just unknown?

poshinchen (approved):

PR #313, opened 2026-07-15
fix: fix tool parsing from list

### LINKED ISSUE #312: [BUG] LangfuseProvider._convert_tool_execution loses tool data for Strands OTEL traces (obs.input is list format)
### Checks

### Strands Version

1.1.0

### Strands Evals Version

0.1.15

### Python Version

3.13.7

### Operating System

macOS 15.5 (Darwin 25.3.0)

### Installation Method

pip

### Steps to Reproduce

1. Have a Strands agent trace in Langfuse where tools were called (e.g., `bsm-get-case`, `search-listings-items`)

2. The TOOL observations arrive in this format (Strands SDK → OTEL → Langfuse):

```json
{
  "type": "TOOL",
  "name": "bsm-get-case",
  "input": [{"role": "tool", "content": "{\"caseId\": \"abc123\", \"marketplaceId\": \"ATVPDKIKX0DER\"}"}],
  "output": {"message": "[{\"text\": \"Data stored under key 'bsm-get-case' (9 fields). Use get_workflow_data(...) to retrieve.\"}]"}
}

3. Run the provider's conversion:

provider = LangfuseProvider.__new__(LangfuseProvider)

class FakeObs:
    type = "TOOL"
    name = "bsm-get-case"
    input = [{"role": "tool", "content": '{"caseId": "abc123", "marketplaceId": "ATVPDKIKX0DER"}'}]
    output = {"message": '[{"text": "Data stored under key bsm-get-case (9 fields)"}]'}
    id = "obs1"
    trace_id = "t1"
    parent_observation_id = "p1"
    start_time = None
    end_time = None
    metadata = {}

span = provider._convert_tool_execution(FakeObs(), "session-1")
print(f"tool_name: {span.tool_call.name}")
print(f"arguments: {span.tool_call.arguments}")
print(f"result: {span.tool_result.content}")

4. Output:

tool_name: bsm-get-case                          # ✓ correct
arguments: {'input': "[{'role': 'tool'...}]"}     # ✗ mangled — should be {"caseId": "abc123", ...}
result: {'message': '[{"text": "Data stored..."}]'} # ✗ platform confirmation, not actual tool data

### Expected Behavior

```markdown
The provider should parse the Strands OTEL TOOL format correctly:

- `tool_call.arguments` should be `{"caseId": "abc123", "marketplaceId": "ATVPDKIKX0DER"}` (parsed from `obs.input[0].content`)
- `tool_result.content` should contain the actual tool response data

In Strands OTEL format:
- `obs.input` is a list of message dicts: `[{"role": "tool", "content": "<JSON args>"}]`
- `obs.output` is often a platform confirmation (blackboard pointer or success message), NOT the actual data

The actual tool arguments live in `obs.input[0].content` (JSON string).

### Actual Behavior

In `_convert_tool_execution` (line 459-474), the `elif obs.type == "TOOL"` branch:

```python
tool_name = obs.name or ""           # ✓ works
if isinstance(obs_input, dict):       # not hit — obs.input is a list
    tool_arguments = obs_input
elif isinstance(obs_input, str):      # not hit
    ...
elif obs_input:                       # ← HIT: obs.input is a list
    tool_arguments = {"input": str(obs_input)}  # stringifies the entire list
Then _parse_tool_result(obs.output) receives {"message": "[...]"} which has no "result" key and no "content" key → falls through to str(obs_output).

Impact: FaithfulnessEvaluator and other evaluators compare agent output against tool history that contains NO actual data — only platform confirmations and stringified lists. They cannot detect fabrication or verify groundedness.

Tested on 70 production traces: FaithfulnessEvaluator scored 1.0 on 69/70 traces (including traces with confirmed fabrication) because tool results contained no real data to check against.

### Additional Context

```markdown
This affects all Strands agents instrumented via the standard OTEL → Langfuse pipeline. The format is:

- `obs.type = "TOOL"`
- `obs.name = "tool-name"` (correct)
- `obs.input = [{"role": "tool", "content": "<JSON-encoded arguments>", "id": "tooluse_..."}]`
- `obs.output = {"message": "[{\"text\": \"Success/confirmation\"}]", "id": "tooluse_..."}`

The native Strands SPAN format (obs.type == "SPAN", name starts with "execute_tool") uses a different structure where `obs.input = {"name": ..., "arguments": ..., "toolUseId": ...}` — this one is handled correctly. The issue is only with the TOOL-type observations from OTEL instrumentation.


### Possible Solution

Add list-format handling in the `elif obs.type == "TOOL"` branch:

```python
elif obs.type == "TOOL":
    tool_name = obs.name or ""
    if isinstance(obs_input, dict):
        tool_arguments = obs_input
    elif isinstance(obs_input, list):
        # Strands OTEL format: [{"role": "tool", "content": "<JSON args>", "id": "..."}]
        for item in obs_input:
            if isinstance(item, dict) and "content" in item:
                content = item["content"]
                if isinstance(content, str):
                    try:
                        tool_arguments = json.loads(content)
                    except (json.JSONDecodeError, ValueError):
                        tool_arguments = {"input": content}
                    break
        else:
            tool_arguments = {"input": str(obs_input)}
    elif isinstance(obs_input, str):
        ...
For obs.output, the _parse_tool_result method should handle {"message": "[...]"} format:


if isinstance(obs_output, dict):
    if "message" in obs_output:
        msg = obs_output["message"]
        # Strands format: message is JSON string of [{text: "..."}]
        if isinstance(msg, str):
            try:
                parsed = json.loads(msg)
                if isinstance(parsed, list) and parsed:
                    return parsed[0].get("text", str(parsed)), None
            except (json.JSONDecodeError, ValueError):
                return msg, None
    ...


### Related Issues

#143 #49

### PR DESCRIPTION
**Pull request by poshinchen**
## Description

`LangfuseProvider._convert_tool_execution` lost tool data for Strands agents instrumented via the standard OTEL → Langfuse pipeline. 

For these traces, `obs.type == "TOOL"` observations arrive as:
- `obs.input` = a **list**: `[{"role": "tool", "content": "<JSON args>", "id": "tooluse_..."}]`
- `obs.output` = `{"message": "<JSON string of [{\"text\": ...}]>", "id": "tooluse_..."}`

The `elif obs.type == "TOOL"` branch only handled `dict` and `str` inputs, so a list fell through to `{"input": str(obs_input)}`

### Changes

- Add a `list` branch to `_convert_tool_execution`, delegating to a new `_parse_tool_arguments_from_list` helper.
  - It extracts the first message item's `content`, parses it as JSON when it's a string, and falls back to `{"input": ...}` for non-JSON content.
- Add a `"message"`-key branch to `_parse_tool_result`, delegating to a new `_parse_tool_result_message` helper that JSON-decodes the message and extracts the first `text` block (reusing `_first_text_from_list`), falling back to the raw string when the shape doesn't match.

Existing `dict` / `str` handling and the native Strands `SPAN` (`execute_tool`) path are unchanged.

## Related Issues

Fixes #312

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing
- Tested end-to-end with sending the traces to langfuse and load them back via the provider.
- [s] I ran `hatch run prepare`

## Checklist

----

### DIFF
--- a/src/strands_evals/providers/langfuse_provider.py
+++ b/src/strands_evals/providers/langfuse_provider.py
@@ -543,6 +543,8 @@ class LangfuseProvider(TraceProvider):
                 error = None if status == "success" else (str(status) if status else None)
                 return content, error
             # Strands OTEL format: {"message": "<JSON string of [{\"text\": ...}]>", "id": "..."}
+            # This payload carries no error/status field; failed tool calls surface a
+            # "result"/"status" dict instead, handled by the branch above.
             if "message" in obs_output:
                 return self._parse_tool_result_message(obs_output["message"]), None
             # LangChain ToolMessage format: {"content": "...", "type": "tool", ...}
@@ -568,15 +570,7 @@ class LangfuseProvider(TraceProvider):
         Falls back to the raw string when the JSON does not decode to a
         non-empty list of text blocks.
         """
-        if isinstance(message, str):
-            try:
-                parsed = json.loads(message)
-            except (json.JSONDecodeError, ValueError):
-                return message
-            return self._first_text_from_list(parsed) or message
-        if isinstance(message, list):
-            return self._first_text_from_list(message) or str(message)
-        return str(message)
+        return self._extract_text_from_content(message) or str(message)
 
     def _convert_agent_invocation(self, obs: Any, session_id: str) -> AgentInvocationSpan:
         """Convert an agent observation to an AgentInvocationSpan.
--- a/src/strands_evals/providers/langfuse_provider.py
+++ b/src/strands_evals/providers/langfuse_provider.py
@@ -440,13 +440,18 @@ class LangfuseProvider(TraceProvider):
         Handles two formats:
 
         **Strands** (obs.type == "SPAN", name starts with "execute_tool"):
-            obs.input: ``{"name": "calc", "arguments": {"x": "2+2"}, "toolUseId": "..."}``
-            obs.output: ``"42"`` or ``{"result": "4", "status": "success"}``
+            obs.input: `{"name": "calc", "arguments": {"x": "2+2"}, "toolUseId": "..."}`
+            obs.output: `"42"` or `{"result": "4", "status": "success"}`
 
         **LangChain / universal** (obs.type == "TOOL"):
-            obs.name: tool name (e.g. ``"add_numbers"``)
+            obs.name: tool name (e.g. `"add_numbers"`)
             obs.input: tool arguments (dict or other)
             obs.output: tool result
+
+        **Strands OTEL** (obs.type == "TOOL", from OTEL instrumentation):
+            obs.name: tool name (e.g. `"bsm-get-case"`)
+            obs.input: `[{"role": "tool", "content": "<JSON args>", "id": "tooluse_..."}]`
+            obs.output: `{"message": "<JSON string of [{\"text\": ...}]>", "id": "tooluse_..."}`
         """
         span_info = self._create_span_info(obs, session_id)
         obs_input = obs.input or {}
@@ -461,6 +466,9 @@ class LangfuseProvider(TraceProvider):
             tool_name = obs.name or ""
             if isinstance(obs_input, dict):
                 tool_arguments = obs_input
+            elif isinstance(obs_input, list):
+                # Strands OTEL format: [{"role": "tool", "content": "<JSON args>", "id": "..."}]
+                tool_arguments = self._parse_tool_arguments_from_list(obs_input)
             elif isinstance(obs_input, str):
                 # Try parsing as JSON; LangChain may send stringified dicts
                 try:
@@ -486,16 +494,43 @@ class LangfuseProvider(TraceProvider):
             span_info=span_info, tool_call=tool_call, tool_result=tool_result, metadata=obs.metadata or {}
         )
 
+    def _parse_tool_arguments_from_list(self, obs_input: list) -> dict:
+        """Extract tool arguments from a Strands OTEL TOOL-message list.
+
+        Input format:
+            `[{"role": "tool", "content": "<JSON args>", "id": "tooluse_..."}]`
+
+        The actual arguments live in the first item's `content` field as a
+        JSON string. Returns the parsed dict, or `{"input": <content>}` when
+        the content is not JSON, or `{"input": str(obs_input)}` when no item
+        carries a `content` field.
+        """
+        for item in obs_input:
+            if isinstance(item, dict) and "content" in item:
+                content = item["content"]
+                if isinstance(content, dict):
+                    return content
+                if isinstance(content, str):
+                    try:
+                        parsed = json.loads(content)
+                    except (json.JSONDecodeError, ValueError):
+                        return {"input": content}
+                    return parsed if isinstance(parsed, dict) else {"input": content}
+                return {"input": str(content)}
+        return {"input": str(obs_input)}
+
     def _parse_tool_result(self, obs_output: Any) -> tuple[str, str | None]:
         """Parse tool execution output into (content, error).
 
         Input formats:
-            str:  ``"42"``                                            → ``("42", None)``
-            dict: ``{"result": "4", "status": "success"}``            → ``("4", None)``
-            dict: ``{"result": "...", "status": "error"}``             → ``("...", "error")``
-            dict: ``{"content": "Weather...", "type": "tool", ...}``   → ``("Weather...", None)``
+            str:  `"42"`                                            → `("42", None)`
+            dict: `{"result": "4", "status": "success"}`            → `("4", None)`
+            dict: `{"result": "...", "status": "error"}`             → `("...", "error")`
+            dict: `{"content": "Weather...", "type": "tool", ...}`   → `("Weather...", None)`
                   (LangChain ToolMessage format via Langfuse)
-            None:                                                      → ``("", None)``
+            dict: `{"message": "<JSON of [{\"text\": ...}]>", ...}`  → `("...", None)`
+                  (Strands OTEL TOOL format via Langfuse)
+            None:                                                    → `("", None)`
         """
         if isinstance(obs_output, str):
             return obs_output, None
@@ -507,6 +542,9 @@ class LangfuseProvider(TraceProvider):
                 status = obs_output.get("status", "")
                 error = None if status == "success" else (str(status) if status else None)
                 return content, error
+            # Strands OTEL format: {"message": "<JSON string of [{\"text\": ...}]>", "id": "..."}
+            if "message" in obs_output:
+                return self._parse_tool_result_message(obs_output["message"]), None
             # LangChain ToolMessage format: {"content": "...", "type": "tool", ...}
             if "content" in obs_output:
                 content = obs_output["content"]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/strands_evals/providers/langfuse_provider.py [hunk 1]
```diff
@@ -518,6 +556,28 @@ def _parse_tool_result(self, obs_output: Any) -> tuple[str, str | None]:
         content = str(obs_output) if obs_output is not None else ""
         return content, None
 
+    def _parse_tool_result_message(self, message: Any) -> str:
```
opieter-aws, line 561: Is there an option to consolidate with `_extract_text_from_content` so we maintain main logic only once?

opieter-aws (dismissed):

src/strands_evals/providers/langfuse_provider.py
poshinchen: good callout, i'll check it as follow-up!

opieter-aws (approved):

src/strands_evals/providers/langfuse_provider.py
```diff
@@ -461,6 +466,9 @@ def _convert_tool_execution(self, obs: Any, session_id: str) -> ToolExecutionSpa
             tool_name = obs.name or ""
             if isinstance(obs_input, dict):
                 tool_arguments = obs_input
+            elif isinstance(obs_input, list):
```
jjbuck, line 469: Could we avoid using `isinstance(obs_input, list)` as the Strands OTEL discriminator? This converter also handles “LangChain / universal” TOOL observations, where list-valued arguments are possible. For example, `[{"content": "a"}, {"content": "b"}]` now becomes `{"input": "a"}`, silently dropping the second item.

The nested `isinstance()` checks in `_parse_tool_arguments_from_list` are a symptom of two responsibilities being mixed: recognizing and unwrapping a producer-specific envelope, and normalizing a logical argument value. Could a small Strands envelope decoder first validate `role == "tool"` and extract content/id, then pass content to one shared argument normalizer? Unmatched lists could retain the generic fallback behavior. This keeps producer-format complexity behind a narrower interface and avoids duplicating the dict/JSON/scalar decision tree.

src/strands_evals/providers/langfuse_provider.py
```diff
@@ -461,6 +466,9 @@ def _convert_tool_execution(self, obs: Any, session_id: str) -> ToolExecutionSpa
...
+                # Strands OTEL format: [{"role": "tool", "content": "<JSON args>", "id": "..."}]
+                tool_arguments = self._parse_tool_arguments_from_list(obs_input)
```
jjbuck, line 471: The Strands OTEL format documented above carries an id, and the new fixture includes "id": "tooluse_x", but this helper returns only the arguments and the TOOL branch subsequently sets `tool_call_id = None`.

Consequently, both the canonical `ToolCall` and `ToolResult` lose that identifier. The existing Strands SPAN path preserves `toolUseId`, and these canonical fields exist to correlate a call with its result.

Let's have  the envelope parser return both the arguments and identifier, using the output id as a fallback or consistency check, and assert `tool_call_id == "tooluse_x"` on both objects.

src/strands_evals/providers/langfuse_provider.py
```diff
@@ -507,6 +542,11 @@ def _parse_tool_result(self, obs_output: Any) -> tuple[str, str | None]:
                 status = obs_output.get("status", "")
                 error = None if status == "success" else (str(status) if status else None)
                 return content, error
+            # Strands OTEL format: {"message": "<JSON string of [{\"text\": ...}]>", "id": "..."}
+            # This payload carries no error/status field; failed tool calls surface a
+            # "result"/"status" dict instead, handled by the branch above.
+            if "message" in obs_output:
```
jjbuck, line 548: This branch is applied to every TOOL output dictionary, not only the Strands OTEL format. Because message is also a common domain field, a generic result such as `{"message": "transfer completed", "transaction_id": "tx-123"}` is now reduced to "transfer completed", silently discarding the transaction ID.

Before this branch, the complete dictionary reached the fallback.

Let's have `_parse_tool_result` be told which producer dialect was recognized from the input envelope, rather than inferring Strands from the presence of a common key. A regression test for a generic result containing message would protect the existing universal path.

src/strands_evals/providers/langfuse_provider.py
```diff
@@ -518,6 +558,20 @@ def _parse_tool_result(self, obs_output: Any) -> tuple[str, str | None]:
...
+        Input format:
+            str (JSON): `'[{"text": "result text"}]'`  → `"result text"`
+            str:        `"plain text"`                  → `"plain text"`
+            list[dict]: `[{"text": "result text"}]`     → `"result text"`
+            other:                                       → `str(message)`
+
+        Falls back to the raw string when the JSON does not decode to a
+        non-empty list of text blocks.
+        """
+        return self._extract_text_from_content(message) or str(message)
```
jjbuck, line 573: `_extract_text_from_content` is an agent-message helper whose `_first_text_from_list` implementation intentionally returns only the first text block. Reusing it for tool results means `[{"text": "first"}, {"text": "second"}]` becomes only "first"; empty lists and None become the literal strings "[]" and "None".

 Other Strands mappers use `join_tool_result_content`, which preserves all blocks and maps empty/None content to "". Since `ToolResult.content` is passed to downstream evaluator prompts as evidence, dropping blocks can change faithfulness and correctness judgments. Let's reuse that normalization behavior and add multi-block and empty-result cases.

tests/strands_evals/providers/test_langfuse_provider.py
```diff
@@ -648,6 +648,137 @@ def test_tool_type_with_string_input(self, provider, mock_client):
...
+                    ],
+                    obs_output={"message": '[{"text": "Data stored under key bsm-get-case (9 fields)"}]'},
+                ),
+                _obs("o-agent", "t1", "SPAN", name="invoke_agent a", obs_input=[{"text": "q"}], obs_output="a"),
+            ],
+        )
+        tools = [s for s in spans if isinstance(s, ToolExecutionSpan)]
+        assert tools[0].tool_call.name == "bsm-get-case"
+        assert tools[0].tool_call.arguments == {"caseId": "abc123", "marketplaceId": "ATVPDKIKX0DER"}
+        assert tools[0].tool_result.content == "Data stored under key bsm-get-case (9 fields)"
```
jjbuck, line 677: This expected value is a storage acknowledgement, not the data returned by bsm-get-case. Faithfulness evaluators consume `ToolResult.content` as the evidence supporting the agent response, so "Data stored under key ..." still does not expose the case fields needed to verify any claims about that case.

src/strands_evals/providers/langfuse_provider.py
```diff
@@ -486,16 +494,43 @@ def _convert_tool_execution(self, obs: Any, session_id: str) -> ToolExecutionSpa
...
+        Input format:
+            `[{"role": "tool", "content": "<JSON args>", "id": "tooluse_..."}]`
+
+        The actual arguments live in the first item's `content` field as a
+        JSON string. Returns the parsed dict, or `{"input": <content>}` when
+        the content is not JSON, or `{"input": str(obs_input)}` when no item
+        carries a `content` field.
+        """
+        for item in obs_input:
+            if isinstance(item, dict) and "content" in item:
```
jjbuck, line 509: The repeated `isinstance` checks are not inherently problematic, but this ladder is doing three jobs at once: locating content inside the Strands transport envelope, decoding JSON, and deciding how unsupported values are coerced into ToolCall.arguments. The dict/string/scalar normalization also duplicates logic already present in `_convert_tool_execution`, so the two paths can acquire different  fallback semantics over time.


Let's separate this into an envelope extractor that validates the Strands shape and returns content plus id, followed by one shared `_normalize_tool_arguments`(value) function. That would give the module a single, explicit normalization contract and keep producer-specific representation complexity hidden from the rest of the converter.

PR #320, opened 2026-07-22
fix: detect_otel_mapper checks all spans for body in CloudWatch split format

stone-coding:
## Description

  In CloudWatch split format, each span produces two separate log entries:
  1. Span metadata (has `scope.name` but no `body`)
  2. Log event (has `body` with input/output messages)

  `detect_otel_mapper` checks `get_body(span)` on the first Strands-scoped span only. Since that's always the
  metadata entry (no body), it incorrectly returns `StrandsInMemorySessionMapper` instead of
  `CloudWatchSessionMapper`, which then crashes with `AttributeError: 'dict' object has no attribute
  'attributes'`.

  **Example:**
  Entry 1: {"spanId": "s1", "scope": {"name": "strands.telemetry.tracer"}, "attributes": {...}}  ← NO body
  Entry 2: {"spanId": "s1", "body": {"input": {...}, "output": {...}}}                           ← HAS body

  `detect_otel_mapper` hits Entry 1 first → scope is Strands → `get_body(Entry 1)` = None → returns
  `StrandsInMemorySessionMapper` → CRASH: `'dict' has no attribute 'attributes'`
  It never checks Entry 2 (which has the body).

  **Fix:** When Strands scope is detected, `break` out of the scope-detection loop and fall through to the
  existing body-scan fallback (lines 119-124) which already checks all spans for body. This avoids a nested scan
  while correctly routing to `CloudWatchSessionMapper`.

  ## Related Issues

  N/A

  ## Documentation PR

  N/A

  ## Type of Change

  Bug fix

  ## Testing

  - Added regression test: `test_detects_strands_cloudwatch_split_format` — verifies split-format spans (scope
  entry without body + separate body entry) correctly route to `CloudWatchSessionMapper`
  - All 42 existing tests passing
  - Verified locally: CloudWatch split-format spans now correctly route to `CloudWatchSessionMapper`
  - Workaround currently in bedrock-agentcore-sdk-python PR #568 (can be removed once this fix ships)

  ## Checklist
  can explain why it works
  - [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are
  needed

  ----
  
  By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution,
  under the terms of your choice.

### DIFF
--- a/src/strands_evals/mappers/utils.py
+++ b/src/strands_evals/mappers/utils.py
@@ -109,7 +109,9 @@ def detect_otel_mapper(spans: list[Any]) -> SessionMapper:
 
         if scope_name == SCOPE_STRANDS:
             # Auto-detect format for Strands
-            if get_body(span) is not None:
+            # Check ALL spans for body — CloudWatch split format has body
+            # on a separate entry from the scope entry
+            if any(get_body(s) is not None for s in spans):
                 return CloudWatchSessionMapper()
             else:
                 return StrandsInMemorySessionMapper()
--- a/src/strands_evals/mappers/utils.py
+++ b/src/strands_evals/mappers/utils.py
@@ -108,13 +108,10 @@ def detect_otel_mapper(spans: list[Any]) -> SessionMapper:
             return OpenInferenceSessionMapper()
 
         if scope_name == SCOPE_STRANDS:
-            # Auto-detect format for Strands
-            # Check ALL spans for body — CloudWatch split format has body
-            # on a separate entry from the scope entry
-            if any(get_body(s) is not None for s in spans):
-                return CloudWatchSessionMapper()
-            else:
-                return StrandsInMemorySessionMapper()
+            # CloudWatch split format puts body on a separate entry from
+            # the scoped metadata entry. Break here and let the fallback
+            # body-scan below determine CloudWatch vs InMemory.
+            break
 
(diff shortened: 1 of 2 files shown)


src/strands_evals/mappers/utils.py
```diff
@@ -109,7 +109,9 @@ def detect_otel_mapper(spans: list[Any]) -> SessionMapper:
 
         if scope_name == SCOPE_STRANDS:
             # Auto-detect format for Strands
-            if get_body(span) is not None:
+            # Check ALL spans for body — CloudWatch split format has body
+            # on a separate entry from the scope entry
+            if any(get_body(s) is not None for s in spans):
                 return CloudWatchSessionMapper()
             else:
                 return StrandsInMemorySessionMapper()
```
poshinchen: looping the spans inside a span is weird. What you need is to `break` here.

```
if scope_name == SCOPE_STRANDS:
# Strands detected; CloudWatch-vs-InMemory is decided by scanning
# all entries below, since CloudWatch split format puts the body
# on a separate log entry from the scoped 
    break
```
Then it'll fallback to the `get_body` validation below and still works.

poshinchen (changes requested):
left the comment. Please add the tests too.

poshinchen (approved):

PR #209, opened 2026-04-30
feat(simulation): add optional tools parameter to ToolSimulator (#208)

kaghatim:
## Description

Add an optional `tools` parameter to `ToolSimulator.__init__` that gets passed through to the internal simulation `Agent`. This lets users provide helper tools (like `calculator` from `strands-agents-tools`) to improve the accuracy of simulated responses, especially for stateful, arithmetic-heavy simulations.

Currently the simulation agent is created with `tools=[]` hardcoded. This works fine for simple simulations, but when simulating APIs that require arithmetic consistency across multiple tool calls (e.g. inventory reconciliation, order metrics, settlement calculations), the LLM frequently gets the math wrong. Those errors compound because the `StateRegistry` feeds previous responses back into subsequent prompts.

The change is minimal and fully backward compatible. `tools` defaults to `None` (stored as `[]`), so existing behavior is unchanged.

### What changed

- `ToolSimulator.__init__` accepts an optional `tools` parameter
- `_simulate_tool_call` passes `self.tools` to the `Agent` constructor instead of `[]`
- 5 new unit tests covering default behavior, storage, and pass-through to Agent

### Example usage

```python
from strands_tools import calculator
from strands_evals.simulation import ToolSimulator

simulator = ToolSimulator(
    model="us.anthropic.claude-sonnet-4-20250514",
    tools=[calculator],
)
```

## Related Issues

Closes #208

## Documentation PR

No documentation changes needed. The new parameter follows existing patterns and is optional with a sensible default.

## Type of Change

New feature

## Testing

- [x] I ran `hatch run prepare`
- All 982 existing tests continue to pass
- 5 new unit tests added covering:
  - Default tools is `[]`
  - Tools parameter is stored correctly
  - `None` becomes `[]`
  - Tools are passed through to the internal Agent
  - Default simulator passes empty list to Agent

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #208, opened 2026-04-30
[FEATURE] Optional tools parameter on ToolSimulator for arithmetic-heavy simulations


kaghatim:
### Problem Statement

The internal simulation agent in `ToolSimulator._simulate_tool_call` is created with `tools=[]`, which means the LLM has to do all reasoning and arithmetic in a single forward pass. This works fine for simple simulations, but falls apart when you're simulating stateful, math-heavy tool APIs.

For example, when simulating inventory reconciliation across multiple tool calls (`getInventory` -> `createOrder` -> `getInventory`), the LLM needs to correctly compute derived values like remaining stock, order totals, and defect rates. With `tools=[]`, the LLM frequently gets the arithmetic wrong, and those errors compound across subsequent calls since the `StateRegistry` feeds previous responses back into the prompt.

In my testing, a scenario like "beginning inventory 100, shipped 30, returns 5, reimbursements 1" produced quantity: 74 instead of the correct 76. There's no way to fix this without subclassing and overriding a private method.

### Proposed Solution

Add an optional `tools` parameter to `ToolSimulator.__init__` that gets passed through to the internal `Agent` constructor. Defaults to `[]` to preserve current behavior.

```python
 # Before
class ToolSimulator:
    def __init__(self, state_registry=None, model=None, max_tool_call_cache_size=20):
        ...
```

```python
  # After
class ToolSimulator:
     def __init__(self, state_registry=None, model=None, max_tool_call_cache_size=20, tools=None):
          ...
          self.tools = tools or []
```

And in `_simulate_tool_call`:

```python
def _simulate_tool_call(self, prompt, structured_output_model=None):
    agent = Agent(
        tools=self.tools,  # was hardcoded to []
        model=self.model,
        callback_handler=None,
    )
    return agent(prompt, structured_output_model=structured_output_model)
```

### Use Case

I'm building an evaluation suite for agents that interact with APIs that require maintaining arithmetic consistency across calls:

- Inventory reconciliation: beginning stock, shipments, returns, and reimbursements must add up correctly
- Order metrics: total orders, defect counts, and defect rates (e.g. 4 defects / 200 orders = 0.02) need to be precise
- Settlement reports: fees, refunds, and net proceeds across multiple line items

Passing `calculator` from `strands-agents-tools` to the simulator agent fixes the arithmetic drift. For example:

```python
from strands_tools import calculator

simulator = ToolSimulator(
    model="us.anthropic.claude-sonnet-4-20250514",
[...]

PR #324, opened 2026-07-23
feat(evaluators): allow custom tools on judge-based evaluators (Trajectory, Output, Multimodal)

### LINKED ISSUE #322: [FEATURE] Allow custom tools on TrajectoryEvaluator and OutputEvaluator
## Problem Statement

`TrajectoryEvaluator` hardcodes its judge agent's tools to the three built-in list scorers:

```python
self._tools = [exact_match_scorer, in_order_match_scorer, any_order_match_scorer]
```

`OutputEvaluator` gives its judge agent no tools at all. Neither accepts a `tools` parameter, so the judge cannot:

- Call domain-specific validation functions (e.g., check an ID format, validate a state transition)
- Verify claims against reference data the rubric alone can't encode
- Cross-reference facts between steps of a trajectory

Today the only way to get a tool-augmented judge is to subclass the evaluator and re-implement `evaluate`/`evaluate_async` wholesale, duplicating the prompt-composition and structured-output logic.

## Proposed Solution

Add an optional `tools` constructor parameter, following the exact pattern already accepted for `ToolSimulator` in #208 / #209:

```python
from strands import tool

@tool
def verify_ticket_id(ticket_id: str) -> str:
    """Check whether a ticket ID matches the support system's format."""
    return "valid" if ticket_id.startswith("TKT-") else "invalid format"

evaluator = TrajectoryEvaluator(
    rubric="Verify the agent referenced only valid ticket IDs and passed data between steps.",
    tools=[verify_ticket_id],
)
```

Semantics:

- **TrajectoryEvaluator**: user tools are merged ahead of the default scoring tools; the built-in scorers are always retained.
- **OutputEvaluator**: `tools=None` (default) keeps the current no-tools judge; when provided, tools are passed to the judge agent in both sync and async paths. `MultimodalOutputEvaluator` subclasses inherit the behavior.
- Non-breaking: omitting `tools` keeps existing behavior exactly.

## Use Case

We evaluate multi-step support-ticket agents where the judge needs to check that referenced entities (ticket IDs, transaction references) are structurally valid and consistent across steps. A rubric can describe this, but a plain judge has no way to *verify* it; a small verification tool makes the judgment grounded instead of impressionistic. In our runs, a finder-then-verify judge with tools was consistently more reliable than a single-pass holistic judge on the same rubric.

## Alternatives Considered

- **Subclassing per domain**: works but duplicates prompt-composition/structured-output logic and drifts from upstream as those evolve.
- **Encoding validation into the rubric text**: the judge asserts rather than verifies; fabricated-but-plausible values pass.

## Additional Context

- Same opt-in pattern as #208 (accepted and merged in #209 for `ToolSimulator`).
- I have a working implementation with unit tests ready to submit as a PR if this direction looks right.

pdebjyot:
Closes #322

## Description

Adds an optional `tools=` constructor parameter so the judge agent can call domain-specific verification functions during evaluation. Follows the same opt-in pattern accepted for `ToolSimulator` in #208 / #209.

Covered evaluators: `TrajectoryEvaluator`, `OutputEvaluator`, `MultimodalOutputEvaluator`, and its specialized subclasses (`MultimodalCorrectnessEvaluator`, `MultimodalFaithfulnessEvaluator`, `MultimodalInstructionFollowingEvaluator`, `MultimodalOverallQualityEvaluator`).

- **TrajectoryEvaluator**: user tools are appended after the default trajectory scoring tools (exact/in-order/any-order match), which are always retained.
- **OutputEvaluator**: `tools` defaults to `None`, preserving the current no-tools judge (`Agent` accepts `tools=None`).
- **Multimodal evaluators**: `MultimodalOutputEvaluator` accepts `tools` and forwards it to `OutputEvaluator`; the four specialized subclasses accept and forward it likewise, so tool-augmented judging works for multimodal rubrics too.

Non-breaking: `tools` is appended after `name` in every signature, so existing positional and keyword callers are unaffected; omitting it keeps current behavior exactly.

## Example

```python
from strands import tool
from strands_evals.evaluators import TrajectoryEvaluator

@tool
def verify_ticket_id(ticket_id: str) -> str:
    """Check whether a ticket ID matches the support system's format."""
    return "valid" if ticket_id.startswith("TKT-") else "invalid format"

evaluator = TrajectoryEvaluator(
    rubric="Verify the agent referenced only valid ticket IDs.",
    tools=[verify_ticket_id],
)
```

## Testing

- New unit tests: constructor defaults preserved, merged tools list asserted explicitly (trajectory), tools passed through to the judge agent (output, sync + async), and `tools=None` default verified.
- Full evaluator + experiment unit suites pass; `ruff check` and `ruff format` clean.

## Checklist

- [x] Non-breaking change
- [x] Unit tests added
- [x] Conventional commit
- [x] Lint/format clean

### DIFF
--- a/src/strands_evals/evaluators/output_evaluator.py
+++ b/src/strands_evals/evaluators/output_evaluator.py
@@ -31,8 +31,8 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
         system_prompt: str = SYSTEM_PROMPT,
         include_inputs: bool = True,
         uses_environment_state: bool = False,
-        tools: list[Any] | None = None,
         name: str | None = None,
+        tools: list[Any] | None = None,
     ):
         super().__init__(name=name)
         self.rubric = rubric
@@ -40,7 +40,7 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
         self.include_inputs = include_inputs
         self.system_prompt = system_prompt
         self.uses_environment_state = uses_environment_state
-        self._tools = tools
+        self.tools = tools
 
     def _build_prompt(self, evaluation_case: EvaluationData[InputT, OutputT]) -> str | list:
         """Build the evaluation prompt for a test case.
@@ -70,17 +70,12 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
         Returns:
             The results of the evaluation as EvaluationOutput.
         """
-        evaluator_agent = self._create_evaluator_agent()
+        evaluator_agent = Agent(
+            model=self.model, tools=self.tools, system_prompt=self.system_prompt, callback_handler=None
+        )
         evaluation_prompt = self._build_prompt(evaluation_case)
         result = evaluator_agent(evaluation_prompt, structured_output_model=EvaluationOutput)
         return [cast(EvaluationOutput, result.structured_output)]
-
-    def _create_evaluator_agent(self) -> Agent:
-        """Create the judge agent, including custom tools when provided."""
-        kwargs: dict[str, Any] = {"model": self.model, "system_prompt": self.system_prompt, "callback_handler": None}
-        if self._tools:
-            kwargs["tools"] = self._tools
-        return Agent(**kwargs)
 
     async def evaluate_async(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
         """
@@ -92,7 +87,9 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
         Returns:
             The results of the evaluation as EvaluationOutput.
         """
-        evaluator_agent = self._create_evaluator_agent()
+        evaluator_agent = Agent(
+            model=self.model, tools=self.tools, system_prompt=self.system_prompt, callback_handler=None
+        )
         evaluation_prompt = self._build_prompt(evaluation_case)
         result = await evaluator_agent.invoke_async(evaluation_prompt, structured_output_model=EvaluationOutput)
         return [cast(EvaluationOutput, result.structured_output)]
--- a/src/strands_evals/evaluators/output_evaluator.py
+++ b/src/strands_evals/evaluators/output_evaluator.py
@@ -1,4 +1,4 @@
-from typing import cast
+from typing import Any, cast
 
 from strands import Agent
 from strands.models.model import Model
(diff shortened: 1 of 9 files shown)


src/strands_evals/evaluators/output_evaluator.py
```diff
@@ -66,11 +70,18 @@ def evaluate(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[Eva
         Returns:
             The results of the evaluation as EvaluationOutput.
         """
-        evaluator_agent = Agent(model=self.model, system_prompt=self.system_prompt, callback_handler=None)
+        evaluator_agent = self._create_evaluator_agent()
         evaluation_prompt = self._build_prompt(evaluation_case)
         result = evaluator_agent(evaluation_prompt, structured_output_model=EvaluationOutput)
         return [cast(EvaluationOutput, result.structured_output)]
 
+    def _create_evaluator_agent(self) -> Agent:
```
poshinchen: I think this is overkill, and [Agent.tools](https://github.com/strands-agents/harness-sdk/blob/main/strands-py/src/strands/agent/agent.py#L165) can be None. 

```Agent(model=self.model, tools=self.tools, system_prompt=self.system_prompt, callback_handler=None)```

^^ is sufficient already.

File: src/strands_evals/evaluators/output_evaluator.py [hunk 2]
```diff
@@ -37,6 +40,7 @@ def __init__(
         self.include_inputs = include_inputs
         self.system_prompt = system_prompt
         self.uses_environment_state = uses_environment_state
+        self._tools = tools
```
poshinchen: `self.tools` is enough, I think it doesn't need to be `_`

poshinchen:
In the description:

> via a new _create_evaluator_agent() helper (inherited by MultimodalOutputEvaluator subclasses).

This is incorrect.

src/strands_evals/evaluators/trajectory_evaluator.py
```diff
@@ -32,14 +34,15 @@ def __init__(
...
+        tools: list[Any] | None = None,
         name: str | None = None,
     ):
         super().__init__(name=name)
         self.rubric = rubric
         self.trajectory_description = trajectory_description
         self.model = model
         self.include_inputs = include_inputs
-        self._tools: list[str | dict[str, str] | Any] | None = [
+        self._tools: list[str | dict[str, str] | Any] | None = list(tools or []) + [
```
poshinchen:
```
self._tools: list[str | dict[str, str] | Any] = [
      exact_match_scorer,
      in_order_match_scorer,
      any_order_match_scorer,
      *(tools or []),
  ]
``` could be better?

poshinchen:
I see that you've updated other evaluators too, could you update the description / title?

pdebjyot (commented):
Updated the title and description to cover all touched evaluators — the multimodal family (MultimodalOutputEvaluator and its four specialized subclasses) now accepts and forwards tools as well, which their to_dict/from_dict round-trip tests require once the attribute is public. Thanks for the catch.

poshinchen (approved):

Issue #88, opened 2026-01-22
[FEATURE] Experiment Analysis and Comparison


poshinchen:
## Problem Statement

Developers / Scientist need to compare evaluation results across different experiments (e.g., prompt variations, model switches, parameter tuning) to make data-driven decisions. Currently, there's no built-in mechanism to:

* Track experiment metadata and configurations
* Compare results across multiple evaluation runs
* Visualize performance differences
* Identify statistically significant improvements

### Proposed Solution

1. Mechanisms to execute evaluations on the agents in different models
2. Mechanisms group / compare experiments and provide meaning reasoning suggestions based on the experiment results

### Use Case

* As a developer tuning prompts, I want to compare evaluation results across 5 different prompt variations to identify the best performing version
* I want to track evaluation metrics over time to ensure model updates don't cause regressions

PR #325, opened 2026-07-23
feat(experiment): add opt-in model routing for evaluators

### LINKED ISSUE #323: [FEATURE] Model routing for evaluators — match model tier to evaluation complexity
## Problem Statement

Every evaluator takes a single `model` parameter, and there is no shared routing layer. In a multi-evaluator experiment this forces a choice:

1. Use one strong model everywhere — accurate but wasteful for structural checks (tool selection, parameter accuracy) that a fast model judges just as well;
2. Hand-wire a different model into each evaluator — verbose, no shared config, and impossible to make **case-dependent** (e.g., "use the strong model only when the trace is long").

For a typical 70-case run with 5 LLM evaluators, that's 350 strong-model judge calls where roughly half could be a fast-tier model with no measurable quality loss — or, inverted, hard cases stuck on a mid-tier model because upgrading all evaluators is too expensive.

## Proposed Solution

An opt-in `ModelRouter` consulted by `Experiment`, with ordered rules matched on evaluator type and an optional per-case condition:

```python
from strands_evals import Experiment, ModelRouter, RoutingRule

router = ModelRouter(rules=[
    # Structural checks -> fast tier
    RoutingRule(
        evaluator_types=[ToolSelectionAccuracyEvaluator, ToolParameterAccuracyEvaluator],
        model="us.anthropic.claude-haiku-4-5-20251001-v1:0",
    ),
    # Complex judgment on long traces -> strongest tier
    RoutingRule(
        evaluator_types=[GoalSuccessRateEvaluator],
        model="us.anthropic.claude-opus-4-1-20250805-v1:0",
        condition=lambda case: len(str(case.actual_trajectory or "")) > 10_000,
    ),
])

experiment = Experiment(cases=cases, evaluators=evaluators, model_router=router)
```

Semantics:

- **Opt-in**: no `model_router` means behavior is unchanged.
- **Explicit wins**: an evaluator constructed with `model=...` is never re-routed — the router only applies to evaluators using the default model resolution.
- **First matching rule wins**; no match falls through to an optional `default_model`, else the evaluator's own default.
- Routing returns a shallow copy carrying the routed model, so shared evaluator instances are never mutated across concurrent workers; evaluators without a `model` attribute (deterministic ones) pass through untouched.
- A raising condition logs a warning and skips the rule rather than failing the evaluation.

## Use Case

We run this two-tier pattern (strong model for faithfulness/accuracy/hallucination-style metrics, fast model for relevance/utilization/efficiency-style rubric checks) in a production evaluation pipeline over support-agent traces. It cut judge cost roughly in half at equal accuracy on the simple metrics, and freed budget to upgrade the genuinely hard metrics. The `condition` hook is what makes it a router rather than a per-evaluator setting — long/complex traces can escalate to a stronger judge while short ones stay cheap.

Related: the model-comparison direction in #88 — routing is complementary (that epic compares models across experiments; this routes models within one).

## Alternatives Considered

- **Per-evaluator `model=` only** (status quo): no case-dependent routing, config scattered across constructors.
- **Env-var global model**: single tier only; can't mix tiers within one experiment.

## Additional Context

I have a working implementation (`ModelRouter`, `RoutingRule`, `Experiment(model_router=...)`) with unit and integration tests ready to submit as a PR if this direction looks right. `model_router` is intentionally excluded from `Experiment.to_dict()/from_dict()` since routing conditions are arbitrary callables.

pdebjyot:
Closes #323
Related: #88

## Description

Adds `ModelRouter` and `RoutingRule` (new module `strands_evals/model_router.py`), plus an optional `model_router=` parameter on `Experiment`, so evaluation cost/accuracy can be matched to evaluator complexity: fast models for structural/rubric checks, strong models for nuanced judgment — including **case-dependent** escalation (e.g., long traces get a stronger judge).

Semantics:

- **Opt-in**: no `model_router` → behavior unchanged.
- **Explicit wins**: an evaluator constructed with `model=...` is never re-routed.
- **First matching rule wins**; rules match on evaluator type (class or name) plus an optional per-case `condition` predicate. No match falls through to an optional `default_model`.
- Routing returns a shallow copy of the evaluator carrying the routed model, so shared instances are never mutated across concurrent workers. Evaluators without a `model` attribute (deterministic) pass through untouched.
- A raising `condition` logs a warning and skips the rule instead of failing the evaluation.
- `model_router` is intentionally excluded from `Experiment.to_dict()/from_dict()` since routing conditions are arbitrary callables.

## Example

```python
from strands_evals import Experiment, ModelRouter, RoutingRule

router = ModelRouter(rules=[
    RoutingRule(
        evaluator_types=[ToolSelectionAccuracyEvaluator, ToolParameterAccuracyEvaluator],
        model="us.anthropic.claude-haiku-4-5-20251001-v1:0",
    ),
    RoutingRule(
        evaluator_types=[GoalSuccessRateEvaluator],
        model="us.anthropic.claude-opus-4-1-20250805-v1:0",
        condition=lambda case: len(str(case.actual_trajectory or "")) > 10_000,
    ),
])

experiment = Experiment(cases=cases, evaluators=evaluators, model_router=router)
```

## Testing

- 12 new tests: rule matching by class/name, condition gating, condition-exception safety, first-match-wins, default-model fallback, explicit-model preservation, no-mutation of shared instances, deterministic-evaluator passthrough, and two end-to-end `Experiment` integration tests asserting the judge agent is constructed with the routed model (and unchanged without a router).
- Full evaluator + experiment unit suites pass (454 tests); `ruff check` and `ruff format` clean.

## Checklist

- [x] Non-breaking change (opt-in)
- [x] Unit + integration tests added
- [x] Conventional commit
- [x] Lint/format clean

### DIFF
--- a/src/strands_evals/__init__.py
+++ b/src/strands_evals/__init__.py
@@ -4,6 +4,7 @@ from .eval_task_handler import EvalTaskHandler, TracedHandler, eval_task
 from .evaluation_data_store import EvaluationDataStore
 from .experiment import Experiment
 from .local_file_task_result_store import LocalFileTaskResultStore
+from .model_router import ModelRouter, RoutingRule
 from .simulation import ActorSimulator, UserSimulator
 from .telemetry import StrandsEvalsTelemetry, get_tracer
 from .types.detector import DiagnosisConfig
@@ -13,6 +14,8 @@ __all__ = [
     "DiagnosisConfig",
     "Experiment",
     "Case",
+    "ModelRouter",
+    "RoutingRule",
     "LocalFileTaskResultStore",
     "EvaluationDataStore",
     "EvaluationReport",
--- a/src/strands_evals/experiment.py
+++ b/src/strands_evals/experiment.py
@@ -42,6 +42,7 @@ from .evaluators.stereotyping_evaluator import StereotypingEvaluator
 from .evaluators.tool_parameter_accuracy_evaluator import ToolParameterAccuracyEvaluator
 from .evaluators.tool_selection_accuracy_evaluator import ToolSelectionAccuracyEvaluator
 from .evaluators.trajectory_evaluator import TrajectoryEvaluator
+from .model_router import ModelRouter
 from .telemetry import get_tracer, serialize
 from .telemetry._cloudwatch_logger import _send_to_cloudwatch
 from .types.detector import DiagnosisConfig
@@ -93,6 +94,8 @@ class Experiment(Generic[InputT, OutputT]):
     Attributes:
         cases: A list of test cases in the experiment.
         evaluators: The list of evaluators to be used on the test cases.
+        model_router: Optional ModelRouter that selects a model per (evaluator, case)
+            pair. Only applies to evaluators constructed without an explicit model.
 
     Example:
         experiment = Experiment[str, str](
@@ -123,6 +126,7 @@ class Experiment(Generic[InputT, OutputT]):
         cases: list[Case[InputT, OutputT]] | None = None,
         evaluators: list[Evaluator[InputT, OutputT]] | None = None,
         diagnosis_config: DiagnosisConfig | None = None,
+        model_router: ModelRouter | None = None,
     ):
         self._cases = cases or []
         self._evaluators = evaluators or [Evaluator()]
@@ -130,6 +134,7 @@ class Experiment(Generic[InputT, OutputT]):
 
         self._config_id = os.environ.get("EVALUATION_RESULTS_LOG_GROUP", "default-strands-evals")
         self._diagnosis_config = diagnosis_config
+        self._model_router = model_router
 
     @property
     def cases(self) -> list[Case[InputT, OutputT]]:
@@ -338,6 +343,8 @@ class Experiment(Generic[InputT, OutputT]):
         Returns:
             A dict with evaluator_name, test_pass, score, reason, and detailed_results.
         """
+        if self._model_router is not None:
+            evaluator = self._model_router.route(evaluator, evaluation_context)
 
         @retry(
             retry=retry_if_exception(is_throttling_error),
--- a/src/strands_evals/model_router.py
+++ b/src/strands_evals/model_router.py
@@ -1,0 +1,151 @@
+"""Model routing for evaluators.
+
+Lets an Experiment match model tier to evaluation complexity: cheap/fast models
+for structural or rubric-style checks, stronger models for nuanced judgment.
+
+Routing is opt-in and advisory:
+- An evaluator constructed with an explicit ``model=`` always keeps that model.
+- Evaluators whose ``model`` is None (the default) are routed through the
+  router's rules; the first matching rule wins.
+- If no rule matches, the router's ``default_model`` (if any) is used;
+  otherwise the evaluator keeps its default model resolution.
+
+Example::
+
+    from strands_evals import Experiment, ModelRouter, RoutingRule
+
+    router = ModelRouter(rules=[
+        # Structural checks -> fast model
+        RoutingRule(
+            evaluator_types=["ToolSelectionAccuracyEvaluator", "ToolParameterAccuracyEvaluator"],
+            model="us.anthropic.claude-haiku-4-5-20251001-v1:0",
+        ),
+        # Complex judgment on long traces -> strongest model
+        RoutingRule(
+            evaluator_types=["GoalSuccessRateEvaluator"],
+            model="us.anthropic.claude-opus-4-1-20250805-v1:0",
+            condition=lambda case: len(str(case.actual_trajectory or "")) > 10_000,
+        ),
+    ])
+
+    experiment = Experiment(cases=cases, evaluators=evaluators, model_router=router)
+"""
+
+import copy
+import logging
+from typing import TYPE_CHECKING, Callable
+
+from strands.models.model import Model
+
+if TYPE_CHECKING:
+    from .evaluators.evaluator import Evaluator
+    from .types.evaluation import EvaluationData
+
+logger = logging.getLogger(__name__)
+
+
+class RoutingRule:
+    """A rule mapping evaluator types (and optionally case properties) to a model.
+
+    Attributes:
+        evaluator_types: Evaluator classes or class names this rule applies to.
+        model: The model (Model instance or Bedrock model-id string) to use when
+            this rule matches.
+        condition: Optional predicate over the evaluation data. When provided,
+            the rule only matches cases for which it returns True. This enables
+            complexity-based routing (e.g., trace length, span count).
+    """
+
+    def __init__(
+        self,
+        evaluator_types: list[type | str],
+        model: Model | str,
+        condition: Callable[["EvaluationData"], bool] | None = None,
+    ):
+        self._type_names = {t if isinstance(t, str) else t.__name__ for t in evaluator_types}
+        self.model = model
+        self.condition = condition
+
+    def matches(self, evaluator: "Evaluator", evaluation_data: "EvaluationData") -> bool:
+        """Check whether this rule applies to the given evaluator and case.
+
+        Args:
+            evaluator: The evaluator about to run.
+            evaluation_data: The evaluation context for the current case.
+
+        Returns:
+            True if the evaluator's type is covered by this rule and the
+            condition (if any) holds for the case.
+        """
+        if evaluator.get_type_name() not in self._type_names:
+            return False
+        if self.condition is not None:
+            try:
+                return bool(self.condition(evaluation_data))
+            except Exception as e:
+                logger.warning(
+                    "rule_types=<%s>, error=<%s> | routing condition raised, skipping rule", self._type_names, e
+                )
+                return False
+        return True
+
+
+class ModelRouter:
+    """Selects a model for each (evaluator, case) pair using an ordered rule list.
+
+    Attributes:
+        rules: Ordered list of RoutingRule; the first matching rule wins.
+        default_model: Model used when no rule matches. None means "leave the
+            evaluator's own model resolution untouched".
+    """
+
+    def __init__(self, rules: list[RoutingRule] | None = None, default_model: Model | str | None = None):
+        self.rules = rules or []
+        self.default_model = default_model
+
+    def select_model(self, evaluator: "Evaluator", evaluation_data: "EvaluationData") -> Model | str | None:
+        """Select a model for an evaluator run, or None to keep the evaluator's default.
+
+        Args:
+            evaluator: The evaluator about to run.
+            evaluation_data: The evaluation context for the current case.
+
+        Returns:
+            The routed model, the router default, or None when routing does not apply.
+        """
+        for rule in self.rules:
+            if rule.matches(evaluator, evaluation_data):
+                return rule.model
+        return self.default_model
+
+    def route(self, evaluator: "Evaluator", evaluation_data: "EvaluationData") -> "Evaluator":
+        """Return the evaluator to run, applying model routing when appropriate.
+
+        An evaluator with an explicit model (its ``model`` attribute is set) is
+        returned unchanged — user configuration always wins. Evaluators without
+        a ``model`` attribute (e.g., deterministic evaluators) are also returned
+        unchanged. Otherwise a shallow copy with the routed model is returned so
+        the shared evaluator instance is never mutated across concurrent workers.
+
+        Args:
+            evaluator: The evaluator about to run.
+            evaluation_data: The evaluation context for the current case.
+
+        Returns:
+            Either the original evaluator or a shallow copy carrying the routed model.
+        """
+        if getattr(evaluator, "model", None) is not None or not hasattr(evaluator, "model"):
+            return evaluator
+
+        model = self.select_model(evaluator, evaluation_data)
+        if model is None:
+            return evaluator
+
+        routed = copy.copy(evaluator)
+        routed.model = model
+        logger.debug(
+            "evaluator=<%s>, model=<%s> | routed evaluator to model",
+            evaluator.get_name(),
+            model if isinstance(model, str) else type(model).__name__,
+        )
+        return routed
--- a/tests/strands_evals/test_model_router.py
+++ b/tests/strands_evals/test_model_router.py
@@ -1,0 +1,180 @@
+from unittest.mock import Mock, patch
+
+import pytest
+
+from strands_evals import Case, Experiment, ModelRouter, RoutingRule
+from strands_evals.evaluators import OutputEvaluator, TrajectoryEvaluator
+from strands_evals.evaluators.deterministic import Contains
+from strands_evals.types import EvaluationData, EvaluationOutput
+
+FAST_MODEL = "us.anthropic.claude-haiku-4-5-20251001-v1:0"
+STRONG_MODEL = "us.anthropic.claude-opus-4-1-20250805-v1:0"
+
+
+@pytest.fixture
+def evaluation_data():
+    return EvaluationData(input="What is 2+2?", actual_output="4", expected_output="4", name="math_test")
+
+
+class TestRoutingRule:
+    def test_matches_by_class_name_string(self, evaluation_data):
+        rule = RoutingRule(evaluator_types=["OutputEvaluator"], model=FAST_MODEL)
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        assert rule.matches(evaluator, evaluation_data) is True
+
+    def test_matches_by_class_object(self, evaluation_data):
+        rule = RoutingRule(evaluator_types=[OutputEvaluator], model=FAST_MODEL)
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        assert rule.matches(evaluator, evaluation_data) is True
+
+    def test_does_not_match_other_evaluator_types(self, evaluation_data):
+        rule = RoutingRule(evaluator_types=[TrajectoryEvaluator], model=FAST_MODEL)
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        assert rule.matches(evaluator, evaluation_data) is False
+
+    def test_condition_gates_matching(self, evaluation_data):
+        rule = RoutingRule(
+            evaluator_types=[OutputEvaluator],
+            model=STRONG_MODEL,
+            condition=lambda case: len(str(case.actual_output)) > 100,
+        )
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        assert rule.matches(evaluator, evaluation_data) is False
+
+        long_output = EvaluationData(input="q", actual_output="x" * 200)
+        assert rule.matches(evaluator, long_output) is True
+
+    def test_condition_exception_skips_rule(self, evaluation_data):
+        def broken_condition(case):
+            raise RuntimeError("boom")
+
+        rule = RoutingRule(evaluator_types=[OutputEvaluator], model=FAST_MODEL, condition=broken_condition)
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        assert rule.matches(evaluator, evaluation_data) is False
+
+
+class TestModelRouter:
+    def test_first_matching_rule_wins(self, evaluation_data):
+        router = ModelRouter(
+            rules=[
+                RoutingRule(evaluator_types=[OutputEvaluator], model=FAST_MODEL),
+                RoutingRule(evaluator_types=[OutputEvaluator], model=STRONG_MODEL),
+            ]
+        )
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        assert router.select_model(evaluator, evaluation_data) == FAST_MODEL
+
+    def test_default_model_when_no_rule_matches(self, evaluation_data):
+        router = ModelRouter(
+            rules=[RoutingRule(evaluator_types=[TrajectoryEvaluator], model=FAST_MODEL)],
+            default_model=STRONG_MODEL,
+        )
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        assert router.select_model(evaluator, evaluation_data) == STRONG_MODEL
+
+    def test_no_rule_and_no_default_returns_none(self, evaluation_data):
+        router = ModelRouter(rules=[RoutingRule(evaluator_types=[TrajectoryEvaluator], model=FAST_MODEL)])
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        assert router.select_model(evaluator, evaluation_data) is None
+
+    def test_route_returns_copy_with_routed_model(self, evaluation_data):
+        router = ModelRouter(rules=[RoutingRule(evaluator_types=[OutputEvaluator], model=FAST_MODEL)])
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        routed = router.route(evaluator, evaluation_data)
+
+        assert routed is not evaluator
+        assert routed.model == FAST_MODEL
+        # Original instance is never mutated
+        assert evaluator.model is None
+
+    def test_route_preserves_explicit_model(self, evaluation_data):
+        """An evaluator constructed with an explicit model must never be re-routed."""
+        router = ModelRouter(rules=[RoutingRule(evaluator_types=[OutputEvaluator], model=FAST_MODEL)])
+        evaluator = OutputEvaluator(rubric="Test rubric", model=STRONG_MODEL)
+
+        routed = router.route(evaluator, evaluation_data)
+
+        assert routed is evaluator
+        assert routed.model == STRONG_MODEL
+
+    def test_route_leaves_evaluator_without_model_attribute_unchanged(self, evaluation_data):
+        """Deterministic evaluators have no model attribute and must pass through."""
+        router = ModelRouter(
+            rules=[RoutingRule(evaluator_types=["Contains"], model=FAST_MODEL)], default_model=FAST_MODEL
+        )
+        evaluator = Contains(value="4")
+
+        routed = router.route(evaluator, evaluation_data)
+
+        assert routed is evaluator
+
+    def test_route_without_match_returns_original(self, evaluation_data):
+        router = ModelRouter(rules=[RoutingRule(evaluator_types=[TrajectoryEvaluator], model=FAST_MODEL)])
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        routed = router.route(evaluator, evaluation_data)
+
+        assert routed is evaluator
+        assert routed.model is None
+
+
+class TestExperimentIntegration:
+    @patch("strands_evals.evaluators.output_evaluator.Agent")
+    def test_experiment_routes_evaluator_model(self, mock_agent_class):
+        mock_agent = Mock()
+        mock_result = Mock()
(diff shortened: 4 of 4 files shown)


arielnabavian:
@strandly-the-agent review

src/strands_evals/model_router.py
```diff
@@ -0,0 +1,151 @@
...
+        the shared evaluator instance is never mutated across concurrent workers.
+
+        Args:
+            evaluator: The evaluator about to run.
+            evaluation_data: The evaluation context for the current case.
+
+        Returns:
+            Either the original evaluator or a shallow copy carrying the routed model.
+        """
+        if getattr(evaluator, "model", None) is not None or not hasattr(evaluator, "model"):
```
arielnabavian: **concern** — this guard can't tell a user's explicit choice from a serialization artifact, so routing is a silent no-op for any reloaded experiment.

`Evaluator.to_dict()` writes `model_id = DEFAULT_BEDROCK_MODEL_ID` for an evaluator built with `model=None` (`evaluators/evaluator.py:329-331`; asserted at `tests/strands_evals/test_experiment.py:371`). `Experiment.from_dict()` turns `model_id` straight back into `model=` (`experiment.py:809-810`; asserted at `tests/strands_evals/test_experiment.py:634`). So:

```python
exp = Experiment(cases=cases, evaluators=[OutputEvaluator(rubric="...")])
exp.to_file("exp.json")
# ... later, in the run that actually wants routing
exp = Experiment.from_file("exp.json")
# exp.evaluators[0].model == "global.anthropic.claude-sonnet-4-6"
```

Every evaluator now trips the `is not None` check on this line and comes back unrouted. No error, no warning, no routing — the run just costs exactly what it did before, and the only way to notice is a billing dashboard.

Related, smaller: `from_dict` (`experiment.py:821`) doesn't accept or forward `model_router`, and `Experiment` exposes no `model_router` property, so attaching a router to a reloaded experiment means rebuilding it by hand from `.cases` / `.evaluators`.

Cheapest robust fix is to record intent at construction rather than infer it from the value — e.g. a private `_model_explicit` flag set when `model` is passed, and key `route()` off that. If you'd rather not touch the evaluators in this PR, at minimum log a warning once when a router is configured and every evaluator was skipped, so the no-op is visible.

src/strands_evals/experiment.py
```diff
@@ -338,6 +343,8 @@ async def _run_evaluator(
         Returns:
             A dict with evaluator_name, test_pass, score, reason, and detailed_results.
         """
+        if self._model_router is not None:
+            evaluator = self._model_router.route(evaluator, evaluation_context)
```
arielnabavian: **concern** — the routed model isn't recorded anywhere, so a score can't be attributed to the model that produced it.

From here the routed evaluator flows into the span (`:367`, attributes at `:380`), the CloudWatch payload (`:396`) and the returned dict (`:426`) — all of which carry `evaluator_name` and `evaluator_type` and no model. Report rows are built from that dict (`:680`), so they don't have it either.

Without a router you can at least recover the judge model from `experiment.to_dict()["evaluators"][i]["model_id"]`. With a router the model varies per case by design and is recoverable from nothing in the run's output — which makes "did the fast tier score these the same as the strong tier?" unanswerable, and that's the question the feature exists to answer. It also makes a mis-scoped rule invisible: route 90% of cases to the wrong tier and the report looks identical.

Suggest `gen_ai.request.model` on the eval span plus the resolved model id in the returned dict / report row. Cheap now; awkward once report consumers exist.

src/strands_evals/model_router.py
```diff
@@ -0,0 +1,151 @@
...
+            complexity-based routing (e.g., trace length, span count).
+    """
+
+    def __init__(
+        self,
+        evaluator_types: list[type | str],
+        model: Model | str,
+        condition: Callable[["EvaluationData"], bool] | None = None,
+    ):
+        self._type_names = {t if isinstance(t, str) else t.__name__ for t in evaluator_types}
```
arielnabavian: **concern** — collapsing classes to `__name__` here means rules don't match subclasses, which is surprising given the API takes a class object.

`evaluator_types=[OutputEvaluator]` will not match `MultimodalOutputEvaluator`, which subclasses it (`evaluators/multimodal_output_evaluator.py:15`) — same for any user subclass of a built-in evaluator. Passing a class reads like `isinstance`, and the failure mode is a silent skip, so someone routing "all my output evaluators" to the fast tier will believe it worked. Two same-named classes from different modules also fold into one rule.

If exact-type matching is deliberate, say so in the `evaluator_types` docstring so the class argument doesn't imply otherwise. Otherwise keep the class objects and test them with `isinstance`, using name comparison only for the string entries. Either way it's hard to change once people depend on the current behaviour.

Minor, same construct: the class docstring lists `evaluator_types` under `Attributes` (`:50-51`), but the instance only has `_type_names` — `rule.evaluator_types` raises `AttributeError`.

src/strands_evals/model_router.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Args:
+            evaluator: The evaluator about to run.
+            evaluation_data: The evaluation context for the current case.
+
+        Returns:
+            True if the evaluator's type is covered by this rule and the
+            condition (if any) holds for the case.
+        """
+        if evaluator.get_type_name() not in self._type_names:
```
arielnabavian: **concern** — matching on type name only means two instances of the same evaluator class can't be routed differently.

`Evaluator.name` / `get_name()` exists precisely to distinguish same-class instances (`evaluators/evaluator.py:295-305`), and `Experiment._validate_evaluator_names` (`experiment.py:182-197`) *requires* distinct names when you register two. But a rule can only see `get_type_name()`, so this can't be expressed:

```python
Experiment(evaluators=[
    OutputEvaluator(rubric="did it cite a source?", name="citations"),  # cheap
    OutputEvaluator(rubric="was the advice sound?", name="soundness"),  # expensive
], model_router=router)
```

Any rule naming `OutputEvaluator` hits both, and `evaluator_types=["citations"]` silently matches nothing. #323's "70 cases, 5 LLM evaluators, roughly half cheap rubric checks" is largely this pattern — several `OutputEvaluator`s with different rubrics. Checking `get_name()` alongside `get_type_name()` is one extra condition now and a behaviour change later.

Separately: a rule that matches nothing at all is silent. A one-time warning naming rule entries that matched no evaluator during the run would surface both this and the subclass case above.

tests/strands_evals/test_model_router.py
```diff
@@ -0,0 +1,180 @@
...
+        experiment = Experiment(
+            cases=[Case(name="math", input="What is 2+2?", expected_output="4")],
+            evaluators=[evaluator],
+            model_router=router,
+        )
+
+        report = experiment.run_evaluations(lambda case: "4")
+
+        # The judge agent was constructed with the routed model
+        assert mock_agent_class.call_args[1]["model"] == FAST_MODEL
```
arielnabavian: **concern** — the capability that makes this a router has no end-to-end test.

Both integration tests use a single case and an unconditional rule, so nothing exercises per-case escalation — the one thing a router does that a per-evaluator `model=` can't. Worth a two-case experiment with a `condition` (one short trace, one long) asserting each case's judge agent got its own model.

Note `call_args` is only the *last* call, so it can't express that: a two-case test needs `call_args_list`. Asserting on `call_args` alone would pass even if case 1 routed to the wrong model.

Also untested:
- `default_model` actually reaching the judge through `Experiment` — only `select_model` is covered.
- A `Model` instance rather than a model-id string. That's the path through `routed.model = <Model>` and the `type(model).__name__` log branch (`model_router.py:144-149`), and nothing runs it today.

src/strands_evals/experiment.py
```diff
@@ -338,6 +343,8 @@ async def _run_evaluator(
...
 
         @retry(
```
arielnabavian: **consider** (low priority) — routing sits above the `try` on the next few lines, outside this method's error isolation.

`matches()` deliberately treats a raising `condition` as non-fatal, but anything raising inside `route()` itself escapes `_run_evaluator` and lands in the case-level handler at `:566`, which marks **every** evaluator on that case failed with `"An error occurred: ..."`. `routed.model = model` is enough to trigger it for a `__slots__` or frozen custom evaluator.

So a routing bug degrades from "advisory, skip the rule" to "lose the whole case". Moving these two lines inside the `try` is the entire fix.

strandly-the-agent (commented):
*Superseded by my consolidated review below ([review 5045343142](https://github.com/strands-agents/evals/pull/325#pullrequestreview-5045343142)) — this earlier draft was published prematurely by one of my review passes. Its findings (round-trip no-op, condition fall-through downgrade, matching semantics, SDK name collision) are folded into the consolidated review's inline comments and Questions.*

src/strands_evals/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .evaluation_data_store import EvaluationDataStore
 from .experiment import Experiment
 from .local_file_task_result_store import LocalFileTaskResultStore
+from .model_router import ModelRouter, RoutingRule
```
strandly-the-agent, line 7: 🔴 **`strands_evals.ModelRouter` collides with the core SDK's `strands.models.ModelRouter` — same name, unrelated semantics, and evals users import both packages.**

The SDK class is a failover plugin among candidate models within one invocation; this one picks a judge model per evaluation. `RoutingRule` also lands next to the SDK's `RoutingStrategy`/`RoutingCandidate`, so the shared vocabulary now means two different things. Anyone using both packages hits this from their first import.

Suggestion: rename to something that says what it selects (e.g. `EvaluatorModelPolicy` / `JudgeModelSelector`), or drop the new nouns entirely in favor of a callable parameter (see Questions in the summary). Cost: a mechanical rename across 4 files today; after release it's a deprecation cycle.

src/strands_evals/experiment.py
strandly-the-agent, line 347: 🟡 **A failure inside routing fails every evaluator for the case and drops the task output from the report — including discarding a healthy evaluator's already-computed result.**

Reproduced: a control run (no router) passes 2/2 evaluators with `actual_output='hi'`; the identical run with a router reports both as `pass=False, "An error occurred: property 'model' of ... has no setter"` and the report row loses `actual_output`. Reached by a user-defined evaluator (a documented extension point) whose `model` isn't a plain settable attribute or that isn't `copy.copy`-able; no built-in evaluator triggers it.

Suggestion: keep routing advisory, per the module's own "opt-in and advisory" contract — degrade to the unrouted evaluator with a warning (this must stay above the retry closure, which default-binds `evaluator`). Cost: a genuine routing bug surfaces as a warning rather than a loud failure.

```suggestion
        if self._model_router is not None:
            try:
                evaluator = self._model_router.route(evaluator, evaluation_context)
            except Exception as e:
                logger.warning(
                    "evaluator=<%s>, error=<%s> | model routing failed, running evaluator as configured",
                    evaluator.get_name(),
                    e,
                )
```

src/strands_evals/model_router.py
strandly-the-agent, line 137: 🟡 **Routing is a silent total no-op for any experiment loaded from a file, and nothing observable records whether routing happened.**

`Evaluator.to_dict()` writes a default `model=None` out as a concrete `model_id`, and `from_dict` maps it back to an explicit `model=` — so after any `to_file`/`from_file` round trip this guard sees every LLM evaluator as "explicitly configured" and does nothing. That's the whole CLI path (`strands-evals run`/`validate`) plus anyone who persists experiments, and the only trace of a routing decision is a DEBUG log — nothing in the report or spans says which model judged a case.

Suggestion: in this PR, document the limitation here and on `Experiment(model_router=...)`, pin it with a characterization test, and put the resolved model on the eval span (or a report field) so "did it route?" is answerable. The underlying serializer behavior is pre-existing — filed as #385, since fixing it changes behavior an existing test asserts.

src/strands_evals/model_router.py
```diff
@@ -0,0 +1,151 @@
...
+            return False
+        if self.condition is not None:
+            try:
+                return bool(self.condition(evaluation_data))
+            except Exception as e:
+                logger.warning(
+                    "rule_types=<%s>, error=<%s> | routing condition raised, skipping rule", self._type_names, e
+                )
+                return False
+        return True
```
strandly-the-agent, line 90: 🟡 **An `async def` condition makes the rule match every case — the exact inverse of a predicate that returns `False`.**

`bool(coroutine)` is always `True`, so an async predicate (an easy mistake in a framework whose evaluators and runner are all async) routes every case to that rule's model, with only a `RuntimeWarning` buried in the logs — a silently wrong, possibly far more expensive judge on all cases.

Suggestion (also add `import inspect` to the imports):

```suggestion
        if self.condition is not None:
            try:
                result = self.condition(evaluation_data)
            except Exception as e:
                logger.warning(
                    "rule_types=<%s>, error=<%s> | routing condition raised, skipping rule", self._type_names, e
                )
                return False
            if inspect.isawaitable(result):
                if inspect.iscoroutine(result):
                    result.close()
                logger.warning(
                    "rule_types=<%s> | routing condition must be synchronous, skipping rule", self._type_names
                )
                return False
            return bool(result)
        return True
```

src/strands_evals/model_router.py
```diff
@@ -0,0 +1,151 @@
...
+    """A rule mapping evaluator types (and optionally case properties) to a model.
+
+    Attributes:
+        evaluator_types: Evaluator classes or class names this rule applies to.
+        model: The model (Model instance or Bedrock model-id string) to use when
+            this rule matches.
+        condition: Optional predicate over the evaluation data. When provided,
+            the rule only matches cases for which it returns True. This enables
+            complexity-based routing (e.g., trace length, span count).
+    """
```
strandly-the-agent, line 57: 🟡 **This public docstring promises things the code doesn't do:** `evaluator_types` is documented as an attribute but never stored (`hasattr` is `False`); the predicate is presented as taking a `case` when it receives an `EvaluationData` (a different exported type — no `session_id` here); exact class-name matching is unstated, so a rule for `OutputEvaluator` silently doesn't cover the shipped `MultimodalOutputEvaluator`; and a raising condition doesn't just "skip" — it falls through to later rules and `default_model`, silently *downgrading* an escalation rule like the module's own example.

Suggestion (verified against ruff/format; also rename `case` → `data` in the module example above):

```suggestion
    Attributes:
        model: The model (Model instance or Bedrock model-id string) to use when
            this rule matches.
        condition: Optional predicate over the case's `EvaluationData` (note: not
            a `Case`). When provided, the rule only matches cases for which it
            returns True, which enables complexity-based routing (e.g., trace
            length, span count). It must be synchronous and side-effect-free: the
            `EvaluationData` it receives is the live object that later lands in
            the report. A predicate that raises is logged and the rule is skipped,
            so matching falls through to the next rule and then to the router's
            `default_model` - which can silently downgrade the model tier.

    Note:
        `evaluator_types` is matched by exact class name (`get_type_name()`), so a
        rule for `OutputEvaluator` does not cover subclasses such as
        `MultimodalOutputEvaluator` - list those explicitly. The constructor
        argument is not retained as an attribute.
    """
```

tests/strands_evals/test_model_router.py
```diff
@@ -0,0 +1,180 @@
...
+        router = ModelRouter(rules=[RoutingRule(evaluator_types=[TrajectoryEvaluator], model=FAST_MODEL)])
+        evaluator = OutputEvaluator(rubric="Test rubric")
+
+        routed = router.route(evaluator, evaluation_data)
+
+        assert routed is evaluator
+        assert routed.model is None
+
+
+class TestExperimentIntegration:
```
strandly-the-agent, line 130: 🟡 **The feature's headline — per-case, condition-driven routing — is never exercised through the runner.** Both integration tests use one case and a condition-free rule, so a refactor that memoizes model selection per evaluator type, or routes on pre-execution data, passes the whole suite while silently mis-routing (checked by mutation: both survive).

Suggestion: one test closes this — two cases, a `condition` on `actual_output` length, `max_workers=2`; assert the two `Agent(model=...)` kwargs via `call_args_list` are the fast and strong models respectively, and `evaluator.model is None` afterwards (no leak onto the shared instance). Such a test was written and confirmed to pass on this head and kill both surviving mutants. While here: `Mock()` without `spec=Agent`, the dead `mock_agent.return_value = mock_result` (the runner goes through `invoke_async`), and `call_args` reading only the *last* call all become traps the moment a second case exists — a small shared `_agent_mock()` fixture removes the duplication too.

strandly-the-agent (changes requested):
**Requesting changes** — the implementation is clean and does what the PR says (its "454 tests pass" reproduced exactly; concurrency and copy-isolation verified under `max_workers=8`); what's unsettled is the design, plus five reachable issues inline.

- 🔴 `ModelRouter` collides with the core SDK's `strands.models.ModelRouter` — renaming is free now, a breaking change later.
- 🟡 inline: a routing failure fails every evaluator for the case (reproduced); routing silently no-ops on any persisted/CLI-loaded experiment; an `async def` condition matches every case; the public docstring documents an attribute that doesn't exist and hides exact-class matching; the headline per-case routing has no end-to-end test.

Two things need a maintainer independent of the code: (1) #323 was filed by the PR author 2 minutes before this PR and its own "if this direction looks right" was never answered — per AGENTS.md that alignment should come first, and epic #88 explicitly preferred keeping `Experiment` unchanged; (2) this changes public API (2 new exports + an `Experiment` param) — this repo has no api-review label, so `design` looks like the closest lever — and given the two new nouns it likely warrants an API discussion rather than a solo approve.

Pre-existing serializer bug factored out to #385. All of this is offered for a human to weigh — dismiss freely.

<details><summary>How this was reviewed + evidence</summary>

Reviewed `pr-325` @ `03e9295` against `main` @ `771a3fe` (+341/−0, 4 files). Pipeline: triage → context build → 5 independent passes (correctness, issue-alignment, API design, adversarial, test quality) → aggregation. The API-design and adversarial passes timed out on the deep-analysis tier and were re-run on the standard tier; every load-bearing finding was re-verified during aggregation.

| Check | Result |
|---|---|
| `pytest tests/strands_evals/test_model_router.py tests/strands_evals/test_experiment.py -q` | ✅ 125 passed (**14** new — PR body says 12) |
| PR body's "454 tests pass" | ✅ reproduced exactly (454 passed) |
| `ruff check` + `ruff format --diff` on both new files | ✅ clean |
| mypy | ⚠️ not run (internal error in review sandbox); note `routed.model = model` only type-checks because `attr-defined` is disabled repo-wide |
| Repro: `async def` condition returning `False` | ✅ routes anyway |
| Repro: `to_dict()`→`from_dict()`→`route()` | ✅ router no-ops (same object returned) |
| Repro: `route()` raising mid-case (read-only `model` property) | ✅ control 2/2 pass → routed 0/2 pass, `actual_output` lost |
| Both suggested fixes | ✅ verified: tests + ruff/format pass; also verified moving the hook *inside* the `try` would silently disable routing (the retry closure default-binds `evaluator`) |
| `copy.copy` audit of all 21 model-bearing evaluators | ✅ no cached `Agent`, no state derived from `model` — shallow copy + rebind is sound |
| Per-case routing, `max_workers=8`, 8 cases | ✅ correct per case; no leak onto the shared instance |
| Mutation testing (7 mutants) | 4 killed; **2 survived** (memoized `select_model`; routing on pre-execution data) → test finding |

Pass verdicts: correctness *changes-requested* · issue-alignment *aligned, gate on design approval* · API design *needs-design-discussion* · adversarial *broke-it (1 reproduced escalation)* · test-quality *gaps-found*. No live model calls were made; a clean full-`tests/` run wasn't obtained (sandbox killed it twice at ~42%) — the affected suites above are green.
</details>

<details><summary>Questions</summary>

**Blocking (maintainer call, not code):**
1. Is this direction approved? #323 is self-filed, uncommented, and closes with "ready to submit as a PR *if this direction looks right*". In-runner hook, or an external `compare/`-style layer (per #88's stated v1 preference)?
2. Do two classes and five public members earn their keep over a plain callable? `condition` being an arbitrary callable already forfeits the declarative benefits (it's why `model_router` can't serialize). An `evaluator_model: Callable[[Evaluator, EvaluationData], Model | str | None]` ctor param gives the same power with zero new exports (the name collision evaporates), `isinstance`/`name=` routing for free, and typos become `NameError`s instead of silent dead rules — and matches the repo's `task=` / `@eval_task` idiom. Alternative: drop `condition` for v1 and make it declarative Pydantic data that round-trips.
3. Apply the `design` label and take the two new public nouns to an API discussion? (This repo has no api-review label workflow; `design` is the closest lever.)
4. Does `default_model` belong in v1? It's the blanket single-tier behaviour #323's own Alternatives section rejects, and it's what makes the subclass-matching gap silent.

**Non-blocking:**
5. Should `matches`/`select_model` be public? `route()` is their only caller, and `select_model` bypasses both of `route()`'s eligibility guards (it returns `default_model` even for `Contains`, which has no `model` at all).
6. Should type matching be `isinstance` (subclasses free), and/or accept an evaluator's `name=` as a rule key? Today two `Contains` instances distinguished by `name=` can never route differently, though `name=` exists precisely to disambiguate them.
7. Hand-rolled top-level classes vs the #210 `DiagnosisConfig` precedent (Pydantic under `types/`, exported from both namespaces) — deliberate? Worth a line in the PR description either way.
8. README entry expected? #210 added +57 README lines and every comparable feature has a "Features at a Glance" subsection; nothing mentions `model_router` in README or docs today.
9. `RedTeamExperiment` doesn't forward `model_router` and `ChaosExperiment` doesn't expose it, though both run LLM judges — intentional for v1?
</details>

<details><summary>Reading order</summary>

1. `src/strands_evals/model_router.py` — the whole feature: `route()` first (the two guards and the copy are the entire contract), then `RoutingRule.matches`, then the module docstring/example as the user-facing spec.
2. `src/strands_evals/experiment.py` — the 2-line hook; read upward to the retry closure (which constrains where the hook can live) and downward to the error-isolation `try` (what the hook sits outside of).
3. `src/strands_evals/__init__.py` — the permanent public surface; the naming and callable-vs-classes questions live here.
4. `tests/strands_evals/test_model_router.py` — the two integration tests are the behavioural contract; what they don't cover is the test finding.
5. Context outside the diff: `evaluators/evaluator.py` `to_dict` + `experiment.py` `from_dict` (the round-trip no-op), and #210 / README for the precedent this diverges from.
</details>

<details><summary>Appendix — non-blocking (12)</summary>

1. ⚪ A typo'd or empty `evaluator_types` yields a silently dead rule; an evaluator *instance* fails with a cryptic `AttributeError`. Cheap: `ValueError` on empty, and a warning at `Experiment.__init__` for rule names matching none of the configured evaluators (the list is known there). Contrast the SDK sibling, which validates its inputs.
2. ⚪ `select_model` applies neither of `route()`'s eligibility guards — document as rule-resolution-only or make it private (also Q5).
3. ⚪ `experiment.py` docstring lists `model_router` under `Attributes:` but no attribute or property exists (`diagnosis_config`, the precedent, is deliberately not listed).
4. ⚪ The guard `getattr(...) is not None or not hasattr(...)` reads backwards; a `getattr(evaluator, "model", _MISSING)` sentinel states the intent directly.
5. ⚪ `rules`/`default_model`/`model`/`condition` are public-mutable with no re-validation; `ModelRouter()` and `RoutingRule(evaluator_types=[])` are legal dead configs.
6. ⚪ (contrived) A user `__copy__` returning `self` breaks the no-mutation guarantee, which the `route()` docstring states unconditionally.
7. ⚪ `RoutingRule(model=None)` (violates the type hint, unvalidated) matches and then silently skips `default_model` too.
8. ⚪ Three meanings of `None` meet on this surface (router `default_model=None` = leave alone; evaluator `model=None` = strands' default `BedrockModel`; `DEFAULT_BEDROCK_MODEL_ID` = serialization-only) — one clarifying docstring line would help.
9. ⚪ The module's own example re-stringifies the whole trajectory once per (evaluator, rule) pair per case, unmemoized — a span-count predicate would be a better example.
10. ⚪ New file uses RST-style double backticks where AGENTS.md mandates single — pattern-consistent with ~20 existing files, so noted only.
11. ⚪ `Callable` from `typing` while `experiment.py` uses `collections.abc` — AGENTS.md allows either; matching `experiment.py` would be tidier.
12. Pre-existing (filed #385): `Evaluator.to_dict()` erases the `model=None` vs explicit-model distinction — the root enabler of the round-trip no-op finding; needs its own decision since an existing test asserts today's behaviour.
</details>

pdebjyot:
Thank you for this — the depth is appreciated, and I agree with the verdict. Taking the findings and the design question in turn.

**The three findings are real — I reproduced all of them against `03e9295`.**

1. **Round-trip disables routing.** Confirmed, and it's two compounding bugs. `Evaluator.to_dict()` writes `model=None` out as `model_id=DEFAULT_BEDROCK_MODEL_ID`, `Experiment.from_dict()` restores it as an explicit `model=`, *and* `Experiment.to_dict()` never serializes the router. So after `to_file`/`from_file` every judge is pinned to Sonnet, the `model is None` sentinel that gates `route()` is destroyed, and the router is gone — with no warning, on the CLI's load path. This is the most serious of the three.
2. **Fail-open conditions.** Confirmed: a raising `condition` is warn-logged and treated as no-match, then falls through to `default_model`, silently swapping the measurement instrument for a cheaper one. Agree fail-open is the wrong default for an eval library.
3. **Name-equality matching + evaluator-invisible conditions.** Confirmed: `evaluator_types` matches on `__name__` string equality (a subclass doesn't match its base), and `condition` only receives `EvaluationData`, so two evaluators distinguished by `name=` can't be routed differently. The rule surface really is "larger than a callable and less capable than one."

**I'm pivoting to the callable seam you sketched.** `Experiment(model_selector: Callable[[Evaluator, EvaluationData], Model | str | None] | None = None)`, consulted for every evaluator, `None` = leave as-is, exceptions propagate (fail-fast), runner owns copy-and-apply, no `model is None` eligibility sentinel ("explicit wins" becomes visible user policy). The deciding argument for me is the one-way door: callable→rules is additive later (a future rule-set object can implement `__call__` and drop into a callable parameter), rules→callable is a deprecation — so for unreleased public names the reversible choice wins. It also dissolves all three findings at once (no sentinel to poison, fail-fast, receives the evaluator so `isinstance` and per-`name` routing both work), keeps the declarative door open, and sidesteps the `strands.models.routing.ModelRouter` name collision entirely. If a one-knob `default_judge_model=` is wanted as the v0 for "all judges on haiku," happy to add it alongside.

**What I've split out already:** the `model=None` → `DEFAULT_BEDROCK_MODEL_ID` serialization rewrite is a pre-existing bug on `main`, independent of routing, and it silently pins every reloaded experiment's judges regardless of this feature. I've put it up as its own small PR (#392) so it can land on its own merits; it's also a prerequisite for any "explicit model wins" logic to survive persistence.

**What I'd ask a maintainer to settle before I re-push the routing surface:**
1. **Priority** — is per-`(evaluator, case)` judge routing something this package wants to own? The #323 two-tier production pattern (materially lower judge cost at equal accuracy) is the motivation; I'd like an explicit yes/no before shipping any public surface.
2. **Provenance first?** Reports don't record which model scored each case today. Routing is unauditable without per-case judge-model provenance in `EvaluationReport` — happy to build that first as the foundation if you agree it's the prerequisite.
3. **`design` label** on both this and #323, since the repo has no API-review workflow.

I've kept the current branch as-is for reference rather than force-pushing, so the reproductions above still line up with the code. Glad to pair on the `model_selector` shape or the provenance piece whenever there's a direction.

Issue #319, opened 2026-07-21
[FEATURE] Add OpenAI Agents SDK + Traceloop session mapper


liramon2:
### Problem Statement

Teams building AI agents with OpenAI Agents SDK cannot use strands-evals to evaluate their agent traces. Without a dedicated mapper, OpenAI users must either manually transform their traces into Session format or forgo trace-based evaluation entirely.

### Proposed Solution

Add mapper for OpenAI Agents instrumented with traceloop. It extends `GenericGenaiSessionMapper` because traceloop traces follow GenAI conventions.

### Use Case

- As a user of OpenAI Agents SDK and Traceloop OpenLLMetry, I want to evaluate my agent with strands-evals without rewriting my trace collection code

liramon2:
Update 8/11:
- Refactoring implementation to use #309 (merged yesterday)
- Split OpenAI + OpenInference combination into separate issue

PR #326, opened 2026-07-23
feat: add ADK mapper

liramon2:
## Description
The evals framework currently cannot evaluate agents built with Google's Agent Development Kit (ADK), which produce OpenTelemetry spans under the `gcp.vertex.agent` scope with a unique attribute layout. This PR adds `ADKOtelSessionMapper` to map ADK spans into Sessions.

## Related Issues

#328  

## Type of Change

New feature

## Testing

## Checklist

----


liramon2:
@strandly-the-agent review this PR.

src/strands_evals/mappers/adk_otel_session_mapper.py
```diff
@@ -0,0 +1,634 @@
...
+        attrs = call_llm_span.get("attributes", {})
+        raw = attrs.get("gcp.vertex.agent.llm_response", "")
+        if not raw or raw == "{}":
+            return None
+        return self._safe_json_parse(raw) if isinstance(raw, str) else None
+
+    def _extract_user_prompt_from_request(self, llm_request: dict) -> str:
+        """Extract the first user text from llm_request.contents."""
+        for content_item in llm_request.get("contents", []):
+            if content_item.get("role") == "user":
```
strandly-the-agent, line 450: ✅ **Fixed at `fa40cc2`.** The mapper now chooses the latest chronological `call_llm` and scans its request contents backward for the latest user text. The new multi-turn regression covers this path, and the original repro now returns `HISTORY ('latest', 'answer-2')`.

src/strands_evals/mappers/adk_otel_session_mapper.py
```diff
@@ -0,0 +1,634 @@
...
+            elif role == "model":
+                assistant_content: list[TextContent | ToolCallContent] = []
+                for part in parts:
+                    if "text" in part:
+                        assistant_content.append(TextContent(text=part["text"]))
+                    elif "function_call" in part:
+                        func_call = part["function_call"]
+                        name = func_call.get("name", "")
+                        args = json.dumps(func_call.get("args", {}), sort_keys=True)
+                        tool_call_id = self._resolve_tool_call_id(
```
strandly-the-agent: ✅ **Superseded by the consolidated current-head review.** Commit `28fd2143` removes the lossy heuristic correlation that caused the original ID swaps. The remaining requirement—preserving authoritative IDs already present in ADK payloads—is finding 1 in the single review summary: https://github.com/strands-agents/evals/pull/326#pullrequestreview-4774350882

src/strands_evals/mappers/adk_otel_session_mapper.py
```diff
@@ -0,0 +1,634 @@
...
+            span_id = self._extract_span_id(span)
+            if span_id:
+                spans_by_id[span_id] = span
+            parent_span_id = self._extract_parent_span_id(span)
+            if parent_span_id:
+                children_by_parent[parent_span_id].append(span)
+
+        # Build tool_call_id lookup keyed by (name, args_json), ordered by start_time.
+        tool_call_id_lookup: dict[tuple[str, str], list[str]] = defaultdict(list)
+        for span in sorted(spans, key=lambda s: s.get("start_time", 0)):
```
strandly-the-agent: ✅ **Fixed and superseded by the consolidated current-head review.** Commit `28fd2143` normalizes the supported ADK numeric/UTC timestamp forms before sorting and handles `attributes=None`. All remaining current findings are consolidated here: https://github.com/strands-agents/evals/pull/326#pullrequestreview-4774350882

src/strands_evals/mappers/utils.py
```diff
@@ -107,6 +108,9 @@ def detect_otel_mapper(spans: list[Any]) -> SessionMapper:
         if scope_name in SCOPES_OPENINFERENCE_FAMILY:
             return OpenInferenceSessionMapper()
```
strandly-the-agent, line 132: ✅ **Fixed at `fa40cc2` via the scope-only option from the original finding.** The revision removes `is_adk_span`, the scope-less fallback claim, and its tests, while retaining instrumentation-scope auto-detection. Scope-less inputs remain supported through explicit `ADKOtelSessionMapper` construction.

strandly-the-agent (dismissed):
**TL;DR — Code-wise approve:** no blockers remain at exact head `bb158ee2e7ff80ad27b8d6e144251713213b5144`.

✅ Focused + adjacent tests: **161 passed**, with 11 existing warnings; local tracked diff is clean.
✅ Hosted lint, all 11 unit jobs, integration, CI Gate, and title checks are green; review trigger is waiting.
✅ The latest test-only delta fixes the fixture filename and expectations; the prior fixture blocker is closed.
ℹ️ Latest-only selection for parallel terminal tool results is an accepted, deliberate compatibility limitation—not a defect requiring further changes.
⚠️ GitHub cannot change this edited review's historical `CHANGES_REQUESTED` state; a maintainer must dismiss that state.
ℹ️ API-owner sign-off remains appropriate for the public behavior, although this repository has no API-review labels.

<details><summary><strong>Final adjudication matrix and staged pass breakdown</strong></summary>

### Final adjudication

| Item | Disposition | Basis |
|---|---|---|
| Fixture filename and stale expectations | **Fixed** | The test-only delta aligns the fixture loader and assertions; focused + adjacent tests pass 161/161. |
| Parallel terminal tool results | **Accepted limitation; nonblocking** | The author deliberately chose latest-only behavior for cross-mapper compatibility. `agent_response` is one string, sibling mappers select one terminal assistant response, every individual tool span remains in history, no evaluator consumes an aggregate, and aggregation would introduce a new schema convention. |
| Parallel behavior documentation | **Accepted decision; optional clarity follow-up** | Inline docs promise raw JSON but do not explicitly state latest-only selection for parallel terminal groups. Saying so would improve clarity, but it is not required for this PR. |
| System-prompt substring assertion | **Suppressed** | Exact matching would be stronger, but the current assertion exercises the intended behavior and does not create a substantive regression risk. |
| Prior closed decisions | **Remain closed** | Thought visibility, multipart extraction, merged bookkeeping exclusion, trace/span ordering, `error.type`, scalar raw-JSON wrapping, scope-only detection, the external-dict `TracedHandler` contract, and global tool-catalog policy are not re-raised. Multi-agent ownership remains follow-up #319. |

The earlier triage issue comment (`issuecomment-5084538480`) was based on an incorrect characterization of the delta and is superseded by this exact-head review.

### Staged-review disposition

- **Correctness / adversarial:** no remaining reproducible defect.
- **API / DevX:** accept latest-only parallel selection as the chosen compatibility contract; API-owner sign-off is still appropriate.
- **Test quality:** fixture defect fixed; focused + adjacent suite passes 161 tests.
- **LLM-context:** no blocker; all individual tool spans remain available in history.
- **Docs:** no required change; explicit parallel latest-only wording would be a nonblocking clarification.
- **Aggregation:** zero findings retained. The architecture disagreement is not continued as a blocker.

</details>

liramon2:
@strandly-the-agent I pushed a fix to the blockers above. Re-review this PR. Keep any findings in the same comment and close any prior conversations when they are fixed.

strandly-the-agent:
@liramon2, I verified the latest-prompt and scope-only detection fixes and marked both original threads ✅. The repository guard does not authorize this agent to resolve thread objects, so please resolve these two fixed conversations:
- https://github.com/strands-agents/evals/pull/326#discussion_r3646265147
- https://github.com/strands-agents/evals/pull/326#discussion_r3646265170

Please leave the tool-ID and malformed/timestamp conversations open.

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around). Avoid leaving findings for future reviews. I want them all up-front.

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around).

strandly-the-agent:
@liramon2, the review is now consolidated in the existing single review summary: https://github.com/strands-agents/evals/pull/326#pullrequestreview-4774350882

The repository guard does not authorize this agent to resolve thread objects. Please resolve the two old superseded conversations so the summary is the only active finding location:
- https://github.com/strands-agents/evals/pull/326#discussion_r3646265156
- https://github.com/strands-agents/evals/pull/326#discussion_r3646265160

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around). I pushed fixes to some blockers. For others, I have some responses:
1.Not applicable. TracedHandler captures spans emitted by the Strands SDK's own TracerProvider — it's wired to StrandsInMemorySessionMapper by design. The intended path for ADK evaluation is the same as LangChain/OpenInference: collect spans externally (CloudWatch, OTLP file, etc.), convert to dicts, and pass to ADKOtelSessionMapper either directly or via detect_otel_mapper(). None of the existing dict-based mappers support TracedHandler either — this isn't a gap unique to the ADK mapper.

4.1.ADK's LlmAgent.instruction is [defined](https://adk.dev/agents/llm-agents/) as "a string (or a function returning a string)." The Gen AI SDK's GenerateContentConfig.system_instruction likewise [takes a string](https://googleapis-python-genai-70.mintlify.app/guides/system-instructions). ADK concatenates the instruction + auto-generated agent identity text and writes it as a plain string into config.system_instruction within the gcp.vertex.agent.llm_request JSON span attribute. Verify if your suggestion is accurate.

4.2.Per [Google's thought signatures documentation](https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures), these are "encrypted representations of the model's internal thought process" used to preserve reasoning state across multi-step interactions. They appear as a sibling field on the same part as text or function_call — not as separate thought-only parts. There is no thought: true flag or distinct thought part to filter. Verify if your suggestion is accurate.

5.None of the current mappers scope tool/inference spans to a specific agent. StrandsInMemorySessionMapper and LangChainOtelSessionMapper produce a flat list with no agent attribution at all. OpenInferenceSessionMapper goes further and removes child AgentInvocationSpans entirely. Implementing agent ownership makes the ADK mapper inconsistent with existing mapper behavior. The concern about evaluators attributing "all tools to each agent" is valid but is a downstream consumer issue and out of scope for this PR. 

6.ADK's [parallel tool execution documentation](https://adk.dev/tools-custom/performance/) describes how tools run concurrently via async def functions but makes no mention of synthetic "merged" bookkeeping spans. The [tracing schema docs](https://adk.dev/observability/traces/) describe one execute_tool span per tool invocation — no aggregation span. Verify if the suggestion is accurate.

11.No existing mapper handles dynamic per-step tool catalogs. Every mapper takes tools from a single source and applies them uniformly: Strands reads `gen_ai.agent.tools` from the one `invoke_agent` span. OpenInference/LangChain collect tool names from all ToolExecutionSpans and back-fill a union into the agent span. CloudWatch/OpenSearch do the same from child tool spans. None track "which tools were available at step N vs step M." The ADK mapper's approach (tools from the first `call_llm` request) is consistent with this pattern. Dynamic per-step catalogs would be a new capability no mapper supports today.

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around). I pushed more fixes to the blockers.

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around). I pushed more fixes to the blockers and documented the wrapped scalar result limitation (we're going with { "result": ...} to be consistent with the other mappers).

strandly-the-agent:
**⛔ Triage: BLOCK re-review — clarify with author**

Re-review triggered at `7ebeb4e` (current head), prior reviewed at `68faaa0`. Three misalignments before specialist passes can route:

### 1. Code ↔ claim mismatch
Author states "fixes pushed" in recent comments, but the delta `68faaa0..7ebeb4e` is **comment/docstring shortening only** — zero mapper logic changes. The retained blocker (terminal skip_summarization selection failing in preamble+function_call and parallel final cases) **remains untouched in source code** (`adk_otel_session_mapper.py:207–228`). 

**Clarify:** Are the fixes on a different branch, or is 7ebeb4e a staging commit pending additional changes?

### 2. CI failures block re-review
- **Hosted macOS unit: FAILED**  
- **CI Gate: FAILED**  
- Other unit jobs: CANCELLED  
- (Integration, lint, title: green)

These failures must be resolved before re-review proceeds. **Root causes?**

### 3. Blocker still live in code
The prior review's single retained blocker (terminal-event/group selection in two live ADK 2.5 producer shapes) is still code. The review adjudication table shows it as "blocking". **Is this being addressed in 7ebeb4e, or deferred to a follow-up?**

**Next step:** Reply with clarity on (1) and (3), then fix CI in (2). Once aligned, re-trigger the review.

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around). I pushed more fixes to some blockers and have arguments for others:

Parallel terminal group: `agent_response` is `str` across all mappers, no other framework has this pattern, and no evaluator consumes multi-result aggregates today. We pick the latest tool by timestamp (consistent with every other mapper's "pick last" convention) and documented the limitation inline.

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around). I rebased and pushed compatibility changes for cloudwatch traces.

strandly-the-agent (changes requested):
**TL;DR — changes requested.** The rebase is clean (no content drift) and this commit's camelCase fallbacks are correct, but two chronological-sort call sites still read only the snake_case key, so with raw (unparsed) ADOT input `agent_response`/tool-result selection can pick the wrong span. One fix plus one open question below.

✅ Reviewed exact head `ae0caee`; the rebase is a no-op (tree identical to the previously reviewed `bb158ee`, same merge-base `771a3fe`).
✅ Focused suites `tests/strands_evals/{mappers,extractors,providers/test_cloudwatch_provider.py}`: **320 passed** (23 pre-existing deprecation warnings); `git diff --check` and worktree clean.
🔴 One deterministic defect, reproduced twice: raw-ADOT spans lose chronological ordering at two sort sites.
ℹ️ Local lint/type-check were not run (tooling unavailable in my sandbox); hosted CI was fully green at the prior commit `d534681`, and for `ae0caee` the runs are still in flight (5 Windows unit jobs already green, the rest queued).

<details open><summary><strong>🔴 Finding: the timestamp fallback is missing at two sort sites (`adk_otel_session_mapper.py:185`, `:222`)</strong></summary>

`_create_span_info` correctly falls back to `startTimeUnixNano` (line 489), but the two `sorted(..., key=lambda s: self._parse_timestamp(s.get("start_time")))` call sites do not. For raw ADOT spans (no `start_time` key), `_parse_timestamp(None)` returns `datetime.now(timezone.utc)` (line 550), so each sort collapses to input-list order — silently making `call_llm_spans[0]`/`[-1]` and `tool_descendants[-1]` order-dependent instead of chronological.

Reproduced twice with identical output. With the committed fixture order, `agent_response` is `15 multiplied by 37 is 555.`; swapping only the two `call_llm` entries in the list (`call-llm-1` ↔ `call-llm-2`) yields `{"result": "555"}` — raw tool JSON surfacing as the agent's final answer — while the same reordered input passed through `CloudWatchLogsParser` normalization still yields the correct text. A second repro showed `tool_descendants[-1]` likewise selecting the earliest tool result (`{"result": "FIRST-42"}`) instead of the latest (`{"result": "SECOND-FINAL-555"}`).

Line 185 (12-space indent):

```suggestion
            key=lambda s: self._parse_timestamp(s.get("start_time") or s.get("startTimeUnixNano")),
```

Line 222 (16-space indent):

```suggestion
                key=lambda s: self._parse_timestamp(s.get("start_time") or s.get("startTimeUnixNano")),
```

Regression ask: add an ADOT fixture whose spans are listed out of chronological order. The current `adk_adot_spans.json` is already in order, so `TestAdotFormat` passes either way and would not catch this. Reverting the mapper hunks does fail that test (1 failed / 30 passed), so the field-extraction part of the new coverage does have teeth — only the ordering path is unguarded.

</details>

<details><summary><strong>❓ Question (non-blocking): which consumer passes raw, unparsed ADOT JSON?</strong></summary>

The real `CloudWatchProvider` path is unaffected either way: `cloudwatch_provider.py:149,162` always runs `CloudWatchLogsParser(...).parse()` first, and `cloudwatch_parser.py:173-174` already normalizes `startTimeUnixNano` → `start_time` (plus `traceId`/`spanId`/`parentSpanId`), always emitting an int rather than `None`. Feeding the new fixture through that pipeline gives byte-identical results with the pre-fix and post-fix mapper.

So this commit's fallbacks — and the gap above — matter only for a caller that hands raw ADOT JSON straight to `ADKOtelSessionMapper.map_to_session()`, which is exactly what `TestAdotFormat` does. That is not "the commit is pointless": direct-raw use has precedent in `cloudwatch_session_mapper.py:82,130,159-161`, and `detect_otel_mapper` does route the raw shape correctly because `scope.name` is nested identically. The question is whether raw-direct ingestion is a supported entry point. If it is, that is precisely why the `:185`/`:222` gap is worth fixing, and an end-to-end `detect_otel_mapper`/provider test on ADOT-shaped ADK spans would be worth adding alongside the unit fixture.

</details>

<details><summary><strong>Rebase and prior-review disposition</strong></summary>

- The previously reviewed tree (`bb158ee`) is byte-identical to `d534681` on this branch, and the merge-base with `origin/main` is unchanged (`771a3fe`) before and after the rebase. The only new material is `ae0caee`: mapper `+22/-8`, the new `adk_adot_spans.json` fixture, and the new `TestAdotFormat` class.
- No other camelCase gap remains on the raw ADOT path: `_is_call_llm_span`, `_get_operation_name`, the `attributes`/`status`/`error.type` reads, and the merged-tool filter all use keys spelled identically in both shapes.
- All prior-round adjudications remain closed and are not re-raised: latest-only parallel terminal tool-result selection, the raw `{"result": ...}` scalar telemetry contract, thought filtering, multipart concatenation, `(merged tools)` exclusion, chronological ordering of converted spans, `error.type` with UNSET, scope-only ADK detection, the external-dict/`TracedHandler` contract, the global tool catalog, multi-agent ownership (tracked in #319), and the system-prompt substring assertion.
- No public API surface changed in this delta; API-owner sign-off remains reasonable but is not a blocker, and this repository has no API-review labels.

</details>

<details><summary><strong>Pass disposition</strong></summary>

Triage: GO on a narrow delta, routing **correctness** and **test quality** only. Skipped as not applicable here: API bar-raiser (no public surface change), LLM-context (no tool/prompt/model-facing text changed), docs accuracy (no docs or wording touched); the repro work for this delta was run directly rather than as a separate adversarial pass.

Aside, not a PR finding: `tests/strands_evals/providers/test_langfuse_provider.py` fails identically on `origin/main` in my sandbox (`AttributeError: module 'strands_evals.providers' has no attribute 'langfuse_provider'`), so it is environmental and unrelated to this diff.

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around). I reverted the raw ADOT changes and changed multi-agent ADK span handling by ensuring each invoke_agent span gets its own trace.

strandly-the-agent (changes requested):
**Verdict: 🔴 changes requested** — the new multi-agent split can leak a nested specialist's tool output into the coordinator's `agent_response` (finding 1). Two 🟡 should-fix items, one design question, and a nit below. This **supersedes my prior review** `PRR_kwDOPWQYlM8AAAABHZI9jg` at `ae0caee`; that review's blocker (raw camelCase ADOT handling) is now moot because `86da5bf` reverted it byte-for-byte.

Reviewed exact head `cf40376`. Net-new versus the previously fully reviewed `d534681` is only `src/strands_evals/mappers/adk_otel_session_mapper.py` (+84/-5): the scope filter at `:86` and the `_build_trace` → `_build_traces`/`_build_single_trace` split (`:105-219`).

✅ Focused suites (`mappers`, `extractors`, `test_cloudwatch_provider.py`): **319 passed** (23 pre-existing deprecation warnings); full suite with optional extras: **1704 passed**.
✅ `ruff check` clean; `mypy -p src` success across 212 files; worktree clean at the exact head.
✅ Hosted CI green at this SHA: lint, all 11 unit jobs, integration, CI Gate, PR title (only `Trigger Strands Review` waiting).
🔴 Two cold repros written and re-run with identical output: cross-agent response leak, and a multi-agent-only span drop.
🟡 The entire split path is untested — both behaviors survive mutation with the suite still green.

<details open><summary><strong>Findings (2 reproduced defects, 1 evidence-backed coverage gap)</strong></summary>

### 🔴 1. Cross-agent response leakage — `adk_otel_session_mapper.py:178` and `:297`

`_build_single_trace` is invoked per agent group at `:178`, but it receives the **full-trace** `children_by_parent` index rather than one scoped to that group. The skip-summarization `agent_response` fallback then calls `self._get_descendants(span_id, children_by_parent)` at `:297`, which walks straight through a nested `invoke_agent` boundary.

Repro (run twice, identical output): a coordinator `invoke_agent` whose only `call_llm` response is a delegation `function_call` with no text, a nested specialist `invoke_agent` under the coordinator's `generate_content`, and the specialist running `book_flight` returning `{"result": "SPECIALIST_TOOL_OUTPUT"}`.

```
trace[0] coordinator  tools=['delegate']      agent_response='{"result": "SPECIALIST_TOOL_OUTPUT"}'   (no ToolExecutionSpan of its own)
trace[1] specialist   tools=['book_flight']   agent_response='Booked seat 4A.'
```

The coordinator's evaluated response is the specialist's raw tool payload, which contradicts the new docstring's stated purpose of evaluating each tool call against the correct agent. No one-line `suggestion` is safe here because the fix spans the call site and the traversal: scope the descendant lookup to the agent's own group before converting (build a per-group `children_by_parent`, or stop `_get_descendants` at a nested `invoke_agent` boundary).

### 🟡 2. Unclaimed spans are dropped only in the multi-agent branch — `adk_otel_session_mapper.py:163-171`

A span not claimed in `span_to_agent` is dropped from every `agent_groups[...]` list, but only when more than one `invoke_agent` exists; the 0/1-agent path keeps everything. Repro: an `execute_tool audit` span parented to the root `invocation` span (a sibling of `invoke_agent`). With two agent spans present the mapped tools are `['book_flight']`; removing the specialist `invoke_agent` so the single-agent path runs yields `['book_flight', 'audit']` — identical input, data lost purely because a second agent exists.

Caveat stated plainly: ADK's documented hierarchy nests tool spans under `invoke_agent`, and I could not identify a real ADK producer that emits root-level tool spans, so the real-world impact is unconfirmed. Ranked should-fix rather than blocking. A reasonable fix is to attach unclaimed spans to the nearest enclosing agent (or the earliest agent's trace) instead of discarding them.

### 🟡 3. The split path and the scope filter are untested — `tests/strands_evals/mappers/test_adk_otel_session_mapper.py`

Coverage of the focused mapper suite reports `Missing: 142-181` (85% statement / 78% branch): every line of the split path is dead in the test run. `adk_live_spans.json` contains exactly one `invoke_agent`, and `test_multiple_traces_grouped` uses two different `trace_id`s rather than two agents in one trace. The scope filter's drop path is uncovered too, because the shared `make_span()` helper hardcodes the ADK scope.

Mutation evidence (throwaway copies): forcing the single-trace branch → 246 passed, unchanged; deleting the scope-filter line → 246 passed, unchanged. So neither behavior is verified by any test today. Two regressions were written and confirmed runnable — a coordinator→specialist case asserting one `Trace` per `invoke_agent` with per-agent `available_tools` and tool isolation (passes at HEAD, fails under the forced-single-trace mutation), and a foreign-scope span asserting it is dropped (passes at HEAD, fails with the filter removed). Please add both.

</details>

<details><summary><strong>Design question, smaller questions, and one nit</strong></summary>

**Design question (non-blocking):** per-agent `Trace` splitting is a new cross-mapper pattern. No other mapper emits more than one `Trace` per OTel `trace_id` — `cloudwatch_session_mapper` synthesizes exactly one `AgentInvocationSpan` per `Trace`; `opensearch` and `strands_in_memory` leave multiple agent spans flat in one `Trace`; `langchain_otel` (`:143-152`) and `openinference` (`:256-261`) collapse to the root agent and discard nested ones. Given the others collapse-to-root, is per-agent splitting the convention we want for ADK? Worth making explicit even if the answer is yes.

**Shared `trace_id` across split traces:** no observable breakage found. `TraceExtractor` iterates `session.traces` positionally and never keys by `trace_id`, and `types/trace.py` imposes no uniqueness. One note: `detectors/utils.py:_group_spans_into_traces` (`:71-93`) does regroup by `trace_id` and would re-merge the split groups for the chunked failure-detection prompt path — no data loss or duplication, just flagging in case that is unintended.

**`cloudwatch_provider.py:_extract_output` (`:256-261`):** output is bit-identical before and after this PR, verified by running the same coordinator→specialist spans against both `d534681` and `cf40376`, because a nested `invoke_agent` always starts after its parent. Whether "most recently started agent" is the right definition of final output is a pre-existing question outside this diff, not a finding here.

**Scope filter and companion instrumentation:** a `generate_content` span with scope `opentelemetry.instrumentation.vertexai` is dropped at HEAD but was converted at `d534681`, confirmed by repro. Google's docs pair ADK with `opentelemetry-instrumentation-vertexai`/`google-genai`, but those emit standard `gen_ai.*` attributes rather than the `gcp.vertex.agent.*` payloads this mapper's extractors require, so they most likely produced empty spans before this PR too. Flagging intent only.

**⚪ Nit:** `ruff format --check` would reformat the new list comprehension at `:131-133` (it fits on one line under the 120-column limit). Not CI-blocking — the workflow runs `hatch run test-lint` → `hatch fmt --linter --check`, and the `test-format` script in `pyproject.toml` is wired to no workflow — so this is purely optional.

</details>

<details><summary><strong>Revert and prior-review disposition</strong></summary>

- `86da5bf` reverts `ae0caee` byte-for-byte on this file (`git diff d534681..86da5bf` on the mapper is empty): no camelCase key handling (`traceId`/`spanId`/`parentSpanId`/`startTimeUnixNano`/`endTimeUnixNano`) remains, and the `adk_adot_spans.json` fixture plus `TestAdotFormat` class are gone. The prior blocker is resolved by the revert rather than worked around.
- The new `s.get("start_time")` sort at `:142` is safe: both supported input shapes always set that key — `readable_spans_to_dicts` in `utils.py` sets it unconditionally, and `cloudwatch_parser.py:173` defaults it to `0`.
- Multi-agent ownership was previously deferred to follow-up issue #319 and accepted as out of scope. The author has now implemented it here, so it is legitimately new in-scope code and was reviewed fresh above rather than judged against the earlier deferral.
- Older adjudicated decisions stay closed and are not re-raised: latest-only parallel terminal tool-result selection, the raw `{"result": ...}` scalar telemetry contract, thought filtering, multipart concatenation, `(merged tools)` exclusion, `error.type` with UNSET, scope-only mapper detection, the external-dict/`TracedHandler` contract, and the global tool catalog. My first review `PRR_kwDOPWQYlM8AAAABHJLQIg` is already dismissed and nothing here reopens it.

</details>

<details><summary><strong>Pass disposition</strong></summary>

Triage: GO — small, targeted diff (+84/-5, one file) in code with real attribution risk. Routed: correctness/safety and test quality; the adversarial repro work was run inline given the branch's risk profile. Skipped: API bar-raiser (no public signature changed), LLM-context (no tool-description/prompt/model-facing text changed), docs accuracy (no docs or wording touched).

Aggregation: 6 raw candidates → 3 posted (1 🔴, 2 🟡), 4 non-blocking questions plus 1 nit surfaced, and no item silently dropped.

</details>

src/strands_evals/mappers/adk_otel_session_mapper.py
```diff
@@ -0,0 +1,655 @@
...
+                return datetime.fromisoformat(value)
+            except ValueError:
+                return datetime.now(timezone.utc)
+        if isinstance(value, (int, float)):
+            if value > 1e15:
+                value = value / 1e9
+            return datetime.fromtimestamp(value, tz=timezone.utc)
+        return datetime.now(timezone.utc)
+
+    def _safe_json_parse(self, content: Any) -> Any:
```
poshinchen: I see that this one is duplicated in other mappers? let's move to utils / base SessionMapper class.

src/strands_evals/mappers/adk_otel_session_mapper.py
```diff
@@ -0,0 +1,655 @@
...
+            return value[2:]
+        return value
+
+    def _parse_timestamp(self, value: Any) -> datetime:
+        """Parse timestamp from ISO 8601, numeric epoch, or nanosecond-string format."""
+        if value is None:
+            return datetime.now(timezone.utc)
+        if isinstance(value, datetime):
+            return value
+        if isinstance(value, str):
```
poshinchen: Will this happen? If so we can move "_parse_timestamp" into SessionMapper then it can be re-used on other mappers.

liramon2: The timestamps can be nanoseconds or ISO 8601 strings. I'll move it into the SessionMapper

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around).

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around).

strandly-the-agent (changes requested):
**Verdict: 🔴 changes requested** — both findings from my last review are properly fixed, but the `_parse_timestamp` hoist into `SessionMapper` silently dropped ADK's decimal-string-nanosecond branch, and the delta rewrote the one test that covered it. One 🟡 test-teeth gap besides that.

Reviewed exact head `b9854bc`. Delta since `cf40376`: `_parse_timestamp` and `safe_json_parse` hoisted into `session_mapper.py`/`utils.py` (addressing `poshinchen`'s two inline asks), the two ADK multi-agent fixes, and +258 lines of tests.

✅ Fixed and verified: the coordinator no longer picks up a nested specialist's tool output (`stop_at_agents`), and unclaimed spans are no longer dropped in the multi-agent branch.
✅ Focused suites (`mappers`, `extractors`, `test_cloudwatch_provider.py`): **323 passed** (23 pre-existing deprecation warnings); ADK mapper file alone: **34 passed**; worktree clean.
✅ Hosted CI green at this SHA: lint, all 11 unit jobs, integration, CI Gate, PR title (only `Trigger Strands Review` waiting).
🔴 Those 323 green tests do not cover finding 1 — the one test that did was changed to assert the broken output.
🟡 Deleting the scope-filter line leaves the mapper suite at 250 passed, unchanged.

<details open><summary><strong>Findings (1 🔴 regression, 1 🟡 test gap)</strong></summary>

### 🔴 1. The shared `_parse_timestamp` lost ADK's decimal-string nanosecond branch — `session_mapper.py:77-109`

At `cf40376` the ADK mapper's own helper had `if value.isdigit(): datetime.fromtimestamp(int(value) / 1e9, tz=timezone.utc)` (`cf40376:adk_otel_session_mapper.py:631`). The hoisted version has no such branch, so a digit string falls into the ISO path, `fromisoformat` raises, and it silently returns `datetime.now(timezone.utc)`.

Verified twice, deterministically:

```
_parse_timestamp("1700000000000000000") -> 2026-07-28 16:54:06+00:00   (expected 2023-11-14 22:13:20+00:00)
```

The path is live rather than hypothetical: `cloudwatch_parser.py:173` passes `startTimeUnixNano` through unchanged, and OTLP/protobuf JSON encodes `uint64` as a JSON **string**, so real CloudWatch ADK exports hit it. End-to-end consequence repro — an ADK trace whose `call_llm` children carry string-nanosecond `start_time` and are listed later-first maps `agent_response` to `'EARLIER'` instead of `'LATER'`.

Blast radius is every ordering-critical call site: `:139` agent-span sort (which also picks `first_agent_id`, the target of the new unclaimed-span fallback), `:265` `call_llm` first/last selection (system prompt, tools, `user_prompt`, final response), `:302` `tool_descendants` last-selection, and `:569-570` `SpanInfo` start/end which feed the within-trace sort at `:218` and the cross-trace session sort at `:100`; it reaches `CloudWatchProvider._extract_output` (`cloudwatch_provider.py:256-262`).

Aggravating factor: this delta rewrote the only test that covered this. `tests/strands_evals/mappers/test_adk_otel_session_mapper.py:716` `test_string_nanosecond_epoch_falls_back_to_now` now asserts `ts_str.year >= 2026`, blessing the fallback-to-now, where it previously asserted 2023. Nothing else in the repo feeds a decimal-string nanosecond timestamp through a mapper (checked `test_utils.py`, `test_cloudwatch_parser.py` — its fixtures use ints and assert passthrough only — `test_cloudwatch_provider.py`, and the langchain/openinference mapper tests; there is no `test_session_mapper.py`). The shared docstring at `session_mapper.py:85` also still claims "String-encoded nanosecond epoch → datetime", which is now false.

Restore the branch in the shared helper (`session_mapper.py:97-103`):

```suggestion
        if isinstance(value, str):
            # Decimal-string nanosecond epochs (OTLP JSON encodes uint64 as strings)
            if value.isdigit():
                return datetime.fromtimestamp(int(value) / 1e9, tz=timezone.utc)
            try:
                if value.endswith("Z"):
                    value = value[:-1] + "+00:00"
                return datetime.fromisoformat(value)
            except ValueError:
                return datetime.now(timezone.utc)
```

Please also revert `test_string_nanosecond_epoch_falls_back_to_now` to assert year 2023 (renaming it accordingly), and fix the docstring. Overriding `_parse_timestamp` in `ADKOtelSessionMapper` instead of extending the base is equally fine if you would rather keep the base minimal.

One aside, not a defect: the threshold also moved from `> 1e15` to `> 1e12` (`session_mapper.py:106`). Real nanosecond values are around `1.7e18`, so both behave identically for realistic input — worth aligning back only if you restore the branch above. `langchain_otel`/`openinference` did **not** regress: their pre-refactor bodies already lacked the `isdigit` branch and already used `> 1e12`, so the hoist is a verified no-op for them.

### 🟡 2. The scope-filter test has no teeth — `TestScopeFiltering::test_foreign_scope_spans_are_dropped`

The fixture's foreign-scope span is a `generate_content` span with `parent_span_id: None`, which `_convert_inference_span` (`adk_otel_session_mapper.py:341-357`) discards anyway via `if not messages: return None`. Mutation proof: deleting the scope-filter line in `map_to_session` leaves the mapper suite at **250 passed**, unchanged — so the filter itself is unverified.

Use a fixture that would convert if unfiltered: a foreign-scope `execute_tool` span carrying the full `gen_ai.tool.*` / `gcp.vertex.agent.tool_*` attributes (it converts unconditionally and needs no parent), or give the foreign `generate_content` span a real parent `call_llm` in the same fixture.

</details>

<details><summary><strong>Resolved since my last review, plus one note and the open design question</strong></summary>

- **Prior 🔴 fixed.** `stop_at_agents` prunes correctly: the `continue` at `:605` fires before both the append and the child enqueue, so a nested `invoke_agent`'s entire subtree is skipped. My previous blocker repro now leaves the coordinator's `agent_response` empty instead of leaking `{"result": "SPECIALIST_TOOL_OUTPUT"}`.
- **Prior 🟡 fixed.** An orphan `execute_tool` parented to the root `invocation` span now lands in exactly one trace (the earliest agent's), with no duplication and no loss. The root `invocation` span is swept in as a raw span but is harmless — conversion only handles `invoke_agent`/`generate_content`/`execute_tool` and skips `invocation`/`call_llm`.
- **Prior test gap largely fixed, with real teeth.** `test_coordinator_specialist_produces_separate_traces` asserts 2 traces, per-agent `available_tools`, tool isolation, and `"SPECIALIST_TOOL_OUTPUT" not in coord_agent.agent_response`; `test_unclaimed_spans_assigned_to_earliest_agent` covers the other fix. Mutation results: forcing the single-trace branch → caught (2 failures); removing the `stop_at_agents` pruning → caught; removing the unclaimed-span fallback → caught; deleting the scope filter → not caught (finding 2).
- **`safe_json_parse`** is a verbatim behavior-preserving extraction across all call sites, and the hoist into `SessionMapper`/`utils` addresses `poshinchen`'s two inline asks.

**Note (non-blocking):** with `stop_at_agents`, a pure-delegation coordinator turn now yields an `AgentInvocationSpan` with `agent_response == ''` — the span is still created because `user_prompt` is non-empty, so the guard at `:307` is False. Session-level output stays correct (`_extract_output` returns the specialist's answer, since traces sort chronologically), but a grader reading the coordinator's own span sees an empty response.

**Open design question (unchanged from my last review):** per-agent `Trace` splitting is still a new cross-mapper pattern — other mappers either collapse to the root agent or leave agent spans flat in one `Trace`. Whether a delegating coordinator should surface an empty response, no span at all, or the delegated answer is a maintainer call rather than something I want to decide for you.

</details>

<details><summary><strong>Prior-review disposition</strong></summary>

- This review supersedes my `PRR_kwDOPWQYlM8AAAABHaA_CQ` (`CHANGES_REQUESTED` at `cf40376`) in full; both of its findings are resolved above. My earlier `PRR_kwDOPWQYlM8AAAABHZI9jg` (at `ae0caee`) was already superseded by that one and is likewise stale — both can be dismissed. `PRR_kwDOPWQYlM8AAAABHJLQIg` is already dismissed and nothing here reopens it.
- Older adjudicated decisions stay closed and are not re-raised: latest-only parallel terminal tool-result selection, the raw `{"result": ...}` scalar telemetry contract, thought filtering, multipart concatenation, `(merged tools)` exclusion, `error.type` with UNSET, scope-only mapper detection, the external-dict/`TracedHandler` contract, and the global tool catalog. Multi-agent ownership was originally deferred to follow-up #319 and is now implemented in-PR, so it is reviewed as new code rather than judged against the deferral.

</details>

<details><summary><strong>Pass disposition</strong></summary>

Triage: GO — a small refactor-plus-fix delta touching shared base-class behavior used by three mappers, so real regression risk. Routed: correctness/safety and test quality; adversarial repro work run inline. Skipped: API bar-raiser (no public signature changed), LLM-context (no tool-description/prompt/model-facing text changed), docs accuracy (no docs or wording touched).

</details>

src/strands_evals/mappers/session_mapper.py
```diff
@@ -72,3 +73,37 @@ def _normalize_to_flat_spans(self, data: Any) -> list[dict]:
 
         # Fallback for unexpected types
         return []
+
+    def _parse_timestamp(self, value: Any) -> datetime:
```
poshinchen: you don't need this to be `_`.

strandly-the-agent (changes requested):
**Verdict: 🔴 changes requested — no code change since my last review.** Head is still `b9854bc` (no new commits since `PRR_kwDOPWQYlM8AAAABHhnPqQ` an hour ago), so both findings stand. I re-verified them cold on a fresh clone anyway, twice, with identical output.

✅ Re-verified at exact head `b9854bc`; focused suites (`mappers`, `extractors`, `test_cloudwatch_provider.py`): **323 passed**; worktree clean.
✅ Hosted CI is now fully green on this SHA (lint, 11 unit jobs, integration, CI Gate, title, and `Auto Strands Review` completed).
🔴 The string-nanosecond timestamp regression still reproduces end to end through `CloudWatchLogsParser`.
🟡 The scope-filter test still has no teeth (the filter itself works; the test just cannot fail).
ℹ️ `poshinchen`'s new comment on `session_mapper.py:77` (drop the `_` prefix) touches the same lines as the 🔴 fix — worth doing in one edit; see below.

<details open><summary><strong>🔴 Still open: shared <code>_parse_timestamp</code> drops the decimal-string nanosecond branch</strong></summary>

`session_mapper.py:97-103` handles only ISO strings, so a digit string raises in `fromisoformat` and silently returns `datetime.now(timezone.utc)`. The ADK mapper had `if value.isdigit(): datetime.fromtimestamp(int(value) / 1e9, tz=timezone.utc)` before the hoist (`cf40376:adk_otel_session_mapper.py:631`).

Fresh cold repro, run twice with identical output:

```
A  _parse_timestamp("1700000000000000000")                -> 2026-07-28 17:53:36+00:00   (expected 2023-11-14)
B  CloudWatchLogsParser keeps start_time as str '17000...' -> mapped SpanInfo year = 2026
C  string-ns call_llm spans listed later-first             -> agent_response 'EARLIER'
C' same spans with int-ns timestamps (control)             -> agent_response 'LATER'
```

Row B is the reachability proof: `cloudwatch_parser.py:173` passes `startTimeUnixNano` through unchanged, and OTLP/protobuf JSON encodes `uint64` as a JSON string, so real CloudWatch ADK exports land here. Row C vs C' isolates the cause to the parse, not the ordering logic. Blast radius is unchanged from my last review: the agent-span sort at `:139` (which also picks `first_agent_id`), `call_llm` first/last selection at `:265`, `tool_descendants` at `:302`, and `SpanInfo` at `:569-570` feeding the within-trace sort (`:218`) and cross-trace sort (`:100`), reaching `CloudWatchProvider._extract_output`.

The suite stays green because `test_adk_otel_session_mapper.py:716` `test_string_nanosecond_epoch_falls_back_to_now` asserts `ts_str.year >= 2026` — it encodes the fallback as intended behavior. Please revert that assertion to year 2023 (and rename the test) along with the fix. The docstring at `session_mapper.py:85` still advertises "String-encoded nanosecond epoch → datetime", which is currently false.

Since `poshinchen` also asked to drop the `_` prefix, both can land in one edit — replacing `session_mapper.py:77-103`:

```suggestion
    def parse_timestamp(self, value: Any) -> datetime:
        """Parse timestamp from various formats.

        Handles:
        - None → current UTC time
        - datetime → passthrough
        - ISO 8601 string (with optional trailing Z) → parsed datetime
        - Numeric (int/float) nanosecond epoch → datetime
        - String-encoded nanosecond epoch → datetime

        Args:
            value: Raw timestamp value from a span dict.

        Returns:
            Timezone-aware datetime in UTC.
        """
        if value is None:
            return datetime.now(timezone.utc)
        if isinstance(value, datetime):
            return value
        if isinstance(value, str):
            # OTLP JSON encodes uint64 nanosecond epochs as decimal strings
            if value.isdigit():
                return datetime.fromtimestamp(int(value) / 1e9, tz=timezone.utc)
            try:
                if value.endswith("Z"):
                    value = value[:-1] + "+00:00"
                return datetime.fromisoformat(value)
            except ValueError:
                return datetime.now(timezone.utc)
```

If you take the rename, the call sites to update are `adk_otel_session_mapper.py:139,265,302,569,570`, `langchain_otel_session_mapper.py:454,455`, and `openinference_session_mapper.py:602,603`. Overriding `_parse_timestamp` in `ADKOtelSessionMapper` instead of extending the base is equally acceptable if you would rather keep the base minimal — the requirement is only that string-encoded nanosecond epochs stop resolving to `now()`.

</details>

<details><summary><strong>🟡 Still open: the scope-filter test cannot fail</strong></summary>

The filter itself works — I re-confirmed a foreign-scope `execute_tool` span (fully populated `gen_ai.tool.*` / `gcp.vertex.agent.tool_*`) yields 0 traces at this head. The problem is only the test: `TestScopeFiltering::test_foreign_scope_spans_are_dropped` uses a foreign `generate_content` span with `parent_span_id: None`, which `_convert_inference_span` (`adk_otel_session_mapper.py:341-357`) discards anyway via `if not messages: return None`. Deleting the filter line in `map_to_session` leaves the mapper suite at **250 passed**, unchanged.

Swapping that fixture for the foreign-scope `execute_tool` shape above (it converts unconditionally and needs no parent) makes the test fail when the filter is removed.

</details>

<details><summary><strong>Unchanged from my last review</strong></summary>

Still resolved and credited: the `stop_at_agents` pruning fix, the unclaimed-spans-to-earliest-agent fix, the two multi-agent regressions with real teeth (mutating the split branch, the pruning, or the unclaimed fallback each fails a test), the verbatim `safe_json_parse` extraction, and `poshinchen`'s de-duplication asks addressed by the hoist.

Still closed, not re-raised: latest-only parallel terminal tool-result selection, the raw `{"result": ...}` scalar telemetry contract, thought filtering, multipart concatenation, `(merged tools)` exclusion, `error.type` with UNSET, scope-only mapper detection, the external-dict/`TracedHandler` contract, the global tool catalog, and multi-agent ownership (originally deferred to #319, now implemented in-PR and reviewed as new code).

Still open as a maintainer call, not a defect: per-agent `Trace` splitting remains a new cross-mapper pattern (others collapse to root or stay flat), and a pure-delegation coordinator turn yields an `AgentInvocationSpan` with an empty `agent_response` while session-level output stays correct.

This review supersedes my `PRR_kwDOPWQYlM8AAAABHhnPqQ` at the same SHA; `PRR_kwDOPWQYlM8AAAABHaA_CQ` (`cf40376`) and `PRR_kwDOPWQYlM8AAAABHZI9jg` (`ae0caee`) are stale and can be dismissed.

Routing: no delta since my last pass, so I scoped this to independent re-verification of the two open findings plus the new maintainer comment rather than re-running the full specialist set.

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate your findings in a single comment (not scattered around). We are going with a threshold of 1e12 because it behaves the same and is aligned with the existing mappers.

strandly-the-agent (approved):
**Verdict: ✅ approve — no blockers remain at `48f2b13`.** Both findings from my last review are fixed and each is now pinned by a test that fails when the fix is removed. Nothing new found; this supersedes all my earlier reviews on this PR.

✅ String-encoded nanosecond timestamps parse correctly again — cold repro twice: `parse_timestamp("1700000000000000000")` → `2023-11-14 22:13:20+00:00`; a CloudWatch-parsed span with a string `startTimeUnixNano` now maps to year 2023; string-ns `call_llm` spans listed later-first select `'LATER'`, matching the int-ns control.
✅ The scope-filter test now fails when the filter is deleted (it previously could not); the three multi-agent mutation guards still fail as intended.
✅ The `_parse_timestamp` → `parse_timestamp` rename is complete — no stale references remain in `src/` or `tests/`, and all six call sites across the ADK, LangChain and OpenInference mappers were updated.
✅ Full suite `tests/strands_evals`: **1708 passed** (at `dc1a4ee`, delta since is a pure rename); focused mapper/extractor/provider suites at this head: **323 passed**; `ruff check` and `mypy -p src` (212 files) clean; worktree clean.
✅ Hosted CI green at this SHA: lint, all 11 unit jobs, integration, CI Gate, PR title (only `Trigger Strands Review` waiting).

<details><summary><strong>What landed and how I verified it</strong></summary>

`dc1a4ee` fixed both open findings:

1. `session_mapper.py:97-98` restores `if value.isdigit(): return datetime.fromtimestamp(int(value) / 1e9, tz=timezone.utc)` as the first statement of the string branch — the shape OTLP JSON actually delivers (`uint64` as a decimal string), which `cloudwatch_parser.py:173` passes through unchanged.
2. `test_string_nanosecond_epoch` now asserts `year == 2023 and month == 11` instead of blessing the fallback to `now()`. Removing the restored branch makes it fail (`assert 2026 == 2023`).
3. `TestScopeFiltering::test_foreign_scope_spans_are_dropped` now uses a foreign-scope `execute_tool` span with full `gen_ai.tool.*` / `gcp.vertex.agent.tool_*` attributes — one that would convert unconditionally if unfiltered — instead of a parentless `generate_content` span that `_convert_inference_span` discarded anyway.

`48f2b13` then took `poshinchen`'s naming request: the shared helper is now `parse_timestamp` (`session_mapper.py:77`), with call sites updated at `adk_otel_session_mapper.py:139,265,302,569,570`, `langchain_otel_session_mapper.py:454,455` and `openinference_session_mapper.py:602,603`, plus the test class. I grepped `src/` and `tests/` for leftovers — clean apart from stale `__pycache__` artifacts.

Mutation results (throwaway copies, baseline 250 in `tests/strands_evals/mappers`): deleting the scope filter → 1 failed (was 250 unchanged before `dc1a4ee`); forcing the single-trace branch → 2 failed; stripping the `stop_at_agents` pruning → 1 failed; removing the unclaimed-span fallback → 1 failed; removing the `isdigit` branch → 1 failed. Every behavior I raised across this series is now guarded.

</details>

<details><summary><strong>Accepted, and two things I looked at but am not raising</strong></summary>

**`1e12` threshold — accepted.** Agreed with your reasoning: real nanosecond values are ~1.7e18, so `1e12` and `1e15` behave identically for every input any supported path produces, and matching the existing mappers is the right call. Not reopening it.

**Short/float digit strings — theoretical only.** `isdigit()` claims any all-digit string, so `"2026"` → 1970-01-01T00:00:00.000002 and `"1700000000"` (seconds as a string) → 1970, while `"1.7e18"` isn't `isdigit()` and still falls back to `now()`. I traced every producer feeding this helper: `readable_spans_to_dicts` (`utils.py:230-236`) always yields ints from the OTel SDK, and `cloudwatch_parser.py:173-174` always yields full ~19-digit nanosecond strings per the OTLP JSON mapping. No supported path can produce those shapes, so this is a docstring caveat at most — your call, and fine to ignore.

**Coverage nit.** The string-nanosecond path is pinned at the unit level but no `map_to_session`-level fixture uses a string-typed `start_time`. One fixture would close it if you're already touching these tests; not worth a round trip on its own.

</details>

<details><summary><strong>Housekeeping and closed items</strong></summary>

My earlier reviews are now stale `CHANGES_REQUESTED` states from older heads with no remaining findings: `PRR_kwDOPWQYlM8AAAABHh-C6Q` and `PRR_kwDOPWQYlM8AAAABHhnPqQ` (`b9854bc`), `PRR_kwDOPWQYlM8AAAABHaA_CQ` (`cf40376`), `PRR_kwDOPWQYlM8AAAABHZI9jg` (`ae0caee`). Please dismiss them so the PR's review state reflects reality — GitHub won't clear them automatically and I can't dismiss my own.

Previously adjudicated items stay closed and are not re-raised: latest-only parallel terminal tool-result selection, the raw `{"result": ...}` scalar telemetry contract, thought filtering, multipart concatenation, `(merged tools)` exclusion, `error.type` with UNSET, scope-only mapper detection, the external-dict/`TracedHandler` contract, the global tool catalog, and multi-agent ownership (originally deferred to #319, implemented in-PR and reviewed as new code). Two design calls remain yours rather than defects: per-agent `Trace` splitting is a new cross-mapper pattern (others collapse to root or stay flat), and a pure-delegation coordinator turn yields an `AgentInvocationSpan` with an empty `agent_response` while session-level output stays correct.

Routing: triage GO on a small, well-targeted delta → correctness and test-quality (with mutation testing) run independently; API, LLM-context and docs passes skipped as not applicable. As always, this is solid work for a human to approve — the merge call is yours.

</details>

src/strands_evals/mappers/constants.py
```diff
@@ -8,6 +8,7 @@
 SCOPE_LANGCHAIN_OTEL = "opentelemetry.instrumentation.langchain"
 SCOPE_OPENINFERENCE = "openinference.instrumentation.langchain"
 SCOPE_OPENINFERENCE_SMOLAGENTS = "openinference.instrumentation.smolagents"
+SCOPE_ADK = "gcp.vertex.agent"
```
poshinchen, line 11: @strandly-the-agent do you think this should be `SCOPE_ADK_GCP` or is it actually correct?

tests/strands_evals/mappers/fixtures/adk_live_spans.json
```diff
@@ -0,0 +1,193 @@
...
+      "gcp.vertex.agent.llm_response": "{\"model_version\":\"gemini-3.5-flash\",\"content\":{\"parts\":[{\"function_call\":{\"id\":\"uwpsprd2\",\"args\":{\"expression\":\"15 * 37\"},\"name\":\"calculator\"},\"thought_signature\":\"EpsBCpgBARFNMg_L4ljyIrSpRvhyg-49q7jvZ0A3gE7AeeMXKikwKIpbZ4fCvCZ3x6_-JUZwfIH8G_P6ttdyXsXCLX9PZAZh20PNVDV42YXQLG2IJ7I4O1cKYqwQWLkMKdEqAhSwbBizYp9no3B72reIWkkpav_C_8ZxfOM4841-reXWyMVvQCXM9nqxDd8sHt2zcH5OYOY8hsZDBqc=\"}],\"role\":\"model\"},\"finish_reason\":\"STOP\",\"usage_metadata\":{\"candidates_token_count\":19,\"prompt_token_count\":157,\"prompt_tokens_details\":[{\"modality\":\"TEXT\",\"token_count\":157}],\"thoughts_token_count\":32,\"total_token_count\":208}}",
```
poshinchen: nit: the signature is a bit itchy here.. consider using something mock

src/strands_evals/mappers/constants.py
poshinchen: nvm, it's correct

src/strands_evals/mappers/adk_otel_session_mapper.py
```diff
@@ -0,0 +1,637 @@
...
+                - List of trace objects: [{"trace_id": "x", "spans": [...]}, ...]
+            session_id: Session identifier.
+
+        Returns:
+            Session object ready for evaluation.
+        """
+        spans = self._normalize_to_flat_spans(data)
+
+        # Filter to only spans from the ADK instrumentation scope.
+        # Include spans with no scope (e.g. to_json format) since this mapper
```
poshinchen, line 83: hmm why? what's the impact?

src/strands_evals/mappers/adk_otel_session_mapper.py
```diff
@@ -0,0 +1,637 @@
...
+        for trace_id, trace_spans in grouped.items():
+            traces = self._build_traces(trace_id, trace_spans, session_id)
+            result_traces.extend(t for t in traces if t.spans)
+
+        # Sort traces chronologically by earliest span start_time
+        result_traces.sort(key=lambda t: min(s.span_info.start_time for s in t.spans))
+
+        return Session(session_id=session_id, traces=result_traces)
+
+    def _build_traces(self, trace_id: str, spans: list[dict], session_id: str) -> list[Trace]:
```
poshinchen, line 104: why is this `list[Trace]` instead of `Trace`?

liramon2: To provide a trace for each agent in a multi-agent setup. I only included `_build_trace` originally, but that led to the low scores we saw yesterday because outputting a single trace for a multi-agent system gave it multiple `invoke_agent` spans. `TraceExtractor` only finds the first one, so it misses attributes like `available_tools` on the other `invoke_agent` spans. Outputting multiple traces solves this by splitting each `invoke_agent` span into its own trace.

src/strands_evals/mappers/adk_otel_session_mapper.py
liramon2: It's defensive code in case the ADK spans arrive alongside non-ADK spans (ex. HTTP client spans, database spans)

tests/strands_evals/mappers/fixtures/adk_live_spans.json
liramon2: Addressed in 41b7de6e27e571918f506e67457bd57c5daaca0f

src/strands_evals/mappers/adk_otel_session_mapper.py
poshinchen: > TraceExtractor only finds the first one, so it misses attributes like available_tools on the other invoke_agent spans.
Shouldn't we fix this? I assume this will fail in other frameworks too?

liramon2: I ran a multi-agent experiment with the langchain OTel mapper and can confirm that I'm seeing similarly low scores. So yes, it is worth fixing. Though, that sounds like it should be in a separate PR. I can revert to the original ADK implementation if we're fixing it in that route.

poshinchen: Let's make it as follow-up

poshinchen:
@strandly-the-agent review the PR before I approve it.

poshinchen (approved):

strandly-the-agent (approved):
**Verdict: ✅ approve — no blockers at `1473ea4`, and the force-push introduced no new risk.** Re-reviewed from scratch on the rebased head; everything I previously blocked on stays fixed and every guard is still live.

⚠️ **One mea culpa first:** my answer to your `SCOPE_ADK` question sat in an unsubmitted pending review since yesterday, so you never saw it — it was stale against the old head and I've cleared it. Substance below; you'd already reached the same conclusion ("nvm, it's correct"), and you were right.

✅ **Source is unchanged from what I approved at `48f2b13`** — all seven PR-owned source/test files are byte-identical; the only new PR content is the `thought_signature` → `MOCK_SIGNATURE` fixture redaction. The rest of `git diff 48f2b13..HEAD` is `main` moving (merge-base is now `c5461d0`).
✅ Full suite `tests/strands_evals`: **1715 passed** (with the `langfuse`/`otel` extras); focused mapper/extractor/provider suites: **323 passed**; ADK mapper file alone: **34 passed**.
✅ `ruff check src tests` clean; `ruff format --check` clean on the PR files; `mypy -p src` reports only 2 `import-not-found` errors for optional extras, identical on the pre-PR merge-base — environmental, not yours.
✅ Cold repros at this head, run twice: string-nanosecond timestamps → 2023 and order correctly; the multi-agent split gives one Trace per `invoke_agent` with no cross-agent tool leakage and no dropped orphan tool span; a fully-populated foreign-scope tool span is dropped; the committed fixture maps end to end (tool id `uwpsprd2`, tokens 157+32 / 222+20).
✅ Hosted CI green at this SHA: lint, all 11 unit jobs, integration, CI Gate, PR title.

<details><summary><strong>The fixture redaction is inert — verified, not assumed</strong></summary>

`thought_signature` appears nowhere in `src/` or `tests/` outside the fixture JSON: the mapper only reads the `thought` boolean marker (`adk_otel_session_mapper.py:452,481,502,518,544`), and no test asserts on the signature, its length, or the surrounding part shape.

Stronger check than reading the code: I mapped the **old** fixture (`git show 48f2b13:…adk_live_spans.json`, real base64 signatures) and the **new** one through `map_to_session()` and diffed the entire resulting `Session` tree — byte-identical. All three occurrences sit in that ignored sibling field (turn-1 `llm_response`, the embedded history in turn-2's `llm_request`, turn-2's `llm_response`). The load-bearing `thought: true` filtering keeps its own synthetic coverage at `test_adk_otel_session_mapper.py:311-316`, untouched. The fixture still parses as valid ADK 2.5 shape — seven spans, one scope, embedded JSON payloads all well-formed — so the redaction cost nothing but the secret-looking blob.

</details>

<details><summary><strong>Rebase interactions and mutation guards</strong></summary>

**Rebase.** The main-side delta the force-push pulled in (`output_evaluator.py`, `trajectory_evaluator.py`, the multimodal evaluators, a `pyproject.toml` bump) is additive `tools: list[Any] | None` plumbing. It touches neither `AgentInvocationSpan.agent_response`, nor `Trace`/`Session` shape, nor anything keyed on `trace_id` — so the two ADK-specific shapes worth worrying about (an empty `agent_response` on a pure-delegation coordinator turn, and multiple `Trace` objects sharing one `trace_id`) don't meet any new assumption. The green full suite confirms it.

**Mutation guards, all four still live** (run against throwaway copies via `PYTHONPATH`, since the editable install otherwise resolves back to the pristine tree and the mutation silently no-ops):

| Mutation | Result |
|---|---|
| Force `len(agent_spans_raw) <= 1` always true | 2 failed (`TestMultiAgentSplitting`) |
| Delete the scope-filter line in `map_to_session` | 1 failed (`test_foreign_scope_spans_are_dropped`) |
| Strip the `stop_at_agents` pruning in `_get_descendants` | 1 failed (specialist-tool-leak assertion) |
| Remove the `isdigit` branch in `parse_timestamp` | 1 failed (`test_string_nanosecond_epoch`) |

Every behaviour I raised across this review series fails at least one test when broken — that's what makes this safe to land.

</details>

<details><summary><strong>The <code>SCOPE_ADK</code> answer that got stuck (for the record)</strong></summary>

The **value is correct**: verified against `google-adk==2.5.0`, where ADK's tracer registers with `instrumenting_module_name="gcp.vertex.agent"` (`google/adk/telemetry/tracing.py:102-105`), and the same string is used for its logger (`:109`) and meter (`telemetry/_metrics.py:43`). That's the scope name landing on every span this mapper consumes.

On the **name**, `SCOPE_ADK` is right: these constants are keyed by producing framework, not by the value's namespace (`SCOPE_STRANDS = "strands.telemetry.tracer"`), and the `_OTEL`/`_SMOLAGENTS` suffixes exist only to disambiguate multiple instrumentation libraries for one framework. There's a single ADK-emitted scope, so `_GCP` would disambiguate nothing and would read as "the GCP flavour of ADK".

Worth knowing anyway: ADK also emits under `google.adk.plugins.bigquery_agent_analytics` (`google/adk/plugins/bigquery_agent_analytics_plugin.py:85`) — an opt-in analytics plugin, not the `invoke_agent`/`call_llm`/`execute_tool` hierarchy, and the scope filter correctly drops it. No change needed; it just means "ADK spans" isn't one scope forever, so this constant is mildly version-sensitive if a future release moves the core tracer to a `google.adk.*` name.

</details>

<details><summary><strong>Closed items and the agreed follow-up</strong></summary>

Not re-raised, all settled earlier: latest-only parallel terminal tool-result selection; the raw `{"result": …}` scalar telemetry contract; thought filtering; multipart concatenation; `(merged tools)` exclusion; `error.type` with UNSET; scope-only mapper detection; the external-dict/`TracedHandler` contract; the global tool catalog; the `1e12` threshold (your call, and I agree — identical behaviour for real ~1.7e18 values); and short/float digit-string timestamps (theoretical, no producer emits them).

**Agreed follow-up, deliberately out of scope here:** `TraceExtractor` only picking the first `AgentInvocationSpan` per `Trace`. `liramon2` reproduced the same low-score symptom with the LangChain OTel mapper, so it's framework-agnostic rather than ADK-specific, and you both landed on fixing it separately — which I think is right: the per-agent `Trace` split here is the pragmatic workaround, and reverting it to wait on the extractor fix would just ship known-bad multi-agent scores meanwhile. Worth filing that follow-up before this merges so the workaround doesn't quietly become the convention; the shape is roughly "attribute per agent span instead of per trace" in `extractors/trace_extractor.py:87-133`.

Routing: triage GO on a rebase-plus-fixture delta → one independent correctness/test pass with mutation testing; API, LLM-context and docs skipped as not applicable. As ever, solid work — the merge call is yours.

</details>

Issue #299, opened 2026-07-06
[FEATURE] Skill-equipped agent evaluation: skill selection accuracy and skill instruction following


sangminwoo:
### Problem Statement

An agent *skill* is a small instruction file (often `SKILL.md`) that teaches an agent a domain-specific capability. The harness shows the agent each skill's name and description, and the agent decides whether to load the full skill and follow it.

Strands Evals can score whether an agent picks the right *tool* (`ToolSelectionAccuracyEvaluator`) but has no equivalent for skills. When an agent runs with skills available, there is no way to answer, from the trajectory: (1) did the agent select an appropriate skill (or correctly select none when nothing fits)? (2) did it follow the skill it selected? These are agent-capability questions parallel to tool selection, and like tool selection they assume the offered skills are well written.

### Proposed Solution

Add two trajectory-based judge evaluators plus a shared helper they call. Each evaluator subclasses the existing `Evaluator` (no base-class change) and emits `EvaluationOutput` (`score`, `test_pass`, `reason`, `label`).

1. **Skill selection accuracy.** A judge reads the task, the offered skills (name + description), and the trajectory, and scores whether the selection was appropriate: a fitting skill is good, an irrelevant one is bad, none is good only when no offered skill fits. One output per run, like `OutputEvaluator`.
2. **Skill instruction following.** Given the selected skill's `SKILL.md` and the trajectory, a judge marks each step `covered` / `partial` / `skipped` and reports a coverage score. One output per invoked skill; not-applicable when none was selected.
3. **`extract_selected_skills` helper.** Being *offered* (listed in an `<available_skills>` block) is not use; only loading the body counts. The helper scans the trajectory for the per-harness invocation signal:

   | Harness | Selection signal |
   |---|---|
   | Claude Code | `tool_use` `name == "Skill"` |
   | Strands | `toolUse` `name == "skills"` (or OTEL span `gen_ai.tool.name == "skills"`) |
   | Codex | `function_call` `name == "skills.read"` |
   | LangChain / LangGraph | `tool_calls` entry `name == "load_skill"` (documented pattern) |
   | Files on disk | generic file-read / command that loads the `SKILL.md` |

The offered-skill list rides on `Case.metadata` for now (it is not recorded on the trace the way `available_tools` is; aligning them is a useful upstream follow-up).

[...]

PR #330, opened 2026-07-28
feat: add skill-level evaluators for skill-equipped agents

sangminwoo:
## Description

A skill is an instruction file (usually `SKILL.md`) that the harness offers to an agent at runtime; the agent decides which, if any, to load. This adds evaluation for both halves of that behavior: which skill the agent selected, and whether it then followed the skill's instructions.

**Parsing helpers (`extractors/skills.py`)**
`parse_available_skills` and `extract_selected_skills` recover the skill signals from a trajectory. These are standalone helpers rather than a `TraceExtractor` level because skills have no structured representation in the trace schema: the available set arrives as harness-injected prompt text rather than a span field, and a skill load surfaces as an ordinary tool call. A load is detected either by a reserved skill-tool name plus its skill-name argument, or by a read of a known `SKILL.md` path. Both helpers accept a `Session` or a raw message list, so harnesses without a `Session` mapper are supported.

**Evaluators**
- `SkillSelectionAccuracyEvaluator` judges whether each invoked skill was an appropriate pick, one output per invoked skill, mirroring `ToolSelectionAccuracyEvaluator`. When no skill was invoked it judges the abstention itself.
- `SkillInstructionFollowingEvaluator` rates how fully each invoked skill's steps were followed on the framework's five-point scale, grounded in per-step covered/partial/skipped evidence.
- `SkillInvoked` is a deterministic presence check, the skill analogue of `ToolCalled`.

Prompts live in versioned template subpackages, following the existing convention. Both judges read the trajectory only, so any harness that emits one works: `actual_trajectory` accepts a `Session` or a raw message list.

**Two supporting fixes**
Both session mappers dropped the system prompt when building `AgentInvocationSpan`, so anything reading `AgentInvocationSpan.system_prompt` saw `None`. This is a bug independent of skills; it surfaced here because the harness advertises its skill catalog in the system prompt. Adds `_extract_system_prompt` handling both GenAI conventions (`gen_ai.system_instructions` and the legacy event form). One note on the OpenInference mapper: span-event message parsing is now unconditional rather than skipped when prompt and response were already recovered from smolagents attributes, because the system prompt lives in those messages. `_get_messages_from_span_events` memoizes per `span_id`, so there is no extra parsing cost.

`calculate_overall_score` now skips rows whose every output is labeled `not_applicable`, so an evaluator that cannot apply to a case does not deflate the aggregate with a placeholder 0.0. Behavior is unchanged for every existing evaluator, since none of them emit that label; two regression tests pin this.

Declares `pyyaml`, used to read `SKILL.md` frontmatter. It was already present transitively via `strands-agents`; this makes the direct dependency explicit using the same version bound.

## Related Issues

#299

## Documentation PR

N/A (README and SKILL.md updated in this PR)

## Type of Change

New feature

## Testing

- [X] I ran `hatch run prepare`

## Checklist
- [X] I have read the CONTRIBUTING document
- [X] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works
- [X] My change is focused and reasonably small; I have split unrelated work into separate PRs
- [X] I have added any necessary tests that prove my fix is effective or my feature works
- [X] I have updated the documentation accordingly
- [X] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [X] My changes generate no new warnings
- [X] Any dependent changes have been merged and published

----


src/strands_evals/evaluators/skill_selection_accuracy_evaluator.py
```diff
@@ -0,0 +1,125 @@
...
+    @staticmethod
+    def _missing_trajectory_row() -> EvaluationOutput:
+        """A missing trajectory is absent data, not a correct abstention, so it is not scored."""
+        return EvaluationOutput(score=0.0, test_pass=False, reason="no trajectory provided", label="not_applicable")
+
+    def evaluate(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
+        if evaluation_case.actual_trajectory is None:
+            return [self._missing_trajectory_row()]
+        invoked = extract_selected_skills(evaluation_case.actual_trajectory)
+        evaluator_agent = Agent(model=self.model, system_prompt=self.system_prompt, callback_handler=None)
```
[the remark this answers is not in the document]
sangminwoo: Leaving this as is. An empty parsed catalog does not necessarily mean the trajectory is broken: Claude Code advertises skills without descriptions and omits the <available_skills> block, while invocation detection still works. Returning not_applicable would therefore discard judgeable selections from a major harness.

src/strands_evals/evaluators/skill_instruction_following_evaluator.py
```diff
@@ -0,0 +1,175 @@
...
+        self.model = model
+        # Drop not-applicable rows from the aggregate so no-skill runs don't deflate the mean.
+        self.aggregator = self._aggregate_dropping_na
+
+    @staticmethod
+    def _aggregate_dropping_na(outputs: list[EvaluationOutput]) -> tuple[float, bool, str]:
+        scored = [o for o in outputs if o.label != _NOT_APPLICABLE]
+        if not scored:
+            reason = " | ".join(o.reason for o in outputs if o.reason) or "not applicable"
+            return (0.0, True, reason)
```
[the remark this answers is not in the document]
sangminwoo: Fixed.

src/strands_evals/evaluators/skill_selection_accuracy_evaluator.py
jjbuck: I noticed that this is inconsistent with the way the flow works in other evaluators like the tool selection accuracy evaluator. 

In the other evaluators, e.g., https://github.com/strands-agents/evals/blob/main/src/strands_evals/evaluators/tool_selection_accuracy_evaluator.py#L50-L57, we instantiate a fresh copy of the agent inside the loop, i.e., for every tool we instantiate a fresh judge agent and invoke that fresh agent. 

Here, we instantiate the judge agent (on L98) and then reuse that agent for each and every skill. I think this means we'll accumulate turns from different skill evaluations in the judge conversation, right? If that's the right reading of the current implementation, we should change it to instantiate the judge agent inside the loop.

src/strands_evals/extractors/skills.py
```diff
@@ -0,0 +1,741 @@
...
+        return True
+    return any(
+        _result_failed(value[key])
+        for key in ("response", "result")
+        if key in value and _as_dict(value[key]) is not None
+    )
+
+
+def _body_from_result(result: Any) -> str | None:
+    """Return actual skill instructions, excluding errors and load acknowledgements."""
```
jjbuck: If the intention is to exclude errors, I think there's a gap here, specifically for Strands’ AgentSkills plugin.

AgentSkills registers configured SKILL.md files and exposes a `skills(skill_name=...)` tool that the model uses to load one. The argument is a lookup key, not a skill definition. For example, if `pdf-processing` is registered but the model requests `pdf-procesing` (one "s", i.e., the model made a typo), Then Strands  v1.42.0 returns the ordinary string "Skill 'pdf-procesing' not found...". 

The tool decorator marks ordinary string returns as `status="success"`, so _result_failed() returns false and this function treats the error message as the loaded skill body. `SkillInvoked` then reports that the skill was loaded, and `SkillInstructionFollowingEvaluator` may judge the error text as instructions. Could we add a regression fixture using this exact SDK response and either recognize it as a failed load rather than a skill body?

src/strands_evals/extractors/skills.py
```diff
@@ -0,0 +1,741 @@
+"""Skill trajectory parsing helpers.
```
jjbuck: Nit (or for follow-up): This is a pretty large monolithic file, and it mixes two different kinds of logic:
1. Understanding each harness’s raw trajectory format.
2. Deciding what counts as a skill invocation.

Those should be separate stages. Right now, Strands, Claude, Codex, Gemini, and other formats are handled through branches spread across the same functions. After parsing, their information is immediately reduced to `InvokedSkill(name, body)`. At that point we can no longer reliably distinguish:

  - A tool call that failed.
  - A successful load whose body was not captured.
  - Multiple loads of the same skill.
  - Loads made by different agents.
  - No load attempt at all.

This is why downstream code has to inspect response strings and make assumptions about message ordering.

I'd recommend that we first convert every supported trajectory format into one common representation. For example:
```
  SkillLoadEvent(
      name=...,
      status="attempted" | "loaded" | "failed",
      body=...,
      error=...,
      call_id=...,
      position=...,
      agent_id=...,
  )
```
  Each harness should have a small adapter whose only job is converting its native messages into these events. The shared extractor and evaluators should operate only on `SkillLoadEvent`, without knowing whether the source was Strands, Claude, or Codex.

A concrete structure could be:
```
  extractors/skills/
      models.py
      extractor.py
      adapters/
          session.py
          strands.py
          claude.py
          codex.py
```
Each adapter should be tested with an authoritative fixture from that harness. serialize_trajectory should move to evaluator prompt formatting because it is unrelated to trajectory extraction. This structure keeps provider-specific changes isolated and gives each evaluator enough information to apply its own definition of selected, invoked, or followed.

jjbuck (changes requested):

src/strands_evals/extractors/skills.py
sangminwoo: Good point. AgentSkills error strings are no longer treated as skill bodies, and named refusals are recorded as failed loads. `Skill 'x' activated (no instructions available)` remains a successful load with no body. I added regression coverage for both raw-list and `Session` trajectories using the exact SDK strings, plus a fixture captured from the real AgentSkills plugin. Matching is anchored to the complete status line so normal skill instructions are not filtered.

src/strands_evals/extractors/skills.py
sangminwoo: Thanks, I agree with the separation. I refactored `extractors/skills.py` into per-harness adapters plus a shared extractor, added the common `SkillLoadEvent(name, status, body, error, call_id, position, agent_id)` representation, and moved `serialize_trajectory` into prompt formatting. This preserves failures, uncaptured bodies, repeated loads, ordering, and agent attribution, while letting each evaluator apply its own semantics.
I also added fixtures captured from Strands AgentSkills, Google ADK, Claude Code, and Codex. They exposed and fixed previously unsupported Google GenAI and Codex rollout formats.

src/strands_evals/evaluators/skill_selection_accuracy_evaluator.py
sangminwoo: Thanks that's a good catch. Reusing the judge accumulated prior turns and resent the trajectory on top of that history. I fixed this by creating a fresh judge per skill while building the shared case context once. The tests now assert one judge instance and one prompt per skill.

poshinchen:
@strandly-the-agent review the changes and tell me if there's any gap / improvements, the harness-sdk is here: https://github.com/strands-agents/harness-sdk

src/strands_evals/types/evaluation_report.py
```diff
@@ -29,6 +29,34 @@ class EvaluationReport(BaseModel):
...
+
+        A case is dropped only when it produced at least one row and every row was
+        not-applicable. One judged row is enough to make the case's score a real number.
+
+        No rows at all is applicable, not vacuously not-applicable. An evaluator that returned
+        nothing did not decline to judge; it failed to, and `_default_aggregator` scores that
+        `(0.0, test_pass=False)`. Dropping the case would take that failure out of every average
+        and report the run as better than it was, so the 0.0 is kept and counted.
+        """
+        return not outputs or any(not output.not_applicable for output in outputs)
```
strandly-the-agent: 🔴 **`overall_score` reports a failed run as a clean sweep.** This drops a case whenever every row is `NOT_APPLICABLE` — *regardless of `test_pass`* — and both new evaluators emit exactly that shape for missing data (`skill_selection_accuracy_evaluator.py:144`, `skill_instruction_following_evaluator.py:120`: `_not_applicable_row("no trajectory provided", test_pass=False)`).

`actual_trajectory` defaults to `None`, so any run where one case fails to capture a trajectory hits this:

```python
ok   = [EvaluationOutput(score=1.0, test_pass=True,  label="Yes")]
fail = [EvaluationOutput(score=0.0, test_pass=False, label=NOT_APPLICABLE, reason="no trajectory provided")]
EvaluationReport.calculate_overall_score([0.0, 1.0], [fail, ok])   # -> 1.0
```

`overall_score` is a stored field on the report, so this isn't cosmetic: a library consumer and any score dashboard read the inflated number, the CLI prints `n/a` in the score column while `test_pass` says `False` on the same case, and `--fail-on threshold:X` won't catch it (`--fail-on any` still will, since it reads `test_passes`).

The label conflates *declined to judge* (`test_pass=True`, legitimately droppable) with *failed to judge* (`test_pass=False`, must count). Only the second is a lie:

```suggestion
        return not outputs or any(not output.not_applicable or not output.test_pass for output in outputs)
```

This is `copilot-pull-request-reviewer`'s objection, raised on 2026-07-29, 08-04 22:35 and 08-04 22:47 — I reproduced it on `789d2ff` and it is correct as stated. The docstring fix in `789d2ff` addressed the adjacent empty-list case, not this one.

src/strands_evals/extractors/skills/extractor.py
```diff
@@ -0,0 +1,629 @@
...
+            continue
+        if verb == "sed" and _SED_IN_PLACE.search(head):
+            continue  # `sed -i` rewrites the file
+        operands = head[len(parts[0]) :]
+        if path := _skill_path_from_text(operands):
+            return path
+    return None
+
+
+def _skill_read_path(tool_name: str, arguments: dict[str, Any]) -> str | None:
```
strandly-the-agent, line 90: 🔴 **Any read of a file named `SKILL.md` is recorded as an invoked skill**, with no check that such a skill was ever on offer. Run against this repo's own root `SKILL.md`:

```python
real = open("SKILL.md").read()
extract_selected_skills(bash_read("cat /path/to/evals/SKILL.md", real))
# [InvokedSkill(name='strands-evals', body='---\nname: strands-evals\n...', status='loaded', error=None)]
#   ^ 50,637 chars of this repo's contributor guide, offered to the judge as a skill's instructions
```

That's a plausible eval subject — a coding agent working on a repo that happens to contain a `SKILL.md` (this one; `harness-sdk`'s `.agents/skills/*/SKILL.md`; any repo using the convention). `SkillSelectionAccuracyEvaluator` then judges whether picking `strands-evals` was appropriate, and `SkillInstructionFollowingEvaluator` grades the agent against a contributor doc's steps — two judge calls and two wrong scores, silently.

When the file has no parsable frontmatter the name comes from the path (`_normalize.py:_skill_name_from_path`), which produces names that are wrong on their face:

```
cat /path/to/evals/SKILL.md  -> name='evals'
cat ./SKILL.md               -> name='.'
cat SKILL.md                 -> name='SKILL.md'
```

Two more shapes, same root cause:
- `cat skills/*/SKILL.md` (an agent surveying what's on offer) collapses to **one** event whose body is every file concatenated — measured: 1 event, 50,637-char body across two files, named from the first frontmatter block it finds. The other reads vanish, and the adherence judge grades several skills' steps as one skill's.
- A user's own tool named `skills` taking a `skill_name` argument becomes a phantom load with the tool's output as the "skill body" (`_patterns.py:_HARNESS_TOOLS`).

**Suggested fix:** gate the file-read path on the skill being in the parsed catalog — `parse_available_skills` is already computed by both evaluators, and the harness's `<available_skills>` block carries a `<location>` element for every filesystem skill (`agent_skills.py:_generate_skills_xml`), which is a ready-made path → name map. Failing that, at minimum require a real parent-directory component so `'.'` and `'SKILL.md'` can't become names, and emit one event per frontmatter block when a body contains several.

src/strands_evals/evaluators/prompt_templates/skill_instruction_following/skill_instruction_following_v0.py
```diff
@@ -0,0 +1,39 @@
...
+nearly all skipped is "Not Followed").
+
+## Guidelines
+- Judge only against the steps this skill prescribes, not generic task quality.
+- Ground each per-step judgment in specific evidence from the run.
+- A plan, unexecuted code snippet, or claim that a step will be done is not evidence
+  that an executable step was carried out. Require an action or result in the trajectory.
+- If the skill has no clearly enumerable steps, treat its core instructions as the steps.
+- If the skill body prescribes nothing at all (it is reference material rather than instructions),
+  return an empty step list rather than inventing steps.
```
strandly-the-agent, line 34: 🟡 (the one I would most want a live-judge check on) **A skill written as a decision tree may be graded on branches the run correctly never entered.** Line 9 asks the judge to "identify the prescribed steps in the SKILL.md body", lines 23-25 require the overall rating to "track the share of steps that were covered", and `SkillStepRating.status` is `Literal["covered", "partial", "skipped"]` — so a judge that *correctly notices* a branch didn't apply has **no honest status to record it with**. Its only options are "skipped" or silently omitting the step.

Take a routing skill of the shape most real `SKILL.md` files have:

```
- If the error mentions a timeout: check the dashboard; if latency is elevated, page on-call.
- If the error mentions an authentication failure: verify credentials; if expired, rotate them.
- Otherwise, escalate to a human.
Always end by posting a one-line status update.
```

A run that flawlessly executes the timeout branch has ~4 of 8 enumerated steps "covered". Per lines 23-25 that's "Partially Followed" → `0.5` → `test_pass=False` (threshold `0.75`). A correct run, graded fail. Tiered 🟡 rather than 🔴 deliberately: a competent judge may well omit untaken branches on its own, and I have not measured one.

Cheapest fix is one guideline line here:

```suggestion
  return an empty step list rather than inventing steps.
- A skill written as a decision tree prescribes one path per run, not every branch. Enumerate only
  the steps on the path this run's situation actually calls for; do not mark an alternative branch
  the run correctly did not enter as "skipped".
```

Worth considering the structural fix too — a fourth `status` of `"not_applicable"` — so the judge can say what it saw instead of being forced into a wrong bucket, and so `coverage` reported in `reason` stops understating correct conditional runs.

Evidence boundary, to be straight with you: the prompt lines and the `Literal` are code, but I did not invoke a live judge, so how often this bites in practice is inferred rather than measured.

src/strands_evals/extractors/skills/_patterns.py
```diff
@@ -0,0 +1,111 @@
...
+    # nothing wrong and the judge would be handed "Skill 'x' not found. Available skills: ..."
+    # as the instructions the agent was supposed to follow.
+    # Matched without DOTALL, like `_ACKNOWLEDGEMENT`: only a result that is nothing but the
+    # status line is discarded, so a multi-line body is never thrown away over its first line.
+    r"^(?:"
+    r"error\s*:\s*skill_name\s+is\s+required"
+    r"|skill\s+(?:'[^']*'|\"[^\"]*\"|[\w.-]+)\s+(?:was\s+|is\s+)?not\s+found"
+    r"|(?:unknown|unrecognized)\s+skill\b"
+    r"|no\s+such\s+skill\b"
+    r")\s*[:.]?\s*.*$",
```
strandly-the-agent: 🟡 **This discards short real skill bodies, and flips their status to `failed`.** The comment above (and its twin at line 103) says "only a result that is nothing but the status line is discarded, so a multi-line body is never thrown away over its first line". `fullmatch` without `DOTALL` does anchor `.` to one line — but the trailing `\s*` in `\s*[:.]?\s*.*$` **crosses newlines**, so a body of *status-looking first line + blank line + exactly one final line* matches whole:

```
'Skill activated.\nStep 1: do the thing.'                  -> discarded (2-line body lost)
'Skill activated.\n\nNow summarise the document.'          -> discarded
"Skill 'legacy-mode' not found.\n\nThat is expected; ..."   -> discarded AND status='failed'
'Skill activated.\n# Title'                                -> discarded
'Skill activated.\n\n\n\nStep one.\nStep two.'             -> kept (3+ lines survive)
```

Filesystem skills are immune, because the harness appends a `\n---\nLocation: …` tail that makes `fullmatch` impossible — but `skill.path` is `None` for `Skill(...)` instances (the plugin's own docstring example) and for **every** `Skill.from_url()` skill, so those bodies arrive bare. In the third case above the plugin *did* activate the skill — that string is a real body whose first line happens to be error-shaped, and the harness's `get_activated_skills()` lists it — while the extractor reports `status='failed'`, `body=None`, and the evaluator scores it `0.0`.

Anchor to a single line in both regexes:

```suggestion
    r")\s*[^\n]*\Z",
```

The comment is worth fixing in its own right: as written it tells the next maintainer this case is handled, so nobody looks.

src/strands_evals/evaluators/skill_instruction_following_evaluator.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+
+    FULLY_FOLLOWED = "Fully Followed"
+    MOSTLY_FOLLOWED = "Mostly Followed"
+    PARTIALLY_FOLLOWED = "Partially Followed"
+    MINIMALLY_FOLLOWED = "Minimally Followed"
+    NOT_FOLLOWED = "Not Followed"
+
+
+def _strip_frontmatter(body: str) -> str:
```
strandly-the-agent, line 34: 🟡 **The judge is told the harness's runtime metadata is part of the SKILL.md.** This strips a *leading* frontmatter block only, but the Strands plugin's tool result is `skill.instructions` (already frontmatter-free) **plus** a trailing `\n---\n` block (`agent_skills.py:_format_skill_response`). `_body_from_result` captures that verbatim, so `_build_prompt` labels this as "## SKILL.md instructions":

```
1. Identify the PDF file path from the user's request.
...
---
Allowed tools: file_read, shell
Compatibility: general-purpose
Location: /tmp/skills/pdf-processing/SKILL.md

Available resources:
  scripts/extract.py
  references/pdf-cheatsheet.md
```

This is the primary supported harness's main path — every filesystem-skill activation, every time. A judge that reads `Available resources: scripts/extract.py` as an instruction invents a step nobody authored; one that reads `Allowed tools:` as a constraint grades against a harness policy the skill author never wrote. Bounded (a good judge probably recognises the shape) but unforced noise in the one section the prompt insists is authoritative.

Suggest trimming the tail by its known keywords rather than a bare `---` split — `SKILL.md` prose can legitimately contain a horizontal rule, so splitting on `\n---\n` risks truncating real instructions.

strandly-the-agent (commented):
Reviewed `789d2ff` (vs merge-base `0d23f47`) at `poshinchen`'s request, cross-checked against `harness-sdk`'s real `AgentSkills` plugin. **This is strong work** — 21 commits of real self-correction, and most of what I threw at it held up. Verdict: **changes requested** — 2 blockers, one of which is Copilot's objection, raised three times, still open and still correct.

✅ **Verified, not asserted:** `pytest tests/` → **1872 passed**, `ruff check` + `mypy -p src` clean. Regenerated `strands_agent_skills.json` with the PR's own capture script against installed `strands-agents` 1.50.2 → **byte-identical** to the checked-in fixture (and that file matches `harness-sdk` HEAD). Drove the real plugin's 8 result branches through *both* the raw-message and mapper→`Session` paths: **they agree**. XML escaping round-trips, including a `</skill>` injection attempt, in Python *and* TypeScript. `from strands import Skill` is valid at the declared floor `strands-agents>=1.42.0`. ⚠️ The branch is **CONFLICTING** with `main`.

🔴 **Blockers**
1. **`overall_score` reports a failed run as `1.00`.** `is_applicable` drops a `NOT_APPLICABLE` row even when `test_pass=False` — exactly what both new evaluators emit for a missing trajectory. Copilot is right; repro inline.
2. **Any read of a file named `SKILL.md` is recorded as an invoked skill**, with no check it was ever on offer. Reading *this repo's own* root `SKILL.md` yields a phantom skill `strands-evals` whose "instructions" are 50k chars of contributor guide — two judge calls, two wrong scores.

🟡 **7 should-fix inline.** The one I'd look at first: a skill written as a decision tree may be graded on branches the run correctly never entered, and `SkillStepRating.status` gives the judge no way to say "not applicable to this run". Then a regex that swallows short skill bodies, the harness's metadata tail presented to the judge as SKILL.md instructions, mapper reach, and three mutation-verified test gaps. ⚪ 13 more in the appendix.

**Gaps & improvements — the thing you actually asked** (8 items, collapsed below). The two I'd act on: **LangChain/LangGraph** is in #299's harness table and shipped nowhere, and **`SkillLoadEvent.agent_id` is permanently `None`** because both mappers pass `metadata={}` while the trace carries `gen_ai.agent.name` — a one-line addition next to the `system_prompt` fix this PR already makes.

<details>
<summary><b>Gaps &amp; improvements</b> — what the harness can do vs. what these evaluators can measure</summary>

Read against `harness-sdk/strands-py/src/strands/vended_plugins/skills/{agent_skills,skill}.py`. Priority: 🅑 block · 🅕 follow-up issue · 🅧 dismissed.

1. 🅕 **LangChain / LangGraph adapter is absent.** #299's selection-signal table lists it alongside Claude Code, Strands and Codex; there is no adapter, fixture or mention. It *may* work incidentally via `adapters/gemini.py`'s generic `{"name":…, "args":…}` fallback, but that's untested and unclaimed. Verifying six harnesses beats claiming a seventh — just say so in the PR body rather than dropping it silently.
2. 🅕 **`agent_id` is structurally present and permanently `None`.** `SkillLoadEvent.agent_id` exists and `_agent_id_of` reads `span.metadata` (`extractor.py:164`), but every mapper constructs `ToolExecutionSpan(..., metadata={})`. A real two-agent run confirms the trace *does* carry the attribution (`gen_ai.agent.name` on each `invoke_agent` span) and the mapper drops it. Since this PR is the one teaching the mapper to carry `system_prompt`, carrying `gen_ai.agent.name` too is a one-line addition — is it in scope?
3. 🅕 **Multi-agent catalog mis-attribution.** `_available_from_session` returns the first `AgentInvocationSpan.system_prompt` that has a block, so in a swarm/graph where sub-agents carry *different* catalogs, a later sub-agent's pick is judged against the first sub-agent's catalog. Out of #299's stated scope; worth a tracking issue before someone points this at a swarm.
4. 🅕 **Missed activation — the agent that should have loaded a skill and didn't — is unmeasurable** except when the case author names the skill up front via `SkillInvoked`. `c14d88a6` removed abstention judging deliberately, and `_no_invocation_row` returns not-applicable with no judge call. Worth being explicit that this is the framework's existing posture, not a skills gap: `ToolSelectionAccuracyEvaluator` has no "should have called a tool" check either. A session-level evaluator, not a re-measurement.
5. 🅕 **No timing / over-activation signal.** `SkillLoadEvent.position` is dropped when folding to `InvokedSkill`, and the selection prompt explicitly tells the judge to score each invocation "on its own merits, not on whether the agent also loaded the other skills". So activating five skills to do one thing, or activating a skill *after* doing the work by hand, both score the same as a clean pick. `extract_skill_load_events` keeps `position`, so the raw material is there.
6. 🅕 **Skill resources are invisible to the extractor.** `_list_skill_resources` advertises `scripts/`, `references/`, `assets/`, and real skills routinely say "read `references/x.md`" — following a skill often *means* reading those. Grep for `references`/`scripts`/`assets` in `extractors/skills/` returns nothing. The judge can see such a read in the serialized trajectory, but nothing links "the skill advertised it" to "the agent read it", and the truncation above is most likely to eat exactly that region.
7. 🅕 **#299's "invocation rate falls out of the same helper, no judge"** didn't ship. Small, and the helper it needs already exists.
8. 🅧 **Dismissed after checking, so nobody re-opens them:** `allowed-tools` enforcement (the harness itself documents it as "not yet enforced", `skill.py:222` — nothing to measure); `strict` mode (raises at load time, before a trajectory exists); duplicate/shadowed skill names (harness dedups transparently; eval sees what the agent saw); URL-sourced skills with no filesystem path (extraction never depends on a path).

**One thing you might expect me to raise and I won't:** the harness keeps an exact ordered activation list in `agent.state["agent_skills"]["activated_skills"]` (`_track_activated_skill`), which looks like better ground truth than regexing prompt text. I checked — it isn't available: no mapper touches `agent.state`, `StateEquals` reads a field the *task author* populates, and the route would be Strands-only. The trajectory-based approach is the right generalist default here.

</details>

<details>
<summary><b>Questions</b> (design / scope — non-blocking except the first)</summary>

1. **Should the `NOT_APPLICABLE` mechanism be its own PR?** It changes what `overall_score` means for *every* evaluator, in a 6,100-line skills diff. It's a genuine prerequisite (without it, a correct "no skill needed" run averages in a placeholder `0.0`), so I'm not arguing for removal — I'm asking whether five new moving parts (a label value, a `not_applicable` property, a public `is_applicable`, an opt-in `_aggregate_dropping_na`, changed `calculate_overall_score`) deserve their own review and a decision record per `team/API_BAR_RAISING.md`. Note this repo has no `needs-api-review` label; the closest lever is `design`.
2. **Is the drop-NA behaviour label-driven or opt-in?** Right now it's both: `calculate_overall_score` drops NA *cases* for anyone who emits the literal label, while `_aggregate_dropping_na` drops NA *rows* only if the evaluator assigns `self.aggregator`. A third-party evaluator emitting the label gets half the semantics and no way to know which half.
3. **`EvaluationOutput(..., not_applicable=True)` is silently accepted and does nothing** (verified: pydantic drops the unknown kwarg, `label` stays `None`). That's the obvious path for an evaluator author. Would a real `EvaluationOutput.not_applicable(reason, ...)` classmethod be better — and would it let both evaluators drop their duplicated `_not_applicable_row` helpers?
4. **Does the extractor layer need to be public at all?** Six new exported names in `extractors`, when the three evaluators are what a user reaches for. `ToolCalled`'s equivalent `_check_session` is private and exports no helper. Under 1.x this asymmetry bites: promoting a private name later is free, withdrawing a public one is a major break — and it's the difference between "we can tighten the `SKILL.md` regex in a patch" being *yes* or *no*. `SkillLoadEvent` is a 7-field `NamedTuple`, so exporting it freezes field *order* too.
5. **Two words for one concept in one signature:** `extract_selected_skills(...) -> list[InvokedSkill]`. Worth `extract_invoked_skills`? And `SkillInvoked` (evaluator) vs `InvokedSkill` (data) differ only in word order — both public, genuinely confusable. Not exporting `InvokedSkill` resolves it for free.
6. **What should someone on an unrecognised harness do?** `_HARNESS_TOOLS` is a private dict inside the library, so the answer today is "open a PR". Defensible while each adapter wants a captured fixture — but the flip side is that a user's *own* tool named `skills` taking a `skill_name` argument becomes a phantom skill load with the tool's output as the "skill body" (same root cause as blocker 2; cross-checking against the parsed catalog fixes both).

</details>

<details>
<summary><b>Appendix — non-blocking (13)</b></summary>

**Docstring / comment drift** (load-bearing here, because the asides are what a future maintainer trusts)
- `models.py:41` — "the fullest body recovered wins" is not what `extractor.py:145` does: the *first* body is kept, and a later one only replaces it when it is longer **and contains the earlier text verbatim**. A garbled 72-char first capture beats a legitimate 141-char re-read.
- `extractor.py:1-6` — "harness-independent" overstates it: `_claude_body_after` and `if call.name == "Skill"` (`extractor.py:541`) are Claude-Code-specific and live here, not in `adapters/claude.py`. `adapters/claude.py:7-10` already says so; the module docstring and `_patterns.py:3-5`'s "adding a harness is a change to this module and its adapter, not to the traversal logic" should admit the exception.
- `_patterns.py:3` cites a "design-doc B.1 table" that isn't in the repo — unresolvable for an outside contributor.
- **PR body is stale in two places**: it describes `extractors/skills.py` (now a 10-module package), and says the selection evaluator "judges the abstention itself" when no skill was invoked — `c14d88a6` reversed exactly that, 7 commits and 9 days before head. README/`SKILL.md`/docstrings were all updated; only the body wasn't.

**Small correctness / hygiene**
- `_available_str` (`skill_selection_accuracy_evaluator.py:60`) renders one skill per line, so a `description: |` block scalar (which the harness passes through verbatim) splits across lines and the catalog stops being parseable by the judge; an empty description renders as a bare `- name: `, which reads like a load failure. `f"- {s.name}: {s.description}" if s.description else f"- {s.name}"` plus a `\n`→` ` fixes both.
- `_canonical_skill_key` folds `data-clean` and `data_clean` into one row. Both are loadable in the harness's default lenient mode, so two real skills can collapse to one.
- `cat …/SKILL.md.bak` is matched as a skill read (`_SKILL_PATH` has no trailing boundary), and `sed -ni 's/a/b/' …/SKILL.md` is reported as a read (`_SED_IN_PLACE` requires whitespace before `-i`).
- `_skill_name_from_body` calls `Skill.from_content`, which emits the harness's own `logger.warning` into eval output — one line per file-read event, and the "unquoted colon in description" case is common enough that the harness ships a dedicated fallback for it.
- The harness's `<available_skills>` block also carries `<location>` for every filesystem skill; `AvailableSkill` discards it. That element is a ready-made `SKILL.md` path → skill name map, i.e. most of the fix for blocker 2.
- An empty catalog is indistinguishable from an un-instrumented trajectory: the harness's literal `<available_skills>\nNo skills are currently available.\n</available_skills>` and "no block at all" both parse to `[]`.
- `_aggregate_dropping_na([])` returns `all_pass=True` where `_default_aggregator([])` returns `False`. Unreachable today; worth aligning while the file is open.
- `test_pass = normalized_score >= 0.75` diverges from the `>= 0.5` used by every other five-point judge (`coherence`, `faithfulness`, `response_relevance`). The inline rationale is sound — prescriptive steps deserve a higher bar — but it silently changes what "pass" means for anyone comparing pass rates across judges in one report, so it belongs in the class docstring, not just a comment. Same file: the docstring claims the scale "mirrors" the existing judges; the numeric mapping does, the labels are new (and better anchored) — worth rewording.
- **Fixture provenance**: `capture_skill_fixtures.py`'s docstring says the hand-written Gemini CLI / OpenAI Agents / OpenHands shapes "are marked as such" in `skill_fixtures.py` — they aren't. A reader of `skill_fixtures.py:91-171` has no signal that those three formats were never checked against a real harness, unlike the five that were.

</details>

<details>
<summary><b>How this review was produced</b></summary>

Seven independent fresh-context passes (routing, context-build, correctness, adversarial/differential-vs-harness, API/DevX, test-quality with mutation testing, LLM-context, docs-accuracy, issue-alignment), then de-duplicated and severity-gated on reachability. Repros were run against installed `strands-agents` 1.50.2 with the real `AgentSkills` plugin; no live model was invoked anywhere, so every judge-behaviour claim is grounded in the prompt text and the output schema rather than measured — I've flagged that where it matters. Then an independent audit pass re-verified every load-bearing claim and sent six back for correction, which is why two things you might expect to see aren't here: a pass reported that `SkillFollowingRating.coverage` feeds the score (it doesn't — it only appears in `reason`), and my first draft of blocker 2 illustrated itself with a repro that didn't reproduce. Both were fixed rather than published.

I'm an experimental agent — treat this as solid work for a human to approve, not a gate. Push back on anything that doesn't hold up.

</details>

src/strands_evals/evaluators/prompt_templates/trajectory_prompt_template.py
```diff
@@ -0,0 +1,41 @@
...
+        return "(no trajectory)"
+    if isinstance(trajectory, Session):
+        value: Any = trajectory.model_dump(mode="json")
+    else:
+        value = [item.model_dump(mode="json") if isinstance(item, BaseModel) else item for item in trajectory]
+    text = json.dumps(value, indent=2, default=str)
+    if max_chars <= 0 or len(text) <= max_chars:
+        return text
+    keep = max_chars // 2
+    return f"{text[:keep]}\n\n... [{len(text) - 2 * keep} characters omitted] ...\n\n{text[-keep:]}"
```
strandly-the-agent: 🟡 **When truncation fires, the judge is never told how to read the gap.** Neither `SYSTEM_PROMPT` mentions omission or truncation, while the instruction-following prompt simultaneously demands each step be grounded "in specific evidence from the run" and offers `"skipped"` as the status for "no evidence". So a step whose evidence sat in the dropped middle reads as skipped — a correct run marked down for being long. Measured on a realistic long trajectory: a tool call occurring only mid-run is present in the full text and absent from what the judge sees.

Cheap mitigation, in the note itself:

```suggestion
    return (
        f"{text[:keep]}\n\n"
        f"... [{len(text) - 2 * keep} characters omitted — evidence for a step may lie in this gap; "
        f"do not mark a step 'skipped' solely because no evidence for it is visible] ...\n\n"
        f"{text[-keep:]}"
    )
```

Worth pairing with a matching guideline line in both `_v0.py` prompts.

Separately, and only if it's cheap: `json.dumps(value, indent=2)` is roughly 2.3× more verbose per unit of information than the compact `Tool call: name(args)` / `Tool result: …` rendering the framework already uses in `evaluator.py:_format_tool_level_prompt`, so the effective 300k/300k budget is about half what the number suggests. Reusing that formatter would push truncation much further out.

src/strands_evals/mappers/openinference_session_mapper.py
```diff
@@ -546,13 +552,18 @@ def _convert_agent_invocation_span(self, span: dict, session_id: str) -> AgentIn
...
-        if not user_prompt or not agent_response:
-            input_messages, output_messages = self._get_messages_from_span_events(span)
-            if not user_prompt:
-                user_prompt = self._extract_user_prompt(input_messages, span)
-            if not agent_response:
-                agent_response = self._extract_agent_response(output_messages, span)
+        # LangGraph / ADOT: extract from structured messages. Parsed unconditionally
+        # (the result is cached) because the system prompt lives here even when the
+        # prompt and response were already recovered from smolagents attributes.
+        input_messages, output_messages = self._get_messages_from_span_events(span)
```
strandly-the-agent, line 610: 🟡 **Making this call unconditional has two effects the comment doesn't cover.**

1. **It can now drop a span that previously converted.** A smolagents AGENT span with malformed `span_events` (e.g. a non-dict `body`) used to skip this path entirely because the old guard was false; now it's parsed, raises, and is swallowed by the per-span handler at `:254` — losing the whole `AgentInvocationSpan`, not just the system prompt. `_extract_messages_from_span`'s `body.get(...)` wants an `isinstance(body, dict)` guard, matching `cloudwatch_parser.py:110`'s own convention.
2. **"the result is cached" is doing more work than it looks.** `_span_messages_cache` (`:685`) is keyed by `span_id` alone and reset per `map_to_session()` call, not per trace — and this method's input is documented as *grouped by `trace_id`*. Two spans from different traces sharing a `span_id` cross-contaminate; I confirmed both directions with constructed spans (trace B inheriting trace A's system prompt, and the reverse blanking trace A's real one). That cache pre-dates this PR and already affects `user_prompt`/`agent_response`, so it's **not this PR's bug** — but this line is what extends its reach to `system_prompt`, and `(trace_id, span_id)` as the key is a one-line fix while you're here.

Also minor: "so there is no extra parsing cost" (PR body) isn't quite right — it *is* new work when both fields were already recovered. Negligible in practice, worth not claiming.

The good news, verified: the guarded `if not user_prompt:` / `if not agent_response:` assignments below are genuinely safe — smolagents-recovered values are never overwritten, even when `span_events` carries conflicting content.

tests/strands_evals/types/test_evaluation_report.py
```diff
@@ -4,12 +4,113 @@
...
+        A case that produced nothing scores 0.0 and that 0.0 is real, so a corpus of one such case
+        and one perfect case averages 0.5. Were the empty case dropped the corpus would report
+        1.0, a clean sweep, with the failure invisible.
+        """
+        judged = [EvaluationOutput(score=1.0, test_pass=True, label="Yes")]
+
+        assert EvaluationReport.calculate_overall_score([0.0, 1.0], [[], judged]) == 0.5
+        # Contrast: a case that declined to judge is dropped, so the mean is the judged case alone.
+        declined = [EvaluationOutput(score=0.0, test_pass=True, label=NOT_APPLICABLE)]
+        assert EvaluationReport.calculate_overall_score([0.0, 1.0], [declined, judged]) == 1.0
```
strandly-the-agent, line 54: 🟡 **This contrast case is why the blocker above shipped three times.** `declined` uses `test_pass=True`, and so does *every* `NOT_APPLICABLE` output constructed anywhere in this PR's tests (`:15,27,53,65,98,104`, both skill-evaluator test files, `test_experiment.py`, `cli/test_run.py:763,796`). The one combination the shipped evaluators actually emit — `NOT_APPLICABLE` **with** `test_pass=False` — is never exercised, so nothing guards the branch that matters. One added assertion here pins it:

```python
        # A case that FAILED to judge is not droppable: the failure must reach the mean.
        failed = [EvaluationOutput(score=0.0, test_pass=False, label=NOT_APPLICABLE)]
        assert EvaluationReport.calculate_overall_score([0.0, 1.0], [failed, judged]) == 0.5
```

Two more gaps found by mutation testing, in case they're worth a guard while you're in here:

- **`extractor.py:145`** — deleting `and kept in candidate` (the containment rule that stops unrelated longer output displacing a real body) leaves **all 111 tests in `test_skills.py` green**. `test_longer_unrelated_output_does_not_displace_a_recovered_body` looks like it covers this but passes for an unrelated reason: its command uses `ls`, which isn't in `_READ_VERBS`, so no second event is ever created. It proves "`ls` isn't a read verb".
- **`strands_in_memory_session_mapper.py:208`** — nulling `direct = attributes.get("gen_ai.system_instructions")` leaves all 219 mapper tests green. That's a real wire shape, not a hypothetical: the SDK's tracer writes `gen_ai.system_instructions` as a span *attribute* whenever Langfuse or `gen_ai_span_attributes_only` is active, and unconditionally for bidi sessions. The one system-prompt test only exercises the event-attribute branch.

Worth saying plainly: the rest of this suite is strong. `_ACKNOWLEDGEMENT`/`_LOAD_ERROR` mutation-killed 9 and 12 tests respectively, `_shell_read_skill_path` has excellent parametrized coverage (redirects, `sed -i`, `tee`, `sudo`, SIGPIPE, `bash -lc`) and died immediately when I disabled redirect-stripping, and both judge evaluators mock `Agent` at the right boundary while still asserting real prompt content.

README.md
```diff
@@ -517,6 +518,35 @@ tool_parameter_evaluator = ToolParameterAccuracyEvaluator(
...
+
+# Did the agent follow the invoked skill's steps? Five-level rating, one result per
+# invoked skill, grounded in per-step covered/partial/skipped evidence.
+following_evaluator = SkillInstructionFollowingEvaluator()
+
+# Deterministic presence check, no model.
+invoked_check = SkillInvoked(skill_name="pdf-processing")
+```
+
+Both judges read the trajectory only, so any harness that emits one works: pass a `Session`
```
strandly-the-agent: 🟡 **"any harness that emits one works" isn't true, and the failure mode is silent.** Detection is an allowlist: `_HARNESS_TOOLS` (`_patterns.py:13-19`) plus the adapter registry — Strands, Claude Code, Codex, Gemini CLI, OpenHands (+ Google ADK shapes). Point this at anything else and every entry point returns empty rather than raising:

```python
parse_available_skills(unrecognised_messages)   # -> []
extract_selected_skills(unrecognised_messages)  # -> []
SkillInvoked(skill_name="pdf-processing").evaluate(case)
# [EvaluationOutput(score=0.0, test_pass=False, reason="skill 'pdf-processing' was not invoked")]
```

That last line comes from a trajectory that *does* contain a real skill load — a confident false `0.0`. A user gets systematically wrong scores with nothing flagging it.

```suggestion
Both judges read the trajectory only: pass a `Session` or a raw message list as
`actual_trajectory`. Skill signals are recognized for the Strands `AgentSkills` plugin,
Claude Code, Codex, Gemini CLI, OpenHands and Google ADK; a harness whose skill calls match
none of those yields empty results rather than an error, so check
`parse_available_skills(trajectory)` returns your catalog before trusting a score. The
signals are recovered by `parse_available_skills` and `extract_selected_skills` from
`strands_evals.extractors`.
```

While here: #299's harness table also lists LangChain/LangGraph, which doesn't ship an adapter or fixture. It may work incidentally through the generic `{"name":…, "args":…}` recognizer in `adapters/gemini.py`, but nothing tests that — worth either a fixture or an explicit "not yet" in the PR body.

sangminwoo:
@strandly-the-agent Thanks, I went through the blockers and should-fix items individually. Most are now addressed, and two I’m leaving out with rationale.

### Fixed
* is_applicable blocker: Confirmed. Applicability now distinguishes declining to judge (test_pass=True, droppable) from failing to judge (test_pass=False, retained), so a missing trajectory can no longer disappear from overall_score. CLI and telemetry use the same predicate.
* \s* regex bug: Confirmed. The regexes could swallow multiline skill bodies and even misclassify a body beginning with refusal-like text as a failed load. Both are now line-bounded.
* Harness metadata leakage: Reproduced with the AgentSkills fixture. The appended Location/resource metadata is now removed by field name rather than splitting on bare ---, which can be valid skill Markdown.
* Unnamed SKILL.md reads: Reads that would resolve to . or SKILL.md no longer produce skill events.
* Truncation guidance: The prompt now tells the judge that relevant evidence may fall inside the omitted middle section.
* Mutation-test gaps: Replaced both tests that were passing for the wrong reason and verified that the intended mutations now fail.

### Decision-tree grading
I also live-checked this against Sonnet 5 five times. The judge did mark untaken branches as skipped, as predicted, but still returned Fully Followed / 1.0 every time. So the proposed 0.5 scoring failure does not reproduce.

The actual issue was the diagnostic coverage in reason: it reported 0.60 for a correct conditional execution. A prompt guideline fixes that to 1.00 / 5/5 in 5/5 runs without changing the rating, so I did not add another status enum.

### Not taking
* Require file-read skills to appear in the catalog: I don’t think we can make this a hard requirement because Codex does not emit <available_skills> at all. Doing so would disable its file-read detection entirely. I kept the safe subset of the fix above. The residual false-positive case is real, but across 3,379 Tessl + SkillsBench trajectories from three agent models I found 0 false positives. I also documented that callers should inspect parse_available_skills before relying on the score. I think fully solving this needs a catalog-optional design rather than gating on catalog presence.
* Span-cache re-keying: Confirmed that this behavior predates and is untouched by this PR. I’d prefer to keep it as a separate change rather than expand this PR further.

jjbuck (approved):

poshinchen (approved):

PR #336, opened 2026-07-30
fix: scope tools to owning agent in multi-agent traces

liramon2:
## Description

`ToolSelectionAccuracyEvaluator` produces false negatives for multi-agent systems because all tool calls are evaluated against a single merged `available_tools` list. This PR fixes the scoping so each tool call is evaluated against only its owning agent's tools.

The fix resolves tool ownership by walking each tool span's parent chain to its nearest agent, rather than assuming one agent per trace. This also removes the per-agent trace splitting that was added to the ADK mapper in #326 — that splitting was a workaround for the extractor's single-agent assumption, and is no longer needed now that the extractor handles multi-agent traces directly. The mapper now emits one faithful Trace per OTel trace and re-parents spans to the nearest converted ancestor, or to the root agent span if no converted ancestors were found.

Additionally, when a framework doesn't emit tool metadata in its spans, the judge previously had no `available_tools` context and no basis for evaluating selection accuracy. A fallback prompt now tells the judge to evaluate based on the conversation context instead, handling the case where multi-agent scoping leaves a child agent's tool list empty.

Note: This PR does not integrate the Langchain OTel and OpenInference mappers because they currently collapse traces to one agent span. Fixing them would require a major overhaul and is deferred to a follow-up.

## Related Issue

Partially addresses #332

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works
- [x] My change is focused and reasonably small; I have split unrelated work into separate PRs
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #338, opened 2026-07-30
fix: include prior tool results in session_history for tool-level evaluators

### LINKED ISSUE #337: [BUG] Tool-level evaluators can't see prior tool results within the same trace
## Checks

- [x] I have searched issues and there are no duplicates of my issue

## Strands Version

Latest (SDK used by evaluator judge agents)

## Strands Evals Version

v1.0.3 (main branch)

## Python Version

3.10.0

## Operating System

macOS (darwin)

## Installation Method

git clone

## Description

`TraceExtractor._extract_tool_level` builds `session_history` at trace-level granularity rather than per-tool-span. All tool spans within the same trace receive an identical history snapshot — one that contains only the user prompt and results from *previous* traces, but none of the tool results from *sibling* tool calls earlier in the same trace.

This causes `ToolSelectionAccuracyEvaluator` and `ToolParameterAccuracyEvaluator` to produce false negatives: the judge model sees a tool call using values that appear "hallucinated" because it cannot see that a prior tool call already returned those values.

## Steps to Reproduce

Run the following script (requires AWS credentials for Bedrock):

```python
import math
from typing import Any

from strands import Agent, tool

from strands_evals.evaluators import ToolSelectionAccuracyEvaluator
from strands_evals.mappers import StrandsInMemorySessionMapper
from strands_evals.telemetry import StrandsEvalsTelemetry

telemetry = StrandsEvalsTelemetry().setup_in_memory_exporter()


@tool
def add_numbers(a: int, b: int) -> str:
    """Return the sum of two numbers.

    Args:
        a: First number.
        b: Second number.
    """
    return str(a + b)


@tool
def multiply_numbers(a: int, b: int) -> str:
    """Return the product of two numbers.

    Args:
        a: First number.
        b: Second number.
    """
    return str(a * b)


@tool
def divide_numbers(a: float, b: float) -> str:
    """Return a divided by b. Returns error message if b is zero.

    Args:
        a: Dividend.
        b: Divisor.
    """
    if b == 0:
        return "Error: division by zero"
    return str(a / b)


@tool
def square_root(n: float) -> str:
    """Return the square root of a non-negative number.

    Args:
        n: The number to take the square root of.
    """
    if n < 0:
        return "Error: cannot take square root of negative number"
    return str(math.sqrt(n))


ALL_TOOLS = [add_numbers, multiply_numbers, divide_numbers, square_root]


def main():
    mapper = StrandsInMemorySessionMapper()

    test_cases = [
        Case[str, str](
            name="sqrt-then-multiply",
            input="What is the square root of 1764, multiplied by 3?",
        ),
    ]

    def task_function(case: Case) -> dict[str, Any]:
        telemetry.in_memory_exporter.clear()

        agent = Agent(
            tools=ALL_TOOLS,
            system_prompt=(
                "You are a math assistant. Use tools for all calculations.\n"
                "For multi-step problems, call tools sequentially — do NOT "
                "compute intermediate values in your head."
            ),
            callback_handler=None,
        )

        result = agent(case.input)

        raw_spans = [
            span for span in telemetry.in_memory_exporter.get_finished_spans()
            if span.instrumentation_scope and span.instrumentation_scope.name != "strands-evals"
        ]

        session = mapper.map_to_session(raw_spans, session_id=case.session_id)
        return {"output": str(result), "trajectory": session}

    evaluators = [ToolSelectionAccuracyEvaluator()]
    experiment = Experiment[str, str](cases=test_cases, evaluators=evaluators)

    report = experiment.run_evaluations(task_function)
    report.run_display(include_actual_output=True)


if __name__ == "__main__":
    main()
```

## Expected Behavior

The agent calls `square_root({n: 1764})` → returns `42.0`, then calls `multiply_numbers({a: 42, b: 3})` → returns `126`. When the judge evaluates `multiply_numbers`, it should see:

```
## Previous conversation history
User: What is the square root of 1764, multiplied by 3?
Tool call: square_root({'n': 1764})
Tool result: 42.0

## Target tool-call to evaluate
Tool call: multiply_numbers({'a': 42, 'b': 3})
```

The judge sees that `a=42` came from the prior `square_root` result and scores the call as justified (1.0).

## Actual Behavior

When the judge evaluates `multiply_numbers`, it sees:

```
## Previous conversation history
User: What is the square root of 1764, multiplied by 3?

## Target tool-call to evaluate
Tool call: multiply_numbers({'a': 42, 'b': 3})
```

The prior `square_root` call and its result are **missing** from `session_history`. The judge concludes that `a=42` is hallucinated/fabricated and scores the call as unjustified (0.0).

## Possible Solution

Move the tool result accumulation inside the loop so each subsequent tool span sees prior results:

```python
for tool_span in tool_spans:
    evaluator_inputs.append(
        ToolLevelInput(
            ...
            session_history=list(session_history),
        )
    )
    # Accumulate this tool's execution so the next tool span sees it
    session_history.append(
        [ToolExecution(tool_call=tool_span.tool_call, tool_result=tool_span.tool_result)]
    )
```

This is a two-line change (one addition inside the loop, removal of the bulk append after the loop). It fixes the issue for all mappers (Strands, LangChain, OpenInference, ADK) since `TraceExtractor` is the shared code path.

### Affected Evaluators

- `ToolSelectionAccuracyEvaluator` — "is this tool call justified?"
- `ToolParameterAccuracyEvaluator` — "do parameters faithfully use prior context?"

Both use `_format_tool_level_prompt` which renders the (incomplete) `session_history`.

liramon2:
## Description

`TraceExtractor._extract_tool_level` builds `session_history` at trace-level granularity rather than per-tool-span. All tool spans within the same trace receive an identical history snapshot — one that contains only the user prompt and results from *previous* traces, but none of the tool results from sibling tool calls earlier in the same trace.

This causes `ToolSelectionAccuracyEvaluator` and `ToolParameterAccuracyEvaluator` to produce false negatives: the judge model sees a tool call using values that appear "hallucinated" because it cannot see that a prior tool call already returned those values.

## Related Issues

Fixes #337

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

Added `test_extract_tool_level_incremental_session_history` which verifies that:
- Tool span 0 (`square_root`) sees only the user prompt in `session_history`
- Tool span 1 (`multiply_numbers`) sees the user prompt **plus** the prior `square_root` call and its result (`42.0`)

The test fails without the fix (`assert 1 == 2` — second tool only sees user prompt) and passes with it.

- [x]  I ran `hatch run prepare`

## Checklist

### DIFF
--- a/src/strands_evals/evaluators/evaluator.py
+++ b/src/strands_evals/evaluators/evaluator.py
@@ -207,7 +207,10 @@ class Evaluator(Generic[InputT, OutputT]):
                     # Handle tool execution lists
                     for tool_exec in msg:
                         history_lines.append(f"Tool call: {tool_exec.tool_call.name}({tool_exec.tool_call.arguments})")
-                        history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
+                        if tool_exec.tool_result.error:
+                            history_lines.append(f"Tool result: ERROR - {tool_exec.tool_result.error}")
+                        else:
+                            history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
                 else:
                     text = msg.content[0].text if msg.content and hasattr(msg.content[0], "text") else ""
                     history_lines.append(f"{msg.role.value.capitalize()}: {text}")
@@ -232,7 +235,10 @@ class Evaluator(Generic[InputT, OutputT]):
                     # Handle tool execution lists
                     for tool_exec in msg:
                         history_lines.append(f"Tool call: {tool_exec.tool_call.name}({tool_exec.tool_call.arguments})")
-                        history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
+                        if tool_exec.tool_result.error:
+                            history_lines.append(f"Tool result: ERROR - {tool_exec.tool_result.error}")
+                        else:
+                            history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
                 else:
                     text = msg.content[0].text if msg.content and hasattr(msg.content[0], "text") else ""
                     history_lines.append(f"{msg.role.value.capitalize()}: {text}")
--- a/src/strands_evals/evaluators/evaluator.py
+++ b/src/strands_evals/evaluators/evaluator.py
@@ -207,10 +207,7 @@ class Evaluator(Generic[InputT, OutputT]):
                     # Handle tool execution lists
                     for tool_exec in msg:
                         history_lines.append(f"Tool call: {tool_exec.tool_call.name}({tool_exec.tool_call.arguments})")
-                        if tool_exec.tool_result.error:
-                            history_lines.append(f"Tool result: ERROR - {tool_exec.tool_result.error}")
-                        else:
-                            history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
+                        history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
                 else:
                     text = msg.content[0].text if msg.content and hasattr(msg.content[0], "text") else ""
                     history_lines.append(f"{msg.role.value.capitalize()}: {text}")
@@ -235,10 +232,7 @@ class Evaluator(Generic[InputT, OutputT]):
                     # Handle tool execution lists
                     for tool_exec in msg:
                         history_lines.append(f"Tool call: {tool_exec.tool_call.name}({tool_exec.tool_call.arguments})")
-                        if tool_exec.tool_result.error:
-                            history_lines.append(f"Tool result: ERROR - {tool_exec.tool_result.error}")
-                        else:
-                            history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
+                        history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
                 else:
                     text = msg.content[0].text if msg.content and hasattr(msg.content[0], "text") else ""
                     history_lines.append(f"{msg.role.value.capitalize()}: {text}")
--- a/src/strands_evals/extractors/trace_extractor.py
+++ b/src/strands_evals/extractors/trace_extractor.py
@@ -110,11 +110,10 @@ class TraceExtractor:
                     )
                 )
 
-            if tool_spans:
-                tool_executions = [
-                    ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result) for span in tool_spans
-                ]
-                session_history.append(tool_executions)
+                # Accumulate this tool's execution so subsequent tool spans see it
+                session_history.append(
+                    [ToolExecution(tool_call=tool_span.tool_call, tool_result=tool_span.tool_result)]
+                )
 
             if agent_span and agent_span.agent_response:
                 session_history.append(AssistantMessage(content=[TextContent(text=agent_span.agent_response)]))
--- a/src/strands_evals/extractors/trace_extractor.py
+++ b/src/strands_evals/extractors/trace_extractor.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import datetime, timezone
 
 from ..types.trace import (
     AgentInvocationSpan,
@@ -18,6 +19,13 @@ from ..types.trace import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+def _to_aware_utc(dt: datetime) -> datetime:
+    """Normalize a datetime to timezone-aware UTC."""
+    if dt.tzinfo is None:
+        return dt.replace(tzinfo=timezone.utc)
+    return dt
 
 
 class TraceExtractor:
@@ -108,7 +116,7 @@ class TraceExtractor:
                 prior_executions = [
                     ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result)
                     for span in tool_spans[:index]
-                    if span.span_info.end_time <= tool_span.span_info.start_time
+                    if _to_aware_utc(span.span_info.end_time) <= _to_aware_utc(tool_span.span_info.start_time)
                 ]
 
                 evaluator_inputs.append(
--- a/src/strands_evals/extractors/trace_extractor.py
+++ b/src/strands_evals/extractors/trace_extractor.py
@@ -85,7 +85,11 @@ class TraceExtractor:
         return evaluation_inputs
 
     def _extract_tool_level(self, session: Session) -> list[ToolLevelInput]:
-        """Extract tool-level inputs with session and tool context."""
+        """Extract tool-level inputs with session and tool context.
+
+        Note: spans are not scoped by parent_span_id, so nested agent-as-tool traces
+        may leak child-agent internals into the parent's history.
+        """
         evaluator_inputs: list[ToolLevelInput] = []
         session_history: list[UserMessage | list[ToolExecution] | AssistantMessage] = []
         available_tools: list[ToolConfig] = []
@@ -100,20 +104,27 @@ class TraceExtractor:
             if agent_span and agent_span.user_prompt:
                 session_history.append(UserMessage(content=[TextContent(text=agent_span.user_prompt)]))
 
-            for tool_span in tool_spans:
+            for index, tool_span in enumerate(tool_spans):
+                prior_executions = [
+                    ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result)
+                    for span in tool_spans[:index]
+                    if span.span_info.end_time <= tool_span.span_info.start_time
+                ]
+
                 evaluator_inputs.append(
                     ToolLevelInput(
                         span_info=tool_span.span_info,
                         available_tools=available_tools,
                         tool_execution_details=tool_span,
-                        session_history=list(session_history),
+                        session_history=list(session_history) + ([prior_executions] if prior_executions else []),
                     )
                 )
 
-                # Accumulate this tool's execution so subsequent tool spans see it
-                session_history.append(
-                    [ToolExecution(tool_call=tool_span.tool_call, tool_result=tool_span.tool_result)]
-                )
+            if tool_spans:
+                tool_executions = [
+                    ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result) for span in tool_spans
+                ]
+                session_history.append(tool_executions)
 
             if agent_span and agent_span.agent_response:
                 session_history.append(AssistantMessage(content=[TextContent(text=agent_span.agent_response)]))
--- a/src/strands_evals/extractors/trace_extractor.py
+++ b/src/strands_evals/extractors/trace_extractor.py
@@ -115,7 +115,10 @@ class TraceExtractor:
             tool_executions = [
                 ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result) for span in tool_spans
             ]
-            tool_end_times = [_to_aware_utc(span.span_info.end_time) for span in tool_spans]
+            tool_end_times = [
+                max(_to_aware_utc(span.span_info.end_time), _to_aware_utc(span.span_info.start_time))
+                for span in tool_spans
+            ]
 
             for index, tool_span in enumerate(tool_spans):
                 target_start = _to_aware_utc(tool_span.span_info.start_time)
--- a/src/strands_evals/extractors/trace_extractor.py
+++ b/src/strands_evals/extractors/trace_extractor.py
@@ -112,11 +112,19 @@ class TraceExtractor:
             if agent_span and agent_span.user_prompt:
                 session_history.append(UserMessage(content=[TextContent(text=agent_span.user_prompt)]))
 
+            tool_executions = [
+                ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result) for span in tool_spans
+            ]
+            tool_end_times = [_to_aware_utc(span.span_info.end_time) for span in tool_spans]
+
             for index, tool_span in enumerate(tool_spans):
+                target_start = _to_aware_utc(tool_span.span_info.start_time)
                 prior_executions = [
-                    ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result)
-                    for span in tool_spans[:index]
-                    if _to_aware_utc(span.span_info.end_time) <= _to_aware_utc(tool_span.span_info.start_time)
+                    tool_executions[position]
+                    for position in range(len(tool_spans))
+                    if position != index
+                    and tool_end_times[position] <= target_start
+                    and (tool_end_times[position] < target_start or position < index)
                 ]
 
                 evaluator_inputs.append(
@@ -129,9 +137,6 @@ class TraceExtractor:
                 )
 
             if tool_spans:
-                tool_executions = [
-                    ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result) for span in tool_spans
-                ]
                 session_history.append(tool_executions)
 
             if agent_span and agent_span.agent_response:
--- a/src/strands_evals/types/trace.py
+++ b/src/strands_evals/types/trace.py
@@ -167,7 +167,12 @@ class TraceLevelInput(BaseEvaluationInput):
 
 
 class ToolLevelInput(BaseEvaluationInput):
-    """Input for tool-level evaluators"""
+    """Input for tool-level evaluators.
+
+    `session_history` is the context preceding `tool_execution_details` and never includes the
+    call under evaluation. Unlike `TraceLevelInput`, where each `list[ToolExecution]` entry is one
+    turn's whole batch, here each entry holds exactly one execution.
+    """
 
     available_tools: list[ToolConfig]
     tool_execution_details: ToolExecutionSpan
--- a/src/strands_evals/types/trace.py
+++ b/src/strands_evals/types/trace.py
@@ -170,8 +170,8 @@ class ToolLevelInput(BaseEvaluationInput):
     """Input for tool-level evaluators.
 
     `session_history` is the context preceding `tool_execution_details` and never includes the
-    call under evaluation. Unlike `TraceLevelInput`, where each `list[ToolExecution]` entry is one
-    turn's whole batch, here each entry holds exactly one execution.
+    call under evaluation. Tool executions that precede it within the same trace are appended as a
+    single `list[ToolExecution]` entry, in span order.
     """
 
     available_tools: list[ToolConfig]
--- a/src/strands_evals/types/trace.py
+++ b/src/strands_evals/types/trace.py
@@ -169,9 +169,9 @@ class TraceLevelInput(BaseEvaluationInput):
 class ToolLevelInput(BaseEvaluationInput):
     """Input for tool-level evaluators.
 
-    `session_history` is the context preceding `tool_execution_details` and never includes the
-    call under evaluation. Tool executions that precede it within the same trace are appended as a
-    single `list[ToolExecution]` entry, in span order.
+    `session_history` never includes the call under evaluation. Within the same trace, only tool
+    executions that completed before this tool started (`end_time <= start_time`) are included,
+    with list position as tiebreaker for equal timestamps. Cross-trace history is unfiltered.
     """
 
     available_tools: list[ToolConfig]
--- a/tests/strands_evals/extractors/test_trace_extractor.py
+++ b/tests/strands_evals/extractors/test_trace_extractor.py
@@ -157,3 +157,41 @@ def test_extract_empty_session_tool_level():
 
     assert isinstance(result, list)
     assert len(result) == 0
+
+
+def test_extract_tool_level_incremental_session_history():
+    """Test that each tool span sees prior tool results in session_history."""
+    now = datetime.now()
+    agent_span = AgentInvocationSpan(
+        span_info=SpanInfo(session_id="test", span_id="a0", start_time=now, end_time=now),
+        user_prompt="What is the square root of 1764, multiplied by 3?",
+        agent_response="126",
+        available_tools=[ToolConfig(name="square_root"), ToolConfig(name="multiply_numbers")],
+    )
+    tool_span_1 = ToolExecutionSpan(
+        span_info=SpanInfo(session_id="test", span_id="t1", parent_span_id="a0", start_time=now, end_time=now),
+        tool_call=ToolCall(name="square_root", arguments={"n": 1764}),
+        tool_result=ToolResult(content="42.0"),
+    )
+    tool_span_2 = ToolExecutionSpan(
+        span_info=SpanInfo(session_id="test", span_id="t2", parent_span_id="a0", start_time=now, end_time=now),
+        tool_call=ToolCall(name="multiply_numbers", arguments={"a": 42, "b": 3}),
+        tool_result=ToolResult(content="126"),
+    )
+
+    trace = Trace(spans=[agent_span, tool_span_1, tool_span_2], trace_id="trace1", session_id="test")
+    session = Session(traces=[trace], session_id="test")
+
+    extractor = TraceExtractor(EvaluationLevel.TOOL_LEVEL)
+    result = extractor.extract(session)
+
+    assert len(result) == 2
+
+    # First tool: sees only the user prompt
+    assert len(result[0].session_history) == 1
+
+    # Second tool: sees user prompt + first tool's execution
+    assert len(result[1].session_history) == 2
+    assert isinstance(result[1].session_history[1], list)
+    assert result[1].session_history[1][0].tool_call.name == "square_root"
+    assert result[1].session_history[1][0].tool_result.content == "42.0"
--- a/tests/strands_evals/extractors/test_trace_extractor.py
+++ b/tests/strands_evals/extractors/test_trace_extractor.py
@@ -160,10 +160,19 @@ def test_extract_empty_session_tool_level():
 
 
 def test_extract_tool_level_incremental_session_history():
-    """Each tool span sees exactly its true predecessors in session_history, in order."""
-    now = datetime.now()
+    """Each tool span sees exactly its causally-completed predecessors in session_history.
+
+    Uses distinct timestamps so the causality filter is exercised:
+    - tool_1 (t=0→1) finishes before tool_2 starts → prior for tool_2 and tool_3
+    - tool_2 (t=1→4) is long-running and still in-flight when tool_3 starts at t=2 → NOT a prior for tool_3
+    - tool_3 (t=2→3) starts while tool_2 is still running
+    """
+    from datetime import timedelta, timezone
+
+    base = datetime(2024, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
+
     agent_span = AgentInvocationSpan(
-        span_info=SpanInfo(session_id="test", span_id="a0", start_time=now, end_time=now),
+        span_info=SpanInfo(session_id="test", span_id="a0", start_time=base, end_time=base + timedelta(seconds=10)),
         user_prompt="What is the square root of 1764, multiplied by 3, then add 1?",
         agent_response="127",
         available_tools=[
@@ -173,17 +182,36 @@ def test_extract_tool_level_incremental_session_history():
         ],
     )
     tool_span_1 = ToolExecutionSpan(
-        span_info=SpanInfo(session_id="test", span_id="t1", parent_span_id="a0", start_time=now, end_time=now),
+        span_info=SpanInfo(
+            session_id="test",
+            span_id="t1",
+            parent_span_id="a0",
+            start_time=base,
+            end_time=base + timedelta(seconds=1),
+        ),
         tool_call=ToolCall(name="square_root", arguments={"n": 1764}),
         tool_result=ToolResult(content="42.0"),
     )
+    # Long-running: finishes at t=4, AFTER tool_3 starts at t=2
     tool_span_2 = ToolExecutionSpan(
-        span_info=SpanInfo(session_id="test", span_id="t2", parent_span_id="a0", start_time=now, end_time=now),
+        span_info=SpanInfo(
+            session_id="test",
+            span_id="t2",
+            parent_span_id="a0",
+            start_time=base + timedelta(seconds=1),
+            end_time=base + timedelta(seconds=4),
+        ),
         tool_call=ToolCall(name="multiply_numbers", arguments={"a": 42, "b": 3}),
         tool_result=ToolResult(content="126"),
     )
     tool_span_3 = ToolExecutionSpan(
-        span_info=SpanInfo(session_id="test", span_id="t3", parent_span_id="a0", start_time=now, end_time=now),
+        span_info=SpanInfo(
+            session_id="test",
+            span_id="t3",
+            parent_span_id="a0",
+            start_time=base + timedelta(seconds=2),
+            end_time=base + timedelta(seconds=3),
+        ),
         tool_call=ToolCall(name="add_numbers", arguments={"a": 126, "b": 1}),
         tool_result=ToolResult(content="127"),
     )
@@ -202,7 +230,7 @@ def test_extract_tool_level_incremental_session_history():
     )
     assert result[0].tool_execution_details.tool_call.name == "square_root"
 
-    # Second tool: sees user prompt + first tool's execution
+    # Second tool: sees user prompt + first tool's execution (square_root finished at t=1 <= t=1)
     assert len(result[1].session_history) == 2, (
         f"second tool should see user prompt + 1 prior execution, got {len(result[1].session_history)} entries"
     )
@@ -210,11 +238,66 @@ def test_extract_tool_level_incremental_session_history():
     assert result[1].session_history[1][0].tool_call.name == "square_root"
     assert result[1].session_history[1][0].tool_result.content == "42.0"
 
-    # Third tool: sees user prompt + [square_root, multiply_numbers] in order
+    # Third tool: only square_root is a valid prior (multiply_numbers ends at t=4 > t=2 start)
     assert len(result[2].session_history) == 2, (
         f"expected 2 entries (user + prior tools), got {len(result[2].session_history)}"
     )
     prior_names = [entry.tool_call.name for entry in result[2].session_history[1]]
-    assert prior_names == ["square_root", "multiply_numbers"], (
-        f"expected [square_root, multiply_numbers] in order, got {prior_names}"
+    assert prior_names == ["square_root"], (
+        f"expected only [square_root] (multiply_numbers still running), got {prior_names}"
     )
+
+
+def test_extract_tool_level_mixed_tz_timestamps():
+    """The causality filter handles mixed naive/aware timestamps without raising TypeError."""
+    from datetime import timedelta, timezone
+
+    base_naive = datetime(2024, 1, 1, 12, 0, 0)  # no tzinfo
+    base_aware = datetime(2024, 1, 1, 12, 0, 1, tzinfo=timezone.utc)  # aware
+
+    agent_span = AgentInvocationSpan(
+        span_info=SpanInfo(
+            session_id="test", span_id="a0", start_time=base_naive, end_time=base_aware + timedelta(seconds=5)
+        ),
+        user_prompt="Mixed tz test",
+        agent_response="Done",
+        available_tools=[ToolConfig(name="tool_x"), ToolConfig(name="tool_y")],
+    )
+    # tool_x: naive timestamps
+    tool_x = ToolExecutionSpan(
+        span_info=SpanInfo(
+            session_id="test",
+            span_id="tx",
+            parent_span_id="a0",
+            start_time=base_naive,
+            end_time=base_naive + timedelta(seconds=1),
+        ),
+        tool_call=ToolCall(name="tool_x", arguments={}),
+        tool_result=ToolResult(content="x_result"),
+    )
+    # tool_y: aware timestamps — starts after tool_x ends
+    tool_y = ToolExecutionSpan(
+        span_info=SpanInfo(
+            session_id="test",
+            span_id="ty",
+            parent_span_id="a0",
+            start_time=base_aware,
+            end_time=base_aware + timedelta(seconds=1),
+        ),
+        tool_call=ToolCall(name="tool_y", arguments={}),
+        tool_result=ToolResult(content="y_result"),
+    )
+
+    trace = Trace(spans=[agent_span, tool_x, tool_y], trace_id="trace1", session_id="test")
+    session = Session(traces=[trace], session_id="test")
+
+    extractor = TraceExtractor(EvaluationLevel.TOOL_LEVEL)
+    # Should not raise TypeError
+    result = extractor.extract(session)
+
+    assert len(result) == 2
+    # tool_y should see tool_x as a prior (naive 12:00:01 <= aware 12:00:01 UTC)
+    assert result[1].tool_execution_details.tool_call.name == "tool_y"
+    assert len(result[1].session_history) == 2
+    prior_names = [e.tool_call.name for e in result[1].session_history[1]]
+    assert prior_names == ["tool_x"]
--- a/tests/strands_evals/extractors/test_trace_extractor.py
+++ b/tests/strands_evals/extractors/test_trace_extractor.py
@@ -160,13 +160,17 @@ def test_extract_empty_session_tool_level():
 
 
 def test_extract_tool_level_incremental_session_history():
-    """Test that each tool span sees prior tool results in session_history."""
+    """Each tool span sees exactly its true predecessors in session_history, in order."""
     now = datetime.now()
     agent_span = AgentInvocationSpan(
         span_info=SpanInfo(session_id="test", span_id="a0", start_time=now, end_time=now),
-        user_prompt="What is the square root of 1764, multiplied by 3?",
-        agent_response="126",
-        available_tools=[ToolConfig(name="square_root"), ToolConfig(name="multiply_numbers")],
+        user_prompt="What is the square root of 1764, multiplied by 3, then add 1?",
+        agent_response="127",
+        available_tools=[
+            ToolConfig(name="square_root"),
+            ToolConfig(name="multiply_numbers"),
+            ToolConfig(name="add_numbers"),
+        ],
     )
     tool_span_1 = ToolExecutionSpan(
         span_info=SpanInfo(session_id="test", span_id="t1", parent_span_id="a0", start_time=now, end_time=now),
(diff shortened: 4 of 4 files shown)


liramon2:
@strandly-the-agent review this PR. Consolidate findings into a single comment.

strandly-the-agent:
**TL;DR — changes requested.** The core fix is right and delivers exactly what #337 asked for. One new failure mode it introduces needs a decision before merge, plus one test gap.

**Verified on `d72732a`** (base `0bf1252`):
- ✅ `pytest`: extractors **42 passed** · evaluators **336 passed** · extractors+evaluators+mappers **628 passed** · `ruff check` and `mypy` clean on `trace_extractor.py`
- ✅ your test genuinely fails on `main` (`assert 1 == 2`) and passes here — the PR-body claim is accurate
- ✅ the rendered judge prompt for the 2nd tool call now matches #337's *Expected Behavior* block verbatim; on `main` it matches *Actual Behavior*
- ✅ no scope creep — the diff is the issue's proposed fix plus one test
- 🔴 in a nested agent-as-tool trace, a sub-agent's private tool result now lands in the **parent** agent's judge prompt (repro below, driven through the real `StrandsInMemorySessionMapper`); a 3-line fix is validated across 4 scenarios
- 🟡 the 2-tool fixture doesn't kill a realistic index mutant · 🟡 quadratic judge-prompt growth, measured 8.6× on the repo's own fixture

I'm `strandly-the-agent` (an AI agent) — this is work for a human to approve, not a gate.

<details>
<summary><b>🔴 Blocking — <code>session_history</code> order is list position, not causality (<code>trace_extractor.py:103-116</code>)</b></summary>

Two failure modes, one root cause: the new accumulation treats "earlier in `trace.spans`" as "happened before", and `_find_tool_execution_spans` (`trace_extractor.py:130-132`) never sorts or looks at the span tree.

**(a) Concurrent siblings.** `CloudWatchSessionMapper`, `OpenSearchSessionMapper` and `AdkOtelSessionMapper` sort by `start_time`; `StrandsInMemorySessionMapper`, `LangChainOtelSessionMapper` and `OpenInferenceSessionMapper` keep the order they receive, and an OTel in-memory exporter emits in span **end** order. A `long_search` (start 00:00.010, end 00:00.900) run concurrently with `quick_lookup` (start 00:00.020, end 00:00.100) is told `quick_lookup`'s result was already available.

**(b) Nested agent-as-tool — the sharper case, and sorting doesn't fix it.** A child agent's `invoke_agent` span *and its tool spans* live inside the parent's `execute_tool` span, in the same OTel trace, and the mapper groups strictly by `trace_id` (`strands_in_memory_session_mapper.py:109-110`). Because a wrapper span strictly contains its children, this isn't a race — it's guaranteed. In exporter order the parent's tool call sees its own child's internals; in `start_time` order the child's inner call sees its enclosing wrapper's result. I re-ran this myself with real spans (`TracerProvider` + `SimpleSpanProcessor` + `InMemorySpanExporter`, explicit ns timestamps) through the real mapper → extractor → `_format_tool_level_prompt`:

```
--- judge prompt for target=hr_assistant (PR head) ---
## Previous conversation history
User: bob pay band
Tool call: query_salary_db({'employee': 'bob'})
Tool result: CONFIDENTIAL SALARY DB: alice=$310k, bob=$95k, carol=$210k

## Target tool-call to evaluate
Tool call: hr_assistant({'query': 'bob pay band'})
```

On `main` that same input's history is just `User: bob pay band`. No wrong causal claim was possible before this diff (no tool saw any sibling), so this one is on the change.

**Validated fix** — intersect the list-order prefix with a "demonstrably completed first" check:

```python
for index, tool_span in enumerate(tool_spans):
    prior = [
        [ToolExecution(tool_call=ts.tool_call, tool_result=ts.tool_result)]
        for ts in tool_spans[:index]
        if ts.span_info.end_time <= tool_span.span_info.start_time
    ]
    evaluator_inputs.append(
        ToolLevelInput(
            span_info=tool_span.span_info,
            available_tools=available_tools,
            tool_execution_details=tool_span,
            session_history=list(session_history) + prior,
        )
    )

# after the loop, fold this trace's executions into the running history
session_history.extend(
    [ToolExecution(tool_call=ts.tool_call, tool_result=ts.tool_result)] for ts in tool_spans
)
```

Measured across four scenarios (`prefix` = this PR as-merged, `naive` = filter over *all* spans, `prefix_causal` = above):

| scenario | as-merged | naive filter | prefix + causality |
|---|---|---|---|
| your new test's fixture (all spans share one `now`) | 1 / 2 ✅ | **3 / 3 ❌** (self + later sibling) | 1 / 2 ✅ |
| sequential, via the real mapper | shows `square_root → 42.0` ✅ | ✅ | ✅ |
| concurrent siblings | shows the sibling ❌ | ✅ | ✅ |
| nested agent-as-tool | child internals → parent ❌ | ✅ | ✅ |

Worth noting the middle column: a filter on `end_time <= start_time` alone **breaks your own new test**, because every span in it shares one timestamp, so the target's own execution and the later sibling both pass. Keeping the list-order prefix is what makes it safe. `SpanInfo.start_time`/`end_time` are non-optional `datetime` (`types/trace.py:85-86`), and a mapper that derives both from one log timestamp (`cloudwatch_parser.py:232`) degrades this to today's behavior rather than a new failure mode.

Residual limitation, not asked for here: a *completed* sibling sub-agent's internals still cross agent scope (`send_email` still sees `query_salary_db`). Closing that needs `parent_span_id` scoping — `ToolExecutionSpan` carries it (`types/trace.py:84`) but the extractor never reads it. Also pre-existing and unchanged: `_find_agent_invocation_span` returns only the first agent span, so `available_tools` / `user_prompt` are already the wrong agent's in nested traces.

</details>

<details>
<summary><b>🟡 Should-fix (4)</b></summary>

**1. A realistic mutant survives the new test — `tests/strands_evals/extractors/test_trace_extractor.py:162-197`**

Mutation-tested with 6 mutants. Killed: self-contamination, duplicate append, reversed order. **Survived:** `append([ToolExecution(tool_call=tool_spans[0].tool_call, ...)])` instead of `tool_span` — a plausible copy-paste bug. With one predecessor slot, "append the current tool" and "always append tool #0" are indistinguishable; with three, the mutant produces `[UserMsg, [tool1], [tool1]]` for `tool3`. Suggested addition (ran green on this head, `ruff`-clean; fails on `main`):

```python
def test_extract_tool_level_three_tools_no_duplication_or_gaps():
    """Each tool span sees exactly its true predecessors, in order — a 2-tool fixture can't
    distinguish that from an off-by-one that always appends the first tool."""
    now = datetime.now()

    def tool_span(span_id, name, result):
        return ToolExecutionSpan(
            span_info=SpanInfo(
                session_id="test", span_id=span_id, parent_span_id="a0", start_time=now, end_time=now
            ),
            tool_call=ToolCall(name=name, arguments={"x": 1}),
            tool_result=ToolResult(content=result),
        )

    agent_span = AgentInvocationSpan(
        span_info=SpanInfo(session_id="test", span_id="a0", start_time=now, end_time=now),
        user_prompt="Do three things",
        agent_response="done",
        available_tools=[ToolConfig(name="tool1"), ToolConfig(name="tool2"), ToolConfig(name="tool3")],
    )
    spans = [agent_span, tool_span("t1", "tool1", "r1"), tool_span("t2", "tool2", "r2"), tool_span("t3", "tool3", "r3")]
    trace = Trace(spans=spans, trace_id="trace1", session_id="test")
    session = Session(traces=[trace], session_id="test")

    result = TraceExtractor(EvaluationLevel.TOOL_LEVEL).extract(session)

    assert len(result) == 3
    seen = [entry[0].tool_call.name for entry in result[2].session_history[1:]]
    assert seen == ["tool1", "tool2"], f"expected [tool1, tool2] in order, got {seen}"
```

**2. A prior tool *failure* renders as a silent success — `evaluators/evaluator.py:210` (and its twin at `:235`)**

The history loop renders `tool_result.content` and drops `tool_result.error` (`types/trace.py:47`), which the mappers populate separately (`strands_in_memory_session_mapper.py:397`, `cloudwatch_session_mapper.py:310-313`). `ToolResult(content="", error="AccessDeniedException")` renders as a bare `Tool result: ` — the judge reads "ran, returned nothing", not "failed". That's load-bearing for both rubrics (`tool_parameter_accuracy_v0.py:15` expects parameters to come from prior API results). The omission predates the PR, but this diff is what first routes tool-level history through it, so it's cheap to fix while the area is open:

```python
if tool_exec.tool_result.error:
    history_lines.append(f"Tool result: ERROR - {tool_exec.tool_result.error}")
else:
    history_lines.append(f"Tool result: {tool_exec.tool_result.content}")
```

**3. Quadratic judge-prompt growth, no truncation anywhere in the path**

Tool-level evaluation already makes one judge call per tool span; call *k* now carries *k−1* full prior results, so per-case volume goes quadratic. No length limit exists anywhere in `trace_extractor.py`, `evaluator.py`, or either tool-level evaluator. Measured on the repo's own fixture (`tests/strands_evals/mappers/fixtures/smolagents_live_spans.json`, 2 real tool calls, `main` behavior reproduced by monkeypatch):

```
main   per-call prompt chars=[245,  427]  TOTAL=672
pr338  per-call prompt chars=[245, 5547]  TOTAL=5792     (8.6x)
```

Extrapolating with that fixture's real 5,050-char `web_search` result: 20 tool calls ≈ 976K chars (~244K tokens) of judge input per case, vs ~4.8K on `main`. A single prompt starts to approach the default judge's window (`DEFAULT_BEDROCK_MODEL_ID = "global.anthropic.claude-sonnet-4-6"`, `evaluator.py:24`) at roughly 50 chatty calls. Caveat: the character counts are measured, the context-window failure is inferred — no live model call was made. Worth a bounded history (last N executions, or truncated result content) either here or as a fast-follow.

**4. `session_history` semantics changed silently, and now mean two things in one class**

The cut point moved from "before this trace" to "before this tool call", and the shape changed: each inner `list[ToolExecution]` used to be a whole turn's batch (`trace_extractor.py:62-70`, still true for `_extract_trace_level`) and is now always exactly one execution (`:113-116`). `5287ac1` (PR #77) wrote the former *"to match `_extract_tool_level` behavior"*, so that deliberate symmetry is now gone with nothing recording it. Nothing pins the old shape (no export, no docs, no prior test), so this is a behavioral fix rather than an API break — but a couple of docstring lines would save the next reader the investigation:

```python
class ToolLevelInput(BaseEvaluationInput):
    """Input for tool-level evaluators.

    `session_history` is the context preceding `tool_execution_details` and never includes the
    call under evaluation. Unlike `TraceLevelInput`, where each `list[ToolExecution]` entry is one
    turn's whole batch, here each entry holds exactly one execution.
    """
```

</details>

<details>
<summary>⚪ Nit</summary>

`test_trace_extractor.py:191-197` — bare `assert len(result[0].session_history) == 1` fails as `assert 1 == 2` with no context. An f-string message would make a CI failure diagnosable without opening the file. Matches the file's existing style, so purely optional.

</details>

<details>
<summary>Follow-ups — explicitly not blockers on this PR</summary>

- **Leaving `_extract_trace_level` / `_extract_session_level` alone is correct**, since they emit one unit per trace and therefore have no sibling-vs-self split. Given PR #77's symmetry, one line in the PR body saying so would help.
- `_extract_trace_level` re-appends the same `tool_spans` list once per `AgentInvocationSpan` in a trace (`trace_extractor.py:51-70`) → duplicate appends for a trace with >1 agent span. Pre-existing, separate issue.
- `tool_selection_accuracy_v0.py` never mentions tool calls, tool results, prior history or failures (grepped), so the newly-populated `## Previous conversation history` arrives with no rubric guidance — unlike `tool_parameter_accuracy_v0.py:15`. A single line such as *"If a prior tool call failed or returned an error, does proceeding with this action still make sense?"* would close the gap.
- Tool results are attacker-controlled in many agents, and they now reach the tool-level judge; a crafted result forges the formatter's own `##` headers (verified — it produced a fake `## Evaluation Guidelines (updated by system)` block). The same payload already reaches the trace-level judge on `main` through untouched code, so this is a pre-existing class whose blast radius widens here. Worth a tracking issue, not a change to this PR.
- The rendered format has no vocabulary for concurrency at all — even with perfect ordering, two parallel calls in one turn look identical to two sequential ones. That needs a turn/step id in `types/trace.py`, not a prompt tweak.
- Found in passing, unrelated: `CoherenceEvaluator` renders an empty user prompt for any turn that used tools, because `_extract_user_prompt` (`evaluator.py:150`) assumes `session_history[-1]` is the user message while `_extract_trace_level` appends the tool batch after it. Worth its own issue.
- For the record: `ToolLevelInput` isn't exported from `strands_evals.types` and appears nowhere in the docs, so no API-review label applies and no prompt-version bump is needed under `AGENTS.md:483`. Current labels look right.

</details>

**Questions**

- **(Blocking)** Do you want to fold the causal-order filter into this PR, or scope this fix explicitly to sequential single-agent traces (a docstring caveat) and track nested/concurrent as an immediate fast-follow? Either works for me — the thing I'd push back on is landing it silently, since the nested case is a demonstrated leak across agent boundaries rather than a hypothetical.
- **(Non-blocking)** Is tool-level evaluation of agent-as-tool traces meant to be supported today at all? `_find_agent_invocation_span` already attributes the wrong agent's prompt and tools in that shape on `main`, so knowing the intent would help calibrate how much the residual `parent_span_id` gap is worth.

liramon2:
@strandly-the-agent review this PR. Consolidate findings into a single comment. As for your questions, the limitations are already being tracked as part of: https://github.com/strands-agents/evals/pull/336

strandly-the-agent:
**Round 2 on `8f0abaa`** — the 🔴 causality blocker from round 1 is genuinely fixed and the 3-tool test landed. Two *new* issues appeared in the fix itself, one of them a hard crash, so: **changes requested** again.

**Verified** (base `main` = `0bf1252`, prev head `d72732a`):
- ✅ `pytest tests/strands_evals/extractors tests/strands_evals/evaluators` → **378 passed**; `ruff check`, `ruff format --check`, `mypy` clean on all 4 changed files
- ✅ round-1 #1: sequential (`square_root → 42.0`), concurrent siblings, nested agent-as-tool and multi-turn all now render correctly — re-run through real OTel spans → real `StrandsInMemorySessionMapper` → extractor → formatter
- ✅ round-1 #2: the 3-tool test kills the `tool_spans[0]` index mutant that survived the 2-tool fixture
- ✅ round-1 #5: `ToolLevelInput` docstring added
- 🔴 NEW: `extract()` raises `TypeError` on tz-mixed timestamps — reproducible through a real mapper, on data that worked on `main`
- 🔴 NEW: the error-rendering change drops the failure message — **that was my round-1 suggestion and it was wrong for this codebase; sorry**
- 🟡 the new causality filter has no real test coverage (two mutants survive all 378 tests) · ⚪ round-1 #4 (quadratic growth) still open, now measured
- On #336: I tried the merge — it does **not** close the residual leak. Evidence + a 2-line fix below.

*Reviewed by `strandly-the-agent`, an AI agent — a human should approve before this merges.*

<details>
<summary><b>🔴 A. New crash — <code>extract()</code> raises <code>TypeError</code> on tz-mixed timestamps</b></summary>

`trace_extractor.py:111` compares two `datetime`s with no normalisation, and `SessionMapper.parse_timestamp` (`mappers/session_mapper.py:92-111`) doesn't deliver what its own docstring promises ("Timezone-aware datetime in UTC"). Executed matrix:

```
parse_timestamp(missing / None)      -> tzinfo=UTC    (session_mapper.py:94)
parse_timestamp("")                  -> tzinfo=UTC    (:105 fallback)
parse_timestamp("2023-11-14T22:13:20") -> tzinfo=None  (:103, naive ISO, no offset)
parse_timestamp(naive datetime)      -> tzinfo=None    (:96, passthrough)
parse_timestamp("garbage")           -> tzinfo=UTC    (:105)
parse_timestamp(ns int / "...Z")     -> tzinfo=UTC
```

So one missing or unparseable timestamp in an otherwise naive export mixes the two, and the comparison blows up. Two repros:

```
1) Session with one naive + one aware tool span:
   extract() RAISED TypeError: can't compare offset-naive and offset-aware datetimes
     File ".../extractors/trace_extractor.py", line 111, in _extract_tool_level

2) Through the REAL OpenInferenceSessionMapper, using the repo's own
   tests/strands_evals/mappers/fixtures/openinference_live_spans.json re-exported with naive ISO
   timestamps and one EARLIER tool span missing end_time:
   trace live-trace-1: tool-finish-early(start tz=None, end tz=UTC), tool-finish(start tz=None, end tz=None)
   extract() RAISED TypeError: can't compare offset-naive and offset-aware datetimes
```

The crash condition is an earlier span's `end_time` being aware while a later span's `start_time` is naive (or vice versa). This is new: `main` and `d72732a` never compared timestamps, so it's a hard failure on data that previously worked — and it hits every `TOOL_LEVEL` consumer, including `extract_agent_tools_used_from_trace` (`tools_use_extractor.py:116`), which doesn't need the filter at all.

It will also look flaky: `SpanInfo`'s `field_serializer` normalises everything to aware UTC, so the crash reproduces on a freshly-mapped session and vanishes after a JSON round-trip through a cache/result store.

**Suggested fix:** a small `_precedes(earlier, later)` helper that normalises (treat naive as UTC) before comparing — or `try/except TypeError` falling back to list order. The root cause is upstream in `parse_timestamp`, worth fixing there too, but this PR is what makes it reachable.

</details>

<details>
<summary><b>🔴 B. New regression — the error rendering drops the actual failure message (my bad suggestion)</b></summary>

Round 1 suggested `f"Tool result: ERROR - {error}"`. That was wrong here: `ToolResult.error` is a **status string**, not a message. `strands_in_memory_session_mapper.py:354` sets `tool_error = None if tool_status == "success" else str(tool_status)` (i.e. `"error"`), while the real text stays in `content`. Same shape in `langchain_otel_session_mapper.py:379` and `openinference_session_mapper.py:510,805`. Through the real Strands mapper:

```
mapper produced ToolResult: content='Error: cannot take square root of negative number'  error='error'

head 8f0abaa ->  Tool result: ERROR - error                                       <- message DROPPED
d72732a      ->  Tool result: Error: cannot take square root of negative number
```

The same edit went into **both** loops — `evaluator.py:211` (tool level) and `:239` (trace level) — so it also degrades trace-level judges that were correct on `main`. Concrete victim: the chaos `failure_communication` judge, whose prompt says it is given "Tool outputs (Tool result:)" and must assess how failures were communicated (`evaluators/chaos/prompt_templates/failure_communication/failure_communication_v0.py:13`); it now sees `ERROR - error`.

And the reverse case bites too: CloudWatch/ADK do put a real message in `error` (`cloudwatch_session_mapper.py:312`, `adk_otel_session_mapper.py:411`), so with `error="Timeout after 30s"` + `content='partial rows: [1,2,3]'` the *content* is what's lost — same bug, other operand. Multi-line/JSON content goes the same way.

**Suggested fix, both loops** — keep both pieces:

```python
result = tool_exec.tool_result
if result.error:
    detail = f"{result.error}: {result.content}" if result.content else result.error
    history_lines.append(f"Tool result: ERROR - {detail}")
else:
    history_lines.append(f"Tool result: {result.content}")
```

No test would have caught it in either direction: all 10 existing `assert "Tool result: ..."` checks in `tests/strands_evals/evaluators/` (e.g. `test_faithfulness_evaluator.py:107`, `test_response_relevance_evaluator.py:117`, `test_harmfulness_evaluator.py:110`, `test_conciseness_evaluator.py:109`) build `ToolResult` without `error`, and `tests_integ/` never asserts on `Tool result:`.

</details>

<details>
<summary><b>🟡 C. The new causality filter has no real test — and its behavior is timestamp-shape dependent</b></summary>

Every span in the updated fixture shares one `now`, so `end_time <= start_time` is always `now <= now` — the comparison is only ever exercised in its degenerate branch. Two mutants, run against `tests/strands_evals/extractors` + `tests/strands_evals/evaluators`:

```
delete the `if` clause entirely (= d72732a semantics)  -> 378 passed   (SURVIVES)
invert `<=` to `>=`                                    -> 378 passed   (SURVIVES)
```

So the fix round 1 asked for is currently unverified by the suite. That gap also hides two shape-dependent behaviors:

- **A missing/empty `end_time`** falls back to aware `now()` (`session_mapper.py:94`) and can silently **drop a legitimate prior** rather than crash — executed: healthy export → `priors=['square_root']`; same export with `tool1.end_time` missing → `priors=[]`, i.e. the #337 fix quietly disappears for that trace. Cross-process clock skew does the same.
- **CloudWatch sets `start_time == end_time` for every span** (`cloudwatch_session_mapper.py:166-176`, `cloudwatch_parser.py:232`), so there the filter is a permanent no-op: #337 still works, but concurrent siblings still leak.

A fixture with **distinct** timestamps and one genuinely overlapping sibling would pin all of it — e.g. `test_extract_tool_level_excludes_concurrent_sibling_by_end_time` (a long-running tool that started first plus a short sibling that finished first; assert the sibling is excluded). That test kills both mutants above.

</details>

<details>
<summary><b>🟡 D. Docstring and implementation now disagree</b></summary>

`types/trace.py:174` says "each entry holds exactly one execution", but the implementation batches all priors into **one** entry (`trace_extractor.py:108-119`; verified: `send_email` gets `n=2` = user message + a 2-element list). That's my round-1 wording applied to a different grouping — the batched shape is fine (it renders identically and restores symmetry with `_extract_trace_level`), so the docstring is the thing to change:

```
    `session_history` is the context preceding `tool_execution_details` and never includes the
    call under evaluation. Tool executions that precede it within the same trace are appended as a
    single `list[ToolExecution]` entry, in span order.
```

Note `test_trace_extractor.py:217-219` hard-codes the batched shape, so today the test and the docstring assert opposite contracts — worth making them agree either way.

</details>

<details>
<summary><b>On PR #336 — I tried the merge; it doesn't close the residual leak</b></summary>

Taking the pointer seriously rather than on faith, here's what I found with `pr336` at `ba65fec`:

- `git merge pr336` into `pr338` **conflicts** in `_extract_tool_level` — both branches rewrite the same loop. Whoever lands second resolves it; the resolution is mechanical (`enumerate` + `prior_executions` alongside `scoped_tools`).
- #336 genuinely fixes the half round 1 flagged: `available_tools` scoped per owning agent, and the **root** agent's `user_prompt` instead of the first agent span's (`_find_root_agent`, `_resolve_tool_ownership`).
- But it doesn't touch `session_history`, so after a mechanical resolution the leak in the new `trace_extractor.py:90` docstring note is **still there** — verified on the merged tree: `send_email` still sees the child agent's `query_salary_db`.
- Closing it is 2 lines, using #336's own ownership map. Verified on the merged tree (`send_email` then sees only `hr_assistant`; extractor tests pass, ruff + mypy clean):

```python
owner = tool_to_agent.get(tool_span.span_info.span_id) if tool_to_agent else None
prior_executions = [
    ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result)
    for span in tool_spans[:index]
    if span.span_info.end_time <= tool_span.span_info.start_time
    and (not tool_to_agent or tool_to_agent.get(span.span_info.span_id) is owner)
]
```

Not suggesting either PR absorb the other — just that #336 as it stands won't retire the caveat this PR just documented, and whichever lands second is the natural place for that check.

</details>

<details>
<summary>⚪ Nits &amp; follow-ups</summary>

- **Round-1 #4 (quadratic growth) is still open and now measurable in CPU too:** `ToolExecution` objects are rebuilt per target, so a 200-tool trace goes `extract()` **2.24 ms (main) → 59.96 ms (8f0abaa)** with 19,900 `ToolExecution` objects in history. Building the trace's execution list once and slicing it restores O(N) construction for free; prompt volume itself is still unbounded.
- **Cross-trace priors are still unfiltered** — a later trace's tools are "priors" for an earlier trace's regardless of timing, because the bulk cross-trace entries are bare `ToolExecution` with no `span_info` to apply the predicate to. Identical on `main`/`d72732a`/`8f0abaa`, so this PR owns the asymmetry rather than the leak. Is that intentionally out of scope, or worth tracking alongside #336?
- **Priors are emitted in list order, which is mapper-dependent** (Strands = span-end order; ADK/OpenInference = start order), so the same logical trace can yield differently-ordered history across sources.
- **Pre-existing, separate issue:** `Evaluator._extract_user_prompt` (`evaluator.py:147-156`) returns `""` for every *trace-level* input whose trace used tools, because `_extract_trace_level` appends the tool batch after the user message (`trace_extractor.py:62-70`) — affects `CorrectnessEvaluator` and `CoherenceEvaluator`. Predates this PR; worth its own ticket.
- **Pre-existing:** `SessionMapper.parse_timestamp`'s docstring promises aware UTC but returns naive for two of its input shapes (`session_mapper.py:96,103`) — the root cause behind finding A.

</details>

liramon2:
@strandly-the-agent review this PR. Consolidate findings into a single comment. I am deferring the error-rendering change because it is unrelated to this one.

strandly-the-agent:
**Round 3 on `c0407b7` — ✅ approve with nits, mergeable as-is.** Everything round 2 asked for landed, and deferring the error-rendering change is the right call: it was never part of #337, and the revert is clean (`evaluator.py` is byte-identical to `main`). Not re-litigating it — it's a real gap, just its own ticket.

**Verified this round:**
- ✅ `pytest extractors+evaluators` → **379 passed**, `+mappers` → **629 passed**; `ruff check`, `ruff format --check`, `mypy` clean
- ✅ both round-2 tz-crash repros now pass (direct tz-mixed `Session`, and the real `OpenInferenceSessionMapper` naive-ISO export with a missing `end_time`)
- ✅ the mutants that survived at `8f0abaa` are now **caught**: deleting the causality clause, inverting `<=`, dropping the `_to_aware_utc` normalisation, shifting naive by +1h
- ✅ #337's acceptance prompt still renders verbatim; concurrent / nested-agent / multi-turn matrix all correct through real OTel spans → real mapper → formatter
- 🟡 one completeness gap left (not a regression, doesn't block): the `[:index]` prefix means the fix doesn't fire on mappers that keep payload order — validated 3-line patch below

*Reviewed by `strandly-the-agent`, an AI agent — a human should approve before merge.*

<details>
<summary><b>🟡 1. The <code>tool_spans[:index]</code> prefix makes the fix a no-op where <code>Trace.spans</code> isn't chronological</b></summary>

`trace_extractor.py:117` only considers spans **earlier in the list**. Three mappers sort chronologically (`adk_otel_session_mapper.py:218`, `cloudwatch_session_mapper.py:97`, `opensearch_session_mapper.py:49`); three don't — `openinference_session_mapper.py:275` and `langchain_otel_session_mapper.py:168` both `return Trace(spans=converted_spans, ...)` in raw payload order (Langfuse likewise). And the repo's own fixture is non-chronological: in `tests/strands_evals/mappers/fixtures/openinference_live_spans.json`, `tool-finish` (start `1700000003…`) is listed before `route_to_agent` (start `1700000001…`).

Repro through the real `OpenInferenceSessionMapper` — two strictly sequential tools, `get_weather` 1s→2s then `book_hotel` 3s→4s:

```
A chronological payload order: tool order = [('get_weather','21','22'), ('book_hotel','23','24')]
    head c0407b7: priors for book_hotel = ['get_weather']
B book_hotel listed first    : tool order = [('book_hotel','23','24'), ('get_weather','21','22')]
    head c0407b7: priors for book_hotel = []           <-- the #337 fix silently doesn't fire
    tie-broken  : priors for book_hotel = ['get_weather']
```

`get_weather` ended a full second before `book_hotel` started — the temporal clause accepts it, the positional prefix alone rejects it. `main` gave no priors either, so this is **not a regression** and not a blocker; the cost is that on those paths the PR's benefit doesn't materialise.

**Don't just drop the prefix** — I checked, and that's worse. With the coarse/zero-duration shape (`cloudwatch_session_mapper.py:175-176` sets `start_time == end_time == ts`) a target then self-includes and sees future siblings: prefix kept → `t1: priors=[]`, `t2: priors=['t1']`; prefix removed → `t1: priors=['t1','t2']`.

**Validated patch** — strictly-earlier-completed always counts regardless of list position; exact ties fall back to list order:

```python
target_start = _to_aware_utc(tool_span.span_info.start_time)
prior_executions = [
    ToolExecution(tool_call=span.tool_call, tool_result=span.tool_result)
    for position, span in enumerate(tool_spans)
    if position != index
    and _to_aware_utc(span.span_info.end_time) <= target_start
    and (_to_aware_utc(span.span_info.end_time) < target_start or position < index)
]
```

I applied it and re-ran everything: **379 passed**, ruff + mypy clean, the scenario matrix unchanged, the equal-timestamp CloudWatch shape byte-identical, and repro B fixed. (Alternative: sort `tool_spans` chronologically inside `_find_tool_execution_spans` and keep the prefix — but that also reorders `_extract_trace_level`'s batch, so it needs its own check.) Fold it in now or track it — your call.

</details>

<details>
<summary><b>🟡 2. The two new tests don't yet pin the semantics they advertise (2 surviving mutants)</b></summary>

Both run against the 379-test suite:

- **`target.start_time` → `target.end_time`** (`trace_extractor.py:119`) → `379 passed`, **survives**. That mutant inverts the rule's meaning: a prior finishing *during* the target would count, i.e. the judge is told the agent had a result it hadn't received. It's invisible because the fixture's `tool_2` (1→4s) **outlives** `tool_3` (2→3s), so both operands reject it — the test docstring at `test_trace_extractor.py:167` advertises "still in-flight when tool_3 starts", but nothing finishes *between* tool_3's start and end. Making `tool_2` end at 2.5s discriminates.
- **`_to_aware_utc` naive branch → `UTC-7`** → `379 passed`, **survives** (so do `naive := epoch` and `naive := dt.astimezone(utc)`). `test_extract_tool_level_mixed_tz_timestamps` only asserts a naive prior at 12:00:01 **is** included — satisfied by any mapping that lands ≤ the aware instant. Adding one naive long-runner (12:00:00→12:00:05) and asserting it is **not** a prior for the aware `tool_y` closes it.

Credit where due: that test does correctly pin the `<=` boundary and does catch the raw `TypeError` — it just needs the exclusion direction too.

</details>

<details>
<summary>⚪ Nits</summary>

- **O(N²) construction is unchanged** (round-1 item, re-measured): one trace with N tool spans → `extract()` N=50: 0.23 ms (`main`) vs 2.83 ms; N=200: 2.13 ms vs **53.17 ms**, 19,900 `ToolExecution` objects retained. The datetime work isn't the cost — it's N(N−1)/2 pydantic constructions; building each once and slicing shared references gives identical output ~10–20× faster (the aliasing objection doesn't apply, since the shallow `list(session_history)` copy already shares those objects). Still just a note.
- **Coarse timestamps turn a parallel batch into a fake sequential chain.** With `start_time == end_time` for every span, `<=` admits every earlier-in-list sibling — verified: `t1: priors=[]`, `t2: priors=['t1']` for two spans sharing one timestamp. That's the deliberate cost of the `<=` boundary (which a test now pins — don't change it to `<`), but it's the one direction where the filter isn't conservative, so it's worth a sentence in the `_extract_tool_level` docstring beside the existing `parent_span_id` caveat.

</details>

<details>
<summary>❓ Questions &amp; follow-ups (all non-blocking)</summary>

- Is "within a trace" the deliberate scope for this invariant? The cross-trace bulk append (`trace_extractor.py:131-135`) applies no causality filter, so a tool from an overlapping/concurrent trace can still appear as a prior. Identical on `main`/`d72732a`/`8f0abaa`/`c0407b7`, so nothing regressed — just asking whether it's intentional.
- Would normalising at the `SpanInfo` boundary (a `field_validator` on `start_time`/`end_time`) be preferable to `_to_aware_utc` as a third copy of the same UTC coercion? It would also cover the mapper below.
- **Pre-existing:** the same tz-mixing `TypeError` still exists in `adk_otel_session_mapper.py:218` (`converted_spans.sort(key=lambda s: s.span_info.start_time)`) and `:100` (`min(...)`) — reproducible on a naive-ISO ADK export with one span missing `start_time`. Mappers are byte-identical to `main` here, so it's not this PR's, but it's the same root cause (`SessionMapper.parse_timestamp` promises aware UTC and returns naive for two input shapes — `session_mapper.py:96,103`).
- **Own tickets:** the deferred `tool_result.error` rendering gap in `evaluator.py`, and `Evaluator._extract_user_prompt` returning `""` for any trace-level input whose trace used tools (`evaluator.py:147-156`).

</details>

liramon2:
@strandly-the-agent review this PR. Consolidate findings into a single comment. For the questions:
1. The scope is deliberate
2. The scope is too big for this PR

strandly-the-agent:
**Round 4 on `2664f4f` — ✅ approve.** Both scope answers accepted: cross-trace history stays unfiltered by design (not raising it again), and the `SpanInfo`-boundary normalisation is out of scope — the ADK mapper's tz crash stays a pre-existing follow-up, not yours here.

**Verified fixed this round:**
- ✅ the tie-broken rule landed — a non-chronological payload through the real `OpenInferenceSessionMapper` now yields the prior in **both** orderings (round-3's gap: it silently gave `[]` before)
- ✅ both mutants that survived at `c0407b7` are now caught: target operand `start_time`→`end_time`, and `_to_aware_utc` mapping naive to the wrong offset. The `2.5s` end time and the naive `tool_long` span are doing real work
- ✅ the O(N²) construction is hoisted: N=400 in one trace went **195 ms → 16.5 ms** (`main`: 5 ms)
- ✅ 379 passed (extractors+evaluators) / 732 with mappers+detectors; `ruff check`, `ruff format --check`, `mypy` clean; #337's acceptance prompt still renders verbatim; concurrent / nested-agent / multi-turn matrix unchanged

One new 🟡 below — a one-line clamp I'd land before merge if it's cheap, otherwise worth tracking, because it's the same bug class as #337. The tie-break rule was my suggestion, so I had an independent pass attack it specifically; that's how this surfaced.

*Reviewed by `strandly-the-agent`, an AI agent — a human still approves the merge.*

<details>
<summary><b>🟡 An inverted span (<code>end_time &lt; start_time</code>) becomes a prior of tools that started earlier</b></summary>

`trace_extractor.py:118` trusts `end_time`, so a span whose `end` precedes its own `start` satisfies `end <= target_start` against *everyone* — including tools dispatched before it.

Reachable without any malformed evals code: `OpenSearchSessionMapper._make_span_info` parses both timestamps with `_parse_time` (`opensearch_session_mapper.py:106-107`), which returns **epoch 1970** for empty/malformed input (`:147-155` — pinned by your own tests at `test_opensearch_session_mapper.py:418-430`). A record with a valid `start_time` and a truncated/garbled `end_time` therefore yields `start=2026, end=1970`. Repro through the real mapper:

```
  search_flights  start=2026-01-01T00:00:01+00:00 end=2026-01-01T00:00:02+00:00
  book_flight     start=2026-01-01T00:00:03+00:00 end=1970-01-01T00:00:00+00:00  <== INVERTED
  send_email      start=2026-01-01T00:00:05+00:00 end=2026-01-01T00:00:06+00:00

HEAD 2664f4f as shipped:
  target=search_flights  priors=['book_flight']   <-- a result from 2s in its future
  target=book_flight     priors=['search_flights']
  'BOOKED ref=XYZ' in search_flights' judge prompt: True
```

`search_flights` and `book_flight` end up as mutual priors, so two tools in one trace get contradictory histories. `main` said nothing here, so this is new — and "the judge is told a tool saw a result that didn't exist yet" is exactly what #337's fix is about. (`strands_in_memory_session_mapper.py:162` `end_time = span.end_time or 0` can produce the same shape; LangChain/OpenInference fall back to `now()` instead, which errs safe.)

**Fix — clamp the end time, verified:**
```python
tool_end_times = [
    max(_to_aware_utc(span.span_info.end_time), _to_aware_utc(span.span_info.start_time))
    for span in tool_spans
]
```
I applied it and re-ran everything: **732 passed**, ruff + format + mypy clean, scenario matrix unchanged, the non-chronological case still fixed, and `search_flights` back to `priors=[]` with no leak into its prompt.

</details>

<details>
<summary><b>⚪ The tie-break clause itself is untested (2 surviving mutants) — optional</b></summary>

`or position < index` is the novelty of this commit, and nothing exercises it, because no fixture uses equal timestamps any more:

```
drop the tie-break clause entirely      -> 379 passed  (survives)
`or position < index` -> `or True`      -> 379 passed  (survives)
```

That's the coarse/zero-duration shape a real mapper produces (`cloudwatch_session_mapper.py:166-176` sets `start_time == end_time == ts` for every span). This test kills both (passes at head, `ruff`-clean, matches the file's style):

```python
def test_extract_tool_level_zero_duration_spans_fall_back_to_span_order():
    """Coarse timestamps (start_time == end_time, e.g. CloudWatch log records) fall back to span order.

    A tool must never appear in its own history, and must never see a later sibling, even when every
    span carries the identical timestamp.
    """
    ts = datetime(2024, 1, 1, 12, 0, 0, tzinfo=timezone.utc)

    def tool_span(span_id, name, result):
        return ToolExecutionSpan(
            span_info=SpanInfo(session_id="test", span_id=span_id, parent_span_id="a0", start_time=ts, end_time=ts),
            tool_call=ToolCall(name=name, arguments={}),
            tool_result=ToolResult(content=result),
        )

    agent_span = AgentInvocationSpan(
        span_info=SpanInfo(session_id="test", span_id="a0", start_time=ts, end_time=ts),
        user_prompt="Weather in Paris, Tokyo and Lima?",
        agent_response="Done",
        available_tools=[ToolConfig(name="get_paris"), ToolConfig(name="get_tokyo"), ToolConfig(name="get_lima")],
    )
    spans = [
        agent_span,
        tool_span("t1", "get_paris", "18C"),
        tool_span("t2", "get_tokyo", "27C"),
        tool_span("t3", "get_lima", "21C"),
    ]
    trace = Trace(spans=spans, trace_id="trace1", session_id="test")
    session = Session(traces=[trace], session_id="test")

    result = TraceExtractor(EvaluationLevel.TOOL_LEVEL).extract(session)

    assert len(result) == 3
    for position, evaluator_input in enumerate(result):
        own_name = evaluator_input.tool_execution_details.tool_call.name
        priors = [e.tool_call.name for m in evaluator_input.session_history if isinstance(m, list) for e in m]
        expected = ["get_paris", "get_tokyo", "get_lima"][:position]
        assert priors == expected, f"{own_name} expected priors {expected}, got {priors}"
        assert own_name not in priors, f"{own_name} appeared in its own history: {priors}"
```

Side note, since it looks like dead code but isn't: `position != index` is redundant for well-formed spans (self can only pass the tie-break when `end < start`) — it's precisely the guard for the inverted-span case above. Keep it.

</details>

<details>
<summary>⚪ Optional polish &amp; follow-ups</summary>

- The rule itself is still documented nowhere. `types/trace.py:172-174` says "Tool executions that precede it within the same trace" — "precede" is now carrying the whole causality rule. Tightening it to *"…from the same trace that completed before it started are appended as a single `list[ToolExecution]` entry, in span order"* would make the invariant (and, per your answer, the deliberate within-a-trace scope) legible without another round.
- Verified sound, for the record: the rule is antisymmetric, transitive, self-excluding and order-independent for all well-formed spans (an exhaustive sweep over timestamp assignments × span orderings found order-dependence confined to exact `end == start` ties, which is what the tie-break defines). The shared `tool_executions` list is not mutated after it's appended, `extract()` is idempotent, and a JSON round-trip is stable.
- Still open by design/agreement, no action asked: prompt volume is uncapped (a 400-tool trace sends ~4.8M tokens of judge input across its calls vs ~16K on `main`) — a cap or a last-N window would be a reasonable follow-up; the deferred `tool_result.error` rendering gap; and the pre-existing tz `TypeError` in `adk_otel_session_mapper.py:100,218`. Happy to file issues for any of those if useful.

</details>

liramon2:
@strandly-the-agent review this PR. Consolidate findings into a single comment.

strandly-the-agent:
**Round 5 on `baa56ee` — ✅ approve-with-nits, mergeable as-is.** The clamp is correct and exhaustively verified; the docstring it shipped with isn't — that's the one thing worth fixing, and it's a one-line swap.

**Independently verified at this head** — five fresh-context passes (adversarial `advanced`, correctness, test-quality, docs-accuracy, LLM-context); triage skipped API/DevX, correctly, since the delta changes a docstring and no signature/export/type:
- ✅ the clamp (`trace_extractor.py:118-121`) makes "prior" a strict partial order: **0** invariant violations over the full 4-timestamp grid (65,536 cases) vs 184,320 antisymmetry + 61,440 transitivity violations before it. It's monotone — it can only ever *remove* a prior, never add one — and it doesn't rewrite `span_info`, which still carries the raw timestamps
- ✅ reachability of the shape it guards against confirmed by driving real mapper code in 5 of 6 mappers (`strands_in_memory:161-162` `span.end_time or 0` is the production path); CloudWatch is structurally immune (`start_time == end_time` always), so there the tie-break carries the weight
- ✅ no regression: every well-formed shape is byte-identical `2664f4f`→`baa56ee`, and all four repo fixtures render identically through their real mappers; 379 tests pass (extractors+evaluators), `ruff`/`mypy` clean on both touched files
- 🟡 **the new docstring's formula is wrong** (`types/trace.py:172-174`): `end_time <= start_time` reads as the raw field, but the code compares `max(end_time, start_time)` against the *target's* `start_time`. Executed counterexample below — the one behaviour this commit exists to define is the one the docstring gets wrong
- 🟡 **the clamp still has no test**: reverting it, neutering it, or defeating the tie-break all leave 379/379 green. Not 🔴 — the shipped behaviour is proven correct on this head, so the exposure is regression protection, not a live defect. Two tests below close it

I'm `strandly-the-agent`, an AI agent — this is my read, not a merge gate; a human approves.

<details>
<summary><b>🟡 Docstring: the executed counterexample, and the fix</b></summary>

Span `A` has a negative duration (starts 10:05, reports `end_time` 10:00); target `B` starts 10:02:

```
A.end_time (10:00) <= B.start_time (10:02) -> True    <- what the docstring's formula literally says
  actual: target=A  priors=['B']
  actual: target=B  priors=[]                          <- A is NOT a prior of B
=> the code compares max(10:00, 10:05) = 10:05 > 10:02, so A is excluded
```

So the comparison is against a *derived* value, not the raw field — and it's the candidate's value against the target's `start_time`, which the self-referential-looking formula doesn't convey. Replacement (ruff-clean, longest line 94 cols):

```python
    """Input for tool-level evaluators.

    `session_history` never includes the call under evaluation. Within the same trace, a
    tool execution is included only if `max(end_time, start_time)` is at or before this tool's
    `start_time` (the max guards against spans with a negative duration), with list position
    as tiebreaker for equal timestamps. Cross-trace history is unfiltered.
    """
```

</details>

<details>
<summary><b>🟡 Two tests that close the gap (verified: pass at head, each kills its mutants)</b></summary>

Both follow the file's existing convention of a local `datetime` import (as at `test_trace_extractor.py:164,248`), are `ruff check`/`ruff format` clean and ≤120 cols. With both added: **381 passed**.

```python
def test_extract_tool_level_clamps_inverted_end_time():
    """A tool with a corrupted (earlier-than-start) end_time is not treated as completing early."""
    from datetime import timedelta, timezone

    base = datetime(2024, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
    agent_span = AgentInvocationSpan(
        span_info=SpanInfo(session_id="test", span_id="a0", start_time=base, end_time=base + timedelta(seconds=20)),
        user_prompt="x",
        agent_response="y",
        available_tools=[ToolConfig(name="broken_tool"), ToolConfig(name="tool_at_5")],
    )
    # starts at t=10 but reports end_time=t=1 (negative duration / corrupted telemetry)
    broken_tool = ToolExecutionSpan(
        span_info=SpanInfo(
            session_id="test",
            span_id="tb",
            parent_span_id="a0",
            start_time=base + timedelta(seconds=10),
            end_time=base + timedelta(seconds=1),
        ),
        tool_call=ToolCall(name="broken_tool", arguments={}),
        tool_result=ToolResult(content="broken_result"),
    )
    tool_at_5 = ToolExecutionSpan(
        span_info=SpanInfo(
            session_id="test",
            span_id="t5",
            parent_span_id="a0",
            start_time=base + timedelta(seconds=5),
            end_time=base + timedelta(seconds=6),
        ),
        tool_call=ToolCall(name="tool_at_5", arguments={}),
        tool_result=ToolResult(content="5_result"),
    )
    trace = Trace(spans=[agent_span, broken_tool, tool_at_5], trace_id="trace1", session_id="test")
    result = TraceExtractor(EvaluationLevel.TOOL_LEVEL).extract(Session(traces=[trace], session_id="test"))

    tool_at_5_input = next(r for r in result if r.tool_execution_details.tool_call.name == "tool_at_5")
    priors = [e.tool_call.name for g in tool_at_5_input.session_history if isinstance(g, list) for e in g]
    assert "broken_tool" not in priors, f"broken_tool (starts t=10) must not be a prior of tool_at_5 (t=5): {priors}"


def test_extract_tool_level_equal_timestamps_break_ties_by_position():
    """Two zero-duration spans sharing a timestamp are ordered by list position, not excluded."""
    from datetime import timedelta, timezone

    base = datetime(2024, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
    agent_span = AgentInvocationSpan(
        span_info=SpanInfo(session_id="test", span_id="a0", start_time=base, end_time=base + timedelta(seconds=5)),
        user_prompt="x",
        agent_response="y",
        available_tools=[ToolConfig(name="tool_first"), ToolConfig(name="tool_second")],
    )
    tool_first = ToolExecutionSpan(
        span_info=SpanInfo(session_id="test", span_id="tf", parent_span_id="a0", start_time=base, end_time=base),
        tool_call=ToolCall(name="tool_first", arguments={}),
        tool_result=ToolResult(content="first_result"),
    )
    tool_second = ToolExecutionSpan(
        span_info=SpanInfo(session_id="test", span_id="ts", parent_span_id="a0", start_time=base, end_time=base),
        tool_call=ToolCall(name="tool_second", arguments={}),
        tool_result=ToolResult(content="second_result"),
    )
    trace = Trace(spans=[agent_span, tool_first, tool_second], trace_id="trace1", session_id="test")
    result = TraceExtractor(EvaluationLevel.TOOL_LEVEL).extract(Session(traces=[trace], session_id="test"))

    priors_by_tool = {
        r.tool_execution_details.tool_call.name: [
            e.tool_call.name for g in r.session_history if isinstance(g, list) for e in g
        ]
        for r in result
    }
    assert priors_by_tool == {"tool_first": [], "tool_second": ["tool_first"]}, (
        f"ties must resolve by list position only, got {priors_by_tool}"
    )
```

Mutation results (each mutant run against these two tests):

```
at head                                    2 passed
revert the clamp (back to raw end_time)    1 failed   <- inverted-end-time test
neuter the clamp to max(end, end)          1 failed   <- inverted-end-time test
max -> min                                 1 failed   <- inverted-end-time test
drop the tie-break clause                  1 failed   <- equal-timestamps test
`or position < index` -> `or True`         1 failed   <- equal-timestamps test
```

The second test has to assert on **both** tools' prior sets: with the tie-break dropped, `tool_second` still sees `tool_first`, and the mutant only shows up as `tool_first` wrongly seeing `tool_second`.

</details>

<details>
<summary>⚪ Nit, no action needed</summary>

`position != index` (`trace_extractor.py:128`) is now unreachable: after clamping, `max(end, start) >= start`, so a tool can never satisfy the tie-break against itself. Verified by removing it — 379/379 unchanged, and exhaustively identical output across the whole timestamp grid. It was load-bearing *before* the clamp (an inverted span could list itself), so it's harmless to keep as defensive code — just don't read it as still doing work.

</details>

mehtarac (approved):

Issue #328, opened 2026-07-27
[FEATURE] Add Google ADK session mapper


liramon2:
### Problem Statement

Teams building AI agents with Google ADK (Agent Development Kit) cannot use strands-evals to evaluate their agent traces. ADK produces OpenTelemetry spans under the `gcp.vertex.agent` instrumentation scope with a distinct attribute schema (`gen_ai.*` / `gcp.vertex.agent.*`) that no existing mapper understands. Without a dedicated mapper, ADK users must either manually transform their traces into Session format or forgo trace-based evaluation entirely.

### Use Case

- As a Google ADK user, I want to evaluate my agent with strands-evals without rewriting my trace collection code

liramon2:
Update 7/28:
- Implementation + unit tests are in PR
- Integration tests will be in upcoming PR

PR #339, opened 2026-07-31
feat: add integration tests for ADK mapper

liramon2:
## Description
This adds integration tests for the ADK mapper added in https://github.com/strands-agents/evals/pull/326.

## Related Issues

https://github.com/strands-agents/evals/issues/328

## Documentation PR

None needed

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,11 +61,11 @@ otel = ["opentelemetry-exporter-otlp-proto-http>=1.30.0,<2.0.0"]
 langchain = [
     "langchain>=0.3.0",
     "langchain-aws>=0.2.0",
-    "langgraph>=0.2.0,!=0.5.0",
+    "langgraph>=0.2.0",
     "openinference-instrumentation-langchain>=0.1.0",
     "opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0",
 ]
-adk = ["google-adk[extensions]>=2.0.0"]
+adk = ["google-adk>=2.0.0,<3.0.0", "litellm>=1.75.0"]
 
 [tool.ruff]
 line-length = 120
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ langchain = [
     "openinference-instrumentation-langchain>=0.1.0",
     "opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0",
 ]
+adk = ["google-adk>=2.0.0"]
 
 [tool.ruff]
 line-length = 120
@@ -73,7 +74,7 @@ include = ["src/**/*.py", "tests/**/*.py", "tests_integ/**/*.py"]
 [tool.hatch.envs.hatch-test]
 installer = "uv"
 extra-args = ["-n", "auto", "-vv"]
-features = ["otel", "langfuse", "langchain", "opensearch"]
+features = ["otel", "langfuse", "langchain", "opensearch", "adk"]
 dependencies = [
     "pytest>=8.0.0,<10.0.0",
     "pytest-cov>=7.0.0,<8.0.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,11 +61,11 @@ otel = ["opentelemetry-exporter-otlp-proto-http>=1.30.0,<2.0.0"]
 langchain = [
     "langchain>=0.3.0",
     "langchain-aws>=0.2.0",
-    "langgraph>=0.2.0",
+    "langgraph>=0.2.0,!=0.5.0",
     "openinference-instrumentation-langchain>=0.1.0",
     "opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0",
 ]
-adk = ["google-adk>=2.0.0"]
+adk = ["google-adk[extensions]>=2.0.0"]
 
 [tool.ruff]
 line-length = 120
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ langchain = [
     "openinference-instrumentation-langchain>=0.1.0",
     "opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0",
 ]
-adk = ["google-adk>=2.0.0,<3.0.0", "litellm>=1.84"]
+adk = ["google-adk>=2.0.0,<3.0.0"]
 
 [tool.ruff]
 line-length = 120
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ langchain = [
     "openinference-instrumentation-langchain>=0.1.0",
     "opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0",
 ]
-adk = ["google-adk>=2.0.0,<3.0.0", "litellm>=1.75.0"]
+adk = ["google-adk>=2.0.0,<3.0.0", "litellm>=1.84"]
 
 [tool.ruff]
 line-length = 120
--- a/tests_integ/test_adk_eval.py
+++ b/tests_integ/test_adk_eval.py
@@ -29,7 +29,7 @@ from strands_evals.evaluators import (
 )
 from strands_evals.mappers import ADKOtelSessionMapper, detect_otel_mapper, readable_spans_to_dicts
 from strands_evals.telemetry import StrandsEvalsTelemetry
-from strands_evals.types.trace import AgentInvocationSpan, ToolExecutionSpan
+from strands_evals.types.trace import AgentInvocationSpan, Session, ToolExecutionSpan
 
 # Uses Gemini 3 Flash via Google AI API (requires GOOGLE_API_KEY env var).
 DEFAULT_MODEL_ID = "gemini-3-flash-preview"
@@ -291,10 +291,7 @@ def test_adk_multi_agent_evaluation(telemetry, create_multi_agent_runner):
         ),
     ]
 
-    captured_session = None
-
     def task_function(case: Case) -> dict:
-        nonlocal captured_session
         telemetry.in_memory_exporter.clear()
         runner = create_multi_agent_runner()
         response = asyncio.run(_run_adk_agent(runner, case.input))
@@ -302,7 +299,6 @@ def test_adk_multi_agent_evaluation(telemetry, create_multi_agent_runner):
         spans = readable_spans_to_dicts(telemetry.in_memory_exporter.get_finished_spans())
         mapper = detect_otel_mapper(spans)
         session = mapper.map_to_session(spans, session_id=case.session_id)
-        captured_session = session
         return {"output": response, "trajectory": session}
 
     evaluators = [
@@ -321,21 +317,21 @@ def test_adk_multi_agent_evaluation(telemetry, create_multi_agent_runner):
     assert len(report.scores) == 6
     assert all(report.test_passes), f"Some evaluations failed: {report.reasons}"
 
-    # Validate the mapped session structure directly — this is what the live trace proves
-    # that synthetic unit tests cannot.
-    assert captured_session is not None
-    assert len(captured_session.traces) == 2, (
-        f"Multi-agent should produce 2 traces (coordinator + specialist), got {len(captured_session.traces)}"
+    # Deserialize the trajectory from the first report case back into a Session object.
+    session = Session.model_validate(report.cases[0]["actual_trajectory"])
+
+    assert len(session.traces) == 2, (
+        f"Multi-agent should produce 2 traces (coordinator + specialist), got {len(session.traces)}"
     )
 
     # Identify traces by agent name in their AgentInvocationSpan metadata
-    agent_spans = [s for t in captured_session.traces for s in t.spans if isinstance(s, AgentInvocationSpan)]
+    agent_spans = [s for t in session.traces for s in t.spans if isinstance(s, AgentInvocationSpan)]
     agent_names = {s.metadata.get("agent_name") for s in agent_spans}
     assert "coordinator" in agent_names, f"Expected coordinator agent, got {agent_names}"
     assert "weather_specialist" in agent_names, f"Expected weather_specialist agent, got {agent_names}"
 
     # Tool spans should only appear on the specialist trace, not double-counted on coordinator
-    for trace in captured_session.traces:
+    for trace in session.traces:
         trace_agent_spans = [s for s in trace.spans if isinstance(s, AgentInvocationSpan)]
         trace_tool_spans = [s for s in trace.spans if isinstance(s, ToolExecutionSpan)]
         trace_agent_names = {s.metadata.get("agent_name") for s in trace_agent_spans}
--- a/tests_integ/test_adk_eval.py
+++ b/tests_integ/test_adk_eval.py
@@ -29,7 +29,9 @@ from strands_evals.mappers import ADKOtelSessionMapper, detect_otel_mapper, read
 from strands_evals.telemetry import StrandsEvalsTelemetry
 from strands_evals.types.trace import AgentInvocationSpan, ToolExecutionSpan
 
-DEFAULT_MODEL = LiteLlm(model="bedrock/us.anthropic.claude-sonnet-4-6", temperature=0)
+DEFAULT_MODEL_ID = "us.anthropic.claude-sonnet-4-20250514-v1:0"
+
+DEFAULT_MODEL = LiteLlm(model=f"bedrock/{DEFAULT_MODEL_ID}", temperature=0)
 
 
 # =============================================================================
--- a/tests_integ/test_adk_eval.py
+++ b/tests_integ/test_adk_eval.py
@@ -2,16 +2,18 @@
 
 Requirements:
     pip install strands-agents-evals[adk]
+    Set GOOGLE_API_KEY environment variable with a valid Google AI API key.
 
 Run with: pytest tests_integ/test_adk_eval.py -v
 """
 
 import asyncio
+import os
 import uuid
 
 import pytest
 from google.adk.agents import Agent
-from google.adk.models.lite_llm import LiteLlm
+from google.adk.models.google_llm import Gemini
 from google.adk.runners import InMemoryRunner
 from google.genai import types
 
@@ -29,9 +31,9 @@ from strands_evals.mappers import ADKOtelSessionMapper, detect_otel_mapper, read
 from strands_evals.telemetry import StrandsEvalsTelemetry
 from strands_evals.types.trace import AgentInvocationSpan, ToolExecutionSpan
 
-DEFAULT_MODEL_ID = "us.anthropic.claude-sonnet-4-20250514-v1:0"
-
-DEFAULT_MODEL = LiteLlm(model=f"bedrock/{DEFAULT_MODEL_ID}", temperature=0)
+# Uses Gemini 3 Flash via Google AI API (requires GOOGLE_API_KEY env var).
+DEFAULT_MODEL_ID = "gemini-3-flash-preview"
+DEFAULT_MODEL = Gemini(model=DEFAULT_MODEL_ID, google_api_key=os.environ.get("GOOGLE_API_KEY", ""))
 
 
 # =============================================================================
@@ -269,8 +271,7 @@ def test_adk_single_agent_multi_tool(telemetry, create_multi_tool_runner):
         assert len(tool_spans) >= 1, "Expected at least one tool execution span"
         for tool_span in tool_spans:
             assert tool_span.tool_call.tool_call_id is not None, (
-                f"tool_call_id should be populated (Bedrock emits gen_ai.tool.call.id), "
-                f"got None for tool '{tool_span.tool_call.name}'"
+                f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
             )
 
 
@@ -345,7 +346,7 @@ def test_adk_multi_agent_evaluation(telemetry, create_multi_agent_runner):
             assert tool_names <= {"transfer_to_agent"}, f"Unexpected tools on coordinator trace: {tool_names}"
         elif "weather_specialist" in trace_agent_names:
             assert len(trace_tool_spans) >= 1, "Specialist trace should have at least one tool span"
-            # Verify tool_call_id is populated (Bedrock emits gen_ai.tool.call.id)
+            # Verify tool_call_id is populated
             for tool_span in trace_tool_spans:
                 assert tool_span.tool_call.tool_call_id is not None, (
                     f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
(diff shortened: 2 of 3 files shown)


liramon2:
@strandly-the-agent Review this PR. Consolidate any findings in a single comment.

strandly-the-agent:
**One real blocker before this leaves draft: the new `adk` extra doesn't pull in `litellm`, so `tests_integ/test_adk_eval.py` fails at *collection*.** That's already why the integ job is red on this SHA. Once fixed, the suite genuinely passes — and it turned up two mapper bugs its own assertions don't catch.

✅ Reviewed `2883ce0` vs `cc0a391` · ✅ ran `pytest tests_integ/test_adk_eval.py -v` with `google-adk[extensions]` installed → **5 passed** (75s), and **5 passed** under `-n auto` (how CI runs it) · ✅ the multi-agent trace split verified **correct** against a real 11-span ADK trace: 2 mapped traces, one agent each, 0 duplicated/dropped tool spans, `available_tools` correctly scoped — that retires the open concern from #326 · ✅ `tool_call_id` confirmed populated end-to-end on this LiteLlm/Bedrock path · 🔴 the extra *as declared* fails at import · ⚪ cleared as non-issues: tracer-provider/xdist ordering, the `us.anthropic.claude-sonnet-4-6` model id (live-invokable), the mixed-tz sort `TypeError` at `adk_otel_session_mapper.py:100,218` (unreachable — real spans are all int-nanos → tz-aware UTC) · not verified: py3.10 specifically (I used py3.12), a real `hatch test` invocation, multi-turn/parallel-tool-call span shapes.

<details>
<summary><b>🔴 Blocker (1)</b></summary>

**`pyproject.toml:68` — `adk = ["google-adk>=2.0.0"]` can't satisfy `tests_integ/test_adk_eval.py:14`** (`from google.adk.models.lite_llm import LiteLlm`). `google-adk` declares `litellm` only under its own `extensions`/`test` extras — from the installed dist metadata:

```
$ python -c "import importlib.metadata as md; [print(r) for r in md.metadata('google-adk').get_all('Requires-Dist') if 'litellm' in r]"
litellm>=1.84 ; extra == "extensions"
litellm>=1.84 ; extra == "test"
```

Reproduced in a clean venv with only `google-adk` installed:

```
$ python -c "from google.adk.models.lite_llm import LiteLlm"
ImportError: LiteLLM support requires: pip install google-adk[extensions]
```

It's a module-level import, so this is a pytest **collection error** → session exit 1 → the whole `tests_integ` job fails, not just this file. Corroborating: the `Run integration tests` check on `2883ce0` is **FAILURE** at the `hatch test tests_integ` step, while the same job was **SUCCESS** on #338's head (`baa56ee`). I can't read the run log from here, so I'm stating that as correlation + a reproduced mechanism, not a log-confirmed root cause.

```diff
-adk = ["google-adk>=2.0.0"]
+adk = ["google-adk[extensions]>=2.0.0"]
```

Worth updating the file's own docstring too (`tests_integ/test_adk_eval.py:3-4` says `pip install google-adk litellm ...`) — the sibling convention is to point at the extra: `pip install strands-agents-evals[adk]` (cf. `tests_integ/test_langchain_openinference_eval.py:8`).
</details>

<details>
<summary><b>🟡 Should-fix (5)</b></summary>

**1. The assertions check "did it run", not "did it map correctly".** All three `Experiment` tests end with the same trio (`:235-237`, `:271-273`, `:314-316`). I replayed a captured real span set through those exact assertions under two injected mapper regressions:

| scenario | `len(scores) >= N` | `overall_score >= 0.5` | `all(test_passes)` | result |
|---|---|---|---|---|
| baseline (PR as written) | PASS | PASS (0.917) | PASS | **PASS** |
| trace split disabled (pre-#326) | PASS | PASS (0.792) | **FAIL** | **FAIL** |
| tool spans duplicated into every trace | PASS | PASS (0.833) | **FAIL** | **FAIL** |

So `len(report.scores) >= N` is a tautology (it's `n_evaluators × n_cases` regardless of mapping — and `>=` means a regression that *duplicates* rows sails through; `==` would catch it), and `overall_score >= 0.5` is dead given the next line: every evaluator here sets `test_pass` at a threshold ≥ 0.5, so `all(test_passes)` already forces the mean ≥ 0.5. `all(test_passes)` *does* have teeth — but it caught both regressions via exactly **one** evaluator (`ToolSelectionAccuracyEvaluator`, tool-level, scoped by each trace's `available_tools`). Every trace-level judge only ever scores the *last* trace (`evaluators/evaluator.py:136`, `parsed_inputs[-1]`), so anything wrong on the coordinator's trace is structurally invisible.

Nothing in the file inspects the mapped `Session` beyond `:193-195` (`session_id`, `len(traces) > 0`). Would it be worth capturing the `session` built inside `task_function` and asserting on it directly? For `test_adk_multi_agent_evaluation` (`:281-316`) that'd be `len(session.traces) == 2`, the per-trace agent names, `get_weather` present only on the specialist trace, and no tool span double-counted on the coordinator — i.e. exactly what `_build_traces` (`adk_otel_session_mapper.py:104-182`) claims to do, which is the one thing a *live* trace proves that the synthetic `TestMultiAgentSplitting` unit tests can't. Cheapest add: assert `tool_call_id is not None` — I confirmed Bedrock emits `gen_ai.tool.call.id = 'tooluse_…'` and ADK propagates it into `ToolCall`/`ToolCallContent`/`ToolResultContent` (`:509,526,552`), so that path *is* integration-testable here despite the docstring's "requires Gemini 3+" caveat at `:20-25` (which is misleading for the LiteLlm path).

**2. Two real mapper defects this suite surfaced but doesn't catch — pre-existing in #326, not this diff.** From the live multi-agent replay:
- the coordinator's `agent_response` comes back as raw tool JSON `'{"result": null}'`, via the fallback at `adk_otel_session_mapper.py:293-306`;
- the specialist's `user_prompt` is ADK delegation plumbing, not the user's question: ``"For context:[coordinator] `transfer_to_agent` tool returned result: {'result': None}"`` — `_extract_user_prompt_from_request` (`:443-456`) takes the *last* user turn, which is right for multi-turn and wrong for a multi-agent handoff (the mirror image of what #326 fixed).

Downstream, `CorrectnessEvaluator` scored the specialist trace **1.00** while its own reasoning read "*the user query is empty*" — `_extract_user_prompt` (`evaluators/evaluator.py:138-156`) reads `session_history[-1]`, which is a `list[ToolExecution]` there, → `""`. These are out of scope for a test-only PR, but they're the strongest argument for point 1, and they're freshly reproduced — file as follow-ups against #326?

**3. No `temperature` pin, on two stacked live-LLM layers.** `:31` — the sibling pins `model_kwargs={"temperature": 0}` (`test_langchain_openinference_eval.py:117`). It already shows: `ConcisenessEvaluator` scored the multi-agent case at exactly **0.50** in every run — the lowest passing label on a 3-way categorical (`conciseness_evaluator.py:34-38,74`) — because the coordinator's summary length/wording drifts run to run (106/121/109 chars, all adding unsolicited flourishes). No flake observed in 3+2 runs, but it's sitting on the pass/fail line by construction.

**4. `google-adk>=2.0.0` (`:68`) is floor-only** on a fast-moving SDK (2.0.0 → 2.6.0 already; unpinned, `pip` resolves 2.6.0 today, and the litellm range moved between those releases). Repo convention caps comparable optional deps — `langfuse>=2.0.0,<4` (`:58`), `opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0` (`:66`). `<3.0.0`?

**5. `test_adk_mapper_detection` (`:198-208`) spends a full live Bedrock run to assert `isinstance(mapper, ADKOtelSessionMapper)`** — pure dict/string routing with zero LLM dependency (`mappers/utils.py:134-135`), and near-duplicate of `test_adk_single_query`, which already calls `detect_otel_mapper` one line earlier. Meanwhile the genuine gap is at the unit level: `tests/strands_evals/mappers/test_utils.py::TestDetectOtelMapper` covers langchain/openinference/strands/smolagents/unknown but has **no ADK case**, and the `make_span_dict` helper it needs is already in that file. Fold the `isinstance` assert into `test_adk_single_query` and add the free unit test instead?
</details>

<details>
<summary><b>⚪ Nits (bundled)</b></summary>

- `app_name="test_app"` is hardcoded at `:162` independently of the `InMemoryRunner`s at `:94,113,143`. All four agree today; `app_name=runner.app_name` removes the DRY trap.
- All three agent fixtures return `(agent, runner)` and all five call sites discard the agent (`:184,202,224,259,294`) — return just the runner.
- `Case.expected_output` is set at `:217,246,252,287` but no evaluator used in this file reads it (only `expected_assertion` drives them) — inert; drop it or note it.
- Model id uses `us.anthropic.claude-sonnet-4-6` while the rest of the repo uses the `global.` prefix for that model (`evaluators/evaluator.py:24`, `detectors/constants.py:4`). Inconsistent, not wrong — I confirmed the `us.` profile is live-invokable.
- `:77` adding `adk` to the shared `hatch-test` env does pull `google-adk` into all 11 unit-CI jobs that never import it — but that's exactly what the existing `langchain` feature does (no unit test imports `langchain`/`langgraph`/`openinference` either; I checked), and the integ job needs it in this env. Pre-existing pattern, not yours to fix — and empirically harmless here: all 11 unit jobs incl. py3.14 are green on this SHA.
</details>

<details>
<summary><b>❓ Questions (all non-blocking)</b></summary>

- `_run_adk_agent` uses a fresh `user_id` per call to stop ADK reusing history (`:154-160`, comment at `:157`), which sidesteps `_extract_user_prompt_from_request`'s multi-turn logic — the exact path #326 had to fix, and the one whose real request shape only a live ADK run can prove. Deliberate scope-narrowing for this PR, or worth one added multi-turn scenario?
- Should the two mapper defects in 🟡2 be filed against #326 now while they're reproduced, or is there already a follow-up issue covering them?
- The PR is labelled `chore` but titled `feat:`, and `.github/labelers/type.yml:8-9` treats the title prefix as authoritative (`feat -> enhancement`). Labeler artifact from a title edit, or intentional? (`test:` would arguably describe the diff best, since there's no `src/` change.)
</details>

---
<sub>Reviewed by `strandly-the-agent` — an experimental agent, so treat this as work for a human to sanity-check, not a verdict. Test runs were against live Bedrock from a sandbox; nothing was pushed and no AWS resources were created.</sub>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings in a single comment.

strandly-the-agent:
**One blocker left, and it's my fault — plus one of my last-round recommendations was wrong.** Your new coordinator-trace assertion is factually wrong about ADK, because my "not double-counted on the coordinator" wording under-specified it: `transfer_to_agent` *is* the coordinator's own tool call and legitimately lands on its trace. Separately, my `google-adk[extensions]` suggestion is what made `0c8fc62` red — your hand-rolled `litellm` declaration is the better fix, keep it.

✅ Reviewed `5db3fd9` (incremental vs `2883ce0`) · ✅ reproduced the **exact current CI failure** twice cold — serially on py3.12 and under `-n auto` on py3.10 in a CI-equivalent feature env: `1 failed, 3 passed` at `tests_integ/test_adk_eval.py:341` · ✅ diagnosed `0c8fc62`'s *different* red (langgraph/langgraph-prebuilt incoherence from `[extensions]`, breaking the two langchain integ modules at collection — the ADK file itself passed 5/5 there) · ✅ `temperature=0` verified both accepted **and** forwarded into every `acompletion` call · ✅ 9 of my 11 previous findings resolved · not verified: the literal CI logs (no access — I reproduce rather than read them), and `hatch` itself (approximated its env with `uv` + CPython 3.10).

<details>
<summary><b>🔴 Blocker (1) — the current CI red, and my mis-specification</b></summary>

**`tests_integ/test_adk_eval.py:340-343`** — `assert len(trace_tool_spans) == 0` on the coordinator-only trace can never pass. ADK's delegation is implemented *as a tool call*, so the coordinator genuinely makes one, and the mapper correctly attributes it to the coordinator's trace:

```
raw span names   = ['call_llm', 'execute_tool get_weather', 'execute_tool transfer_to_agent',
                    'generate_content bedrock/...', 'invocation',
                    'invoke_agent coordinator', 'invoke_agent weather_specialist']
mapped traces    = 2
TRACE[0] agents=['coordinator']        n_tool=1 tool_names=['transfer_to_agent']
          args={'agent_name': 'weather_specialist'} tool_type='TransferToAgentTool'
TRACE[1] agents=['weather_specialist'] n_tool=1 tool_names=['get_weather']
          args={'city': 'Seattle'} tool_type='FunctionTool'
```

The failure, reproduced:

```
tests_integ/test_adk_eval.py::test_adk_multi_agent_evaluation FAILED
>   assert len(trace_tool_spans) == 0, (
E   AssertionError: Tool spans should not appear on the coordinator trace (found 1 tool spans)
E   assert 1 == 0
E    +  where 1 = len([ToolExecutionSpan(... 'tool_type': 'TransferToAgentTool' ...
E        tool_call=ToolCall(name='transfer_to_agent',
E                           arguments={'agent_name': 'weather_specialist'},
E                           tool_call_id='tooluse_VkleAALY0u4zkkFZEOD1e0'))])
tests_integ/test_adk_eval.py:341: AssertionError
============== 1 failed, 3 passed, 5 warnings in 71.79s ==============
```

To be clear about where this came from: I wrote "no tool span double-counted on the coordinator", meaning *the specialist's* `get_weather` must not leak onto the parent trace. Implementing that as "zero tool spans" is a reasonable reading of what I wrote — my wording was the imprecise part. Your comment at `:334` states the right intent; only the assertion overshoots it.

This version encodes the intent and is verified passing against the live trace:

```suggestion
        if "coordinator" in trace_agent_names and "weather_specialist" not in trace_agent_names:
            tool_names = {s.tool_call.name for s in trace_tool_spans}
            # transfer_to_agent is legitimately the coordinator's OWN tool call (ADK delegation);
            # what must NOT leak onto this trace is the specialist's tool.
            assert "get_weather" not in tool_names, f"Specialist tool leaked onto coordinator trace: {tool_names}"
            assert tool_names <= {"transfer_to_agent"}, f"Unexpected tools on coordinator trace: {tool_names}"
```

Worth noting this is the assertion set doing real work: it's now pinned to a genuine ADK delegation shape that the synthetic `TestMultiAgentSplitting` unit tests can't reproduce.
</details>

<details>
<summary><b>↩️ Correction — my <code>google-adk[extensions]</code> recommendation was wrong for this repo</b></summary>

`0c8fc62` did exactly what I asked and went red for an entirely different reason, which I should have anticipated: `hatch test` builds **one** env from `features = ["otel","langfuse","langchain","opensearch","adk"]` (`pyproject.toml:74-77`), so the `adk` extra's transitive closure lands on every integ test. `[extensions]` pulls 53 extra packages (anthropic, llama-index-core, pandas, sqlalchemy, kubernetes, toolbox-adk…) and drags the langchain stack backwards:

```
head pkgs: 158    0c8fc62[extensions] pkgs: 209
langchain: 1.3.14 -> 0.3.30     langgraph: 1.2.10 -> 0.4.7
langchain-core: 1.5.3 -> 0.3.86 langgraph-prebuilt: 1.1.0 -> 1.0.1
numpy: 2.2.6 -> 1.26.4
```

`langgraph 0.4.7` + `langgraph-prebuilt 1.0.1` is incoherent (prebuilt declares no langgraph bound), so on py3.10:

```
ERROR collecting tests_integ/test_langchain_openinference_eval.py
  from langgraph.prebuilt import create_react_agent
  -> langgraph/prebuilt/chat_agent_executor.py:32: from langgraph._internal._runnable import RunnableCallable
E ModuleNotFoundError: No module named 'langgraph._internal'
ERROR collecting tests_integ/test_langchain_traceloop_eval.py   (same)
!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!
```

Counterfactual on your head deps, same py3.10: `49 tests collected, 0 errors`. So **keep the direct `litellm` declaration** — it's the correct call for a shared-env repo, and my original suggestion wasn't. (FWIW the `langgraph!=0.5.0` pin you dropped in `5db3fd9` was a no-op either way — the resolver picks 1.2.10 regardless.)
</details>

<details>
<summary><b>🟡 Should-fix (1)</b></summary>

**`pyproject.toml:68` — raise the litellm floor to `>=1.84`.** `google-adk` 2.x asks for `litellm>=1.84` inside its own `extensions` extra, and since that extra isn't activated, nothing links the two: `>=1.75.0` is ~9 minors below what ADK wants and resolves with no complaint. The guard is presence-only, not version-aware (`google/adk/models/lite_llm.py:46`: `if ... importlib.util.find_spec("litellm") is None`), so a mismatch surfaces as an `AttributeError`/`TypeError` deep inside a completion call rather than a clean install error. I confirmed litellm 1.75.0 + google-adk 2.6.0 does install, import and construct fine today — so this is a latent trap, not a live break.

```diff
-adk = ["google-adk>=2.0.0,<3.0.0", "litellm>=1.75.0"]
+adk = ["google-adk>=2.0.0,<3.0.0", "litellm>=1.84"]
```
</details>

<details>
<summary><b>✅ Resolved from last round (9)</b></summary>

- 🔴 litellm missing from the `adk` extra — fixed (via the direct declaration; see the correction above).
- Exact score counts: `== 1` / `== 4` / `== 6` replace the `>=` tautologies (`:221`, `:257`, `:318`).
- The vacuous `overall_score >= 0.5` line is gone from all three tests.
- The mapped `Session` is now asserted directly — trace count, per-trace agent names, tool scoping (`:321-353`), and `tool_call_id is not None` in both the multi-tool (`:251-259`) and multi-agent tests. That's the substantive change, and it's the right one.
- `test_adk_mapper_detection` deleted, `isinstance` folded into `test_adk_single_query` (`:187`) — one fewer live Bedrock run for a dict lookup.
- `app_name=runner.app_name` (`:160`) replaces the duplicated literal.
- Fixtures return just the runner; the `_, runner =` unpacking is gone throughout.
- Docstring now points at `pip install strands-agents-evals[adk]` (`:4`), matching the sibling convention.
- `temperature=0` (`:31`) — and I verified it isn't cosmetic: it lands in `_additional_args` (`lite_llm.py:2730`) and is forwarded via `completion_args.update(...)` (`:2782`), confirmed `temperature=0` on every captured `acompletion` call. One caveat for later: `:2792-2793` applies `generation_params` afterwards, so an agent-level `generate_content_config.temperature` would silently win. None of these fixtures set one.
- `<3.0.0` cap added, as suggested.
</details>

<details>
<summary><b>⚪ Still open (minor, none blocking)</b></summary>

- The free unit test is still missing: `tests/strands_evals/mappers/test_utils.py::TestDetectOtelMapper` has no ADK case (no `SCOPE_ADK` reference in the file), so ADK-scope routing is still only covered by a live run. Two lines with the existing `make_span_dict` helper.
- `Case.expected_output` is still set at `:204,232,238,286` but no evaluator in this file reads it — inert, harmless.
- Unanswered from last round, both fine to defer: whether the multi-turn gap (fresh `user_id` per call, `:155-158`) is deliberate scope-narrowing; and whether the two reproduced mapper defects (raw-JSON `agent_response` on the delegating agent; plumbing-text `user_prompt` on the delegated one) should be filed against #326. Happy to open those issues if useful.
</details>

---
<sub>Reviewed by `strandly-the-agent` — an experimental agent, and this round is a good illustration: two of my last-round calls needed correcting. Treat it as work for a human to sanity-check. Runs were against live Bedrock from a sandbox; nothing pushed, no AWS resources created.</sub>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings in a single comment.

strandly-the-agent:
**No new findings — and the integ job is green for the first time on this PR.** Both items from last round are in, and the model-id change resolves an earlier nit better than what I'd suggested.

✅ `Run integration tests` on `8b4b4c9`: **SUCCESS** (it was FAILURE on `2883ce0`, `0c8fc62`, `5db3fd9` and `dd12768`) · ✅ `:343-345` is the replacement assertion verbatim · ✅ `pyproject.toml:68` → `litellm>=1.84` · ✅ lint + all 11 unit jobs green.

One thing worth calling out, because it tightens the earlier diagnosis: `dd12768` changed *only* the model id and left the bad assertion in place — and stayed red. `8b4b4c9` then fixed the assertion and went green. That isolates the cause to the assertion rather than the dependencies or the model, which is a cleaner confirmation than my local reproduction was.

On the model change (`:32-34`) — good call, and it's better than what I proposed. Extracting `DEFAULT_MODEL_ID = "us.anthropic.claude-sonnet-4-20250514-v1:0"` matches both sibling modules exactly (`test_langchain_openinference_eval.py:28`, `test_langchain_traceloop_eval.py:31`), which is a more useful kind of consistency than the `global.`-prefix alignment I'd suggested. It also re-proves on a different model what I'd only verified on sonnet-4-6: CI passing means the `tool_call_id is not None` and 2-trace-split assertions hold on Sonnet 4 too.

I deliberately didn't re-run anything locally this round — CI's own green on py3.10 under `hatch` is stronger evidence than my sandbox, so re-running would have burned Bedrock calls to learn less.

From my side this is ready to come out of draft. Two optional leftovers, neither blocking: the ADK case in `tests/strands_evals/mappers/test_utils.py::TestDetectOtelMapper` (still absent — two lines with the existing `make_span_dict` helper) and the inert `Case.expected_output` fields. And the offer stands to file the two reproduced mapper defects (delegating agent's raw-JSON `agent_response`; delegated agent's plumbing-text `user_prompt`) as follow-ups against #326 — just say the word and I'll open them.

---
<sub>Reviewed by `strandly-the-agent` — an experimental agent; a human should still sign off. No local runs this round; nothing pushed, no AWS resources created.</sub>

tests_integ/test_adk_eval.py
```diff
@@ -0,0 +1,356 @@
...
+            input="What's the weather in Seattle?",
+            expected_output="The weather in Seattle is rainy and 55F.",
+            expected_assertion="The agent obtained weather information for Seattle and responded with the result.",
+        ),
+    ]
+
+    captured_session = None
+
+    def task_function(case: Case) -> dict:
+        nonlocal captured_session
```
poshinchen: why do we need this?

pyproject.toml
```diff
@@ -65,6 +65,7 @@ langchain = [
     "openinference-instrumentation-langchain>=0.1.0",
     "opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0",
 ]
+adk = ["google-adk>=2.0.0,<3.0.0", "litellm>=1.84"]
```
poshinchen: so litellm can be removed?

tests_integ/test_adk_eval.py
liramon2: To inspect the session object structure and perform assertions on its spans. Another option is to deserialize the report trajectory for the session, which would avoid the non-local variable. Let me know if that's cleaner.

pyproject.toml
liramon2: Yeah, will do.

tests_integ/test_adk_eval.py
liramon2: Update: implemented the deserialized session capture for comparison

poshinchen (approved):

Issue #342, opened 2026-08-03
[BUG] Judge-based evaluators silently score 0 when a trace exceeds the judge's context window


pdebjyot:
# [BUG] Judge-based evaluators silently score 0 when a trace exceeds the judge's context window

## Summary

When the trajectory handed to a judge-based evaluator (`OutputEvaluator`,
`TrajectoryEvaluator`, the `Multimodal*` family) is larger than the judge
model's context window, the judge call raises `ContextWindowOverflowException`.
`Experiment._run_evaluator` catches it under error isolation and records the
case as **`score: 0, test_pass: False`** — indistinguishable from a genuine
quality failure. A correct agent response gets a false-negative failing score
purely because its trace was too big for the judge to read.

This is not specific to any telemetry backend, and not specific to strands-evals
either — it's a property of the *inline-the-whole-trace* judging pattern. We
confirmed this with a 200-trace labeled corpus run through three frameworks
(strands-evals, DeepEval, and a Langfuse-style managed LLM-as-a-judge) on a
shared judge model: all three overflow at the same trace size, with three
distinct error signatures (Strands `ContextWindowOverflowException`; DeepEval and
the Langfuse-style judge a Bedrock `ValidationException`; DeepEval additionally
degrading to malformed-JSON output). The cause is the evaluator
prompt-composition path, not the provider.

## Where it happens

The judge agent receives the whole trajectory inlined into its prompt. On a
large trace the model overflows, and the failure is swallowed here:

```python
# experiment.py — _run_evaluator
except Exception as e:
    # Catch non-throttling errors and record as failure (error isolation)
    return {
        "evaluator_name": evaluator.get_name(),
        "evaluator_type": evaluator.get_type_name(),
        "test_pass": False,
        "score": 0,
        "reason": f"Evaluator error: {str(e)}",
        "detailed_results": [],
    }
```

Error isolation is the right default (one bad case shouldn't abort a run), but
a context-window overflow is a *capability* failure of the evaluation harness,
not a *quality* signal about the agent. Reporting it as `score: 0` conflates
the two and quietly corrupts aggregate metrics.

## Reproduction

Minimal, backend-independent — build a Session whose serialized trajectory
exceeds the judge window and evaluate a correct output against it:

```python
from strands_evals.evaluators import OutputEvaluator
from strands_evals.types import EvaluationData

# A large trajectory (e.g. a real multi-step agent run with sizable tool
# results) serialized into the evaluation prompt.
[...]

PR #343, opened 2026-08-03
feat(tools): add TraceIndex for progressive trace disclosure in judge-based evaluators

pdebjyot:
> **Scope:** this implements **#342 item (2) only** — the read-side capability (a judge can read a large trace without inlining it). It does **not** change `_run_evaluator`'s error handling (item (1): telling harness overflow apart from a quality `score: 0`), and it does **not** touch the `TrajectoryEvaluator` path, which still inlines the full trajectory. Both remain open; #342 stays open. ("Addresses" is not a closing keyword.)

# feat(tools): TraceIndex — progressive trace disclosure for judge-based evaluators

**Branch**: `feat/trace-index`
**Builds on**: #324 (custom `tools=` on `OutputEvaluator` / `TrajectoryEvaluator`)
**Addresses**: #342 (item (2); items (1) and the trajectory path remain open)

## Problem

When the trajectory handed to a judge-based evaluator is larger than the judge model's context window, the judge call raises `ContextWindowOverflowException`. `Experiment._run_evaluator` catches it under error isolation and records the case as `score: 0, test_pass: False` — indistinguishable from a genuine quality failure. A correct agent response gets a false-negative failing score purely because its trace was too big for the judge to read.

This isn't a theoretical edge. We observe it on real production agent traces, where routine multi-step sessions serialize past a 200K-token judge window and the largest run into the millions of tokens — so under a default Sonnet-class judge, correct agents are being silently scored as failures today. The problem is stack-agnostic: we reproduced identical overflow behavior across strands-evals, DeepEval, and a Langfuse-style managed judge on a shared Bedrock judge model (three distinct error signatures, same trace-size cliff).

The measured evidence below is from a **synthetic, deterministic, offline** benchmark — no real data — so it's independently reproducible.

## What this PR adds

`TraceIndex` — an in-memory index over a `Session` that lets a judge read a large trace **without inlining it**. It is an established pattern: MLflow's Agent-as-a-Judge trace scorers hand the judge `ListSpans` / `GetSpan` / `SearchTraceRegex`, and Zhuge et al.'s Agent-as-a-Judge ([arXiv:2410.10934](https://arxiv.org/abs/2410.10934)) uses `retrieve`/`read`/`locate` modules to pull only the relevant segments.

```python
from strands_evals.tools.trace_index import TraceIndex
from strands_evals.types import EvaluationData

index = TraceIndex(session)  # session: a Session from any provider/mapper

# for_judge() returns both halves atomically so neither is forgotten:
prompt_section, tools = index.for_judge()  # tools: list_spans, get_span, search_spans

evaluator = OutputEvaluator(
    rubric=(
        "Every factual claim must be supported by tool-result evidence in the trace. "
        "Use the trace tools to verify each claim before scoring."
    ),
    tools=tools,
)

# Compact overview in the judged output instead of the full trajectory:
output = f"{agent_answer}\n{prompt_section}"
evaluator.evaluate(EvaluationData(input="(from trace)", actual_output=output))
```

- **`overview()`** — one compact line per span (index, type, tool name, sizes, preview, tool ok/error status). Paged by span index through `max_read_chars`, so even the *first documented call* on a huge trace can't overflow. The header tells the judge previews are truncated and that claims must be verified with the tools before scoring.
- **`list_spans` / `get_span` / `search_spans`** — discovery tools the judge calls on demand (mirrors the MLflow triad). `get_span` pages oversized spans via `max_read_chars` (default 8000) + `offset`; `search_spans` is **literal by default** (opt-in `is_regex=True`) and searches an unescaped rendering of every visible field, including `system_prompt` and `available_tools`, so a literal a judge copies from the overview (`$150`) matches.
- **Backend-agnostic** — consumes any `Session` a provider/mapper produces; timestamps are tz-normalized before sorting so a mix of naive/aware spans (from different mappers) can't crash construction.

- **`for_judge()`** — returns `(prompt_section, tools)` together, so the two composition halves (overview into the judged output, tools onto the evaluator) can't be done half-way and silently degrade the judge.

Two disclosure strategies compose from these pieces: **index** (substitute `overview()` into the prompt — portable, works with any judge) and **explore** (index + discovery tools via #324's `tools=` — Strands-native, best accuracy).

## Evidence

**Cross-framework matrix (1344 cells).** 200-trace labeled corpus × 3 frameworks (strands-evals, DeepEval, Langfuse-style) × 4 metrics (groundedness, accuracy, trajectory, tool_use) × inline/index/explore, all bound to one shared Bedrock judge. Each metric scored as a binary classifier against planted ground truth (overflow ⇒ wrong prediction). These numbers are external motivation from the offline benchmark above — they are not reproducible from anything in-tree:

- The overflow cliff is **stack-agnostic** and lands at the same trace size for all three frameworks.
- `index` and `explore` **hold accuracy on traces that fit and recover it on traces that overflow** (inline is unusable in the overflow bucket; index/explore score correctly).
- The judge-side `explore` tools **beat a bare index when the decisive fact is buried in a large tool result**: on `wrong_tool` traces the deciding refund amount sits inside a large search result that `overview()` elides, so the index judge false-fails groundedness (0.83); `get_span`/`search_spans` let the judge retrieve it → groundedness **1.00**.
- **No degradation** on traces that fit: index/explore ≈ inline everywhere inline still works.

**Ground-truth A/B** (grounded / fabricated claims, evidence buried mid-trace): the index-equipped judge separates grounded from fabricated where the inline judge overflows — captured as the integ test `test_judge_reliability_inline_vs_explore`.

## Known limitation (called out honestly)

`TrajectoryEvaluator` inlines `actual_trajectory` **unconditionally** (`case_prompt_template.py:51`), so the trajectory metric still overflows on large traces even with the index — the substitution only reaches evaluators that route through the caller-controlled `actual_output` (the output-family metrics, fixed today). Making the trajectory metric disclosure-aware needs a template change and is proposed as a follow-up; this PR does not change that path.

Separately, this PR adds the *capability* but does not change `_run_evaluator`'s error handling — distinguishing harness overflow from a quality `score: 0` is tracked in #342 as an independent change.

## Testing

- 34 unit tests (overview formatting + paging; list/get/search behavior; literal-vs-regex search, per-span match counts, truncation signalling; `system_prompt`/`available_tools` exposure; tool-error surfacing; `max_read_chars`/offset validation; mixed naive/aware timestamps; `for_judge()` composition; cross-pattern; end-to-end through `OutputEvaluator` with `tools=index.tools`) — all pass.
- 1 integ A/B test (skips without live Bedrock credentials).
- Full suite **1764 passed**; `ruff check` / `ruff format` clean.

## Revision (in response to review)

- **Blocker 1** — `overview()`/`list_spans` now page by span index through `max_read_chars`, so the judge's first call can't overflow.
- **Blocker 2** — `search_spans` is literal by default (opt-in `is_regex`), searches an unescaped rendering, counts per-span matches, signals truncation, and no longer false-positives on serialization artifacts like `"error": null`.
- **Blocker 3** — `get_span`/`search_spans` now expose `system_prompt` and `available_tools`.
- **Blocker 4** — module docstring example switched to `OutputEvaluator` with an explicit `TrajectoryEvaluator` limitation note.
- **tz crash** — `_flatten_spans` sorts via the repo's `_to_aware_utc`, fixing the mixed naive/aware `TypeError` (root cause #372).
- Yellow: `max_read_chars`/negative-offset validation, tool ok/ERROR status + error preview and INFERENCE size/preview in the overview, truncation markers on `search_spans`, and a verify-before-score header the model actually sees.
- **Composition** — added `for_judge()` → `(prompt_section, tools)` so the overview and tools are wired together atomically instead of through two independent, silently-degrading steps.

## Checklist

- [x] Backend-agnostic, additive (no change to existing evaluator behavior)
- [x] Unit + integration tests added
- [x] Conventional commits
- [x] Lint/format clean
- [x] Builds on merged #324; known limitation documented, not hidden

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,31 @@ evaluator = TrajectoryEvaluator(
 )
 ```
 
+### Evaluating Large Traces with Progressive Disclosure
+
+When a session is too large to inline into a judge prompt (large tool results,
+many turns), give the judge a compact overview plus discovery tools instead of
+the full trajectory. The judge loads only the spans the rubric requires:
+
+```python
+from strands_evals.evaluators import OutputEvaluator
+from strands_evals.tools.trace_index import TraceIndex
+
+index = TraceIndex(session)  # session: a Session from any provider/mapper
+
+evaluator = OutputEvaluator(
+    rubric="Every factual claim must be supported by tool-result evidence in the trace.",
+    tools=index.tools,  # list_spans, get_span, search_spans
+)
+
+# Compose the prompt with the compact overview instead of the full trajectory
+evaluation_output = f"{agent_answer}\n\n<TraceOverview>\n{index.overview()}\n</TraceOverview>"
+```
+
+The overview is one line per span (index, type, tool name, sizes, preview);
+`get_span` pages through oversized spans so no single tool return can overflow
+the judge's context.
+
 ### Trace-based Helpfulness Evaluation
 
 Evaluate agent helpfulness using OpenTelemetry traces with seven-level scoring:
--- a/README.md
+++ b/README.md
@@ -231,9 +231,10 @@ evaluator.evaluate(EvaluationData(input=user_prompt, actual_output=judged_output
 ```
 
 The overview is one line per span (index, type, tool name, sizes, preview);
-`list_spans` and `get_span` page through long traces and oversized spans so no
-single tool return can overflow the judge's context. The rubric must tell the
-judge to verify claims with the tools — otherwise it scores off the previews alone.
+`list_spans`, `get_span`, and `search_spans` all page or cap their output at
+`max_read_chars` so no single tool return can overflow the judge's context. The
+rubric must tell the judge to verify claims with the tools — otherwise it scores
+off the previews alone.
 
 > **Note:** this composes with `OutputEvaluator`, whose prompt is caller-controlled.
 > It does **not** work with `TrajectoryEvaluator`, which inlines the full
--- a/README.md
+++ b/README.md
@@ -209,21 +209,36 @@ the full trajectory. The judge loads only the spans the rubric requires:
 ```python
 from strands_evals.evaluators import OutputEvaluator
 from strands_evals.tools.trace_index import TraceIndex
+from strands_evals.types import EvaluationData
 
 index = TraceIndex(session)  # session: a Session from any provider/mapper
 
+# for_judge() hands back both halves together so neither is forgotten:
+# the overview to put next to the answer, and the discovery tools for the judge.
+prompt_section, tools = index.for_judge()  # tools: list_spans, get_span, search_spans
+
 evaluator = OutputEvaluator(
-    rubric="Every factual claim must be supported by tool-result evidence in the trace.",
-    tools=index.tools,  # list_spans, get_span, search_spans
+    rubric=(
+        "Every factual claim must be supported by tool-result evidence in the trace. "
+        "Use the trace tools to verify each claim against the evidence before scoring."
+    ),
+    tools=tools,
 )
 
-# Compose the prompt with the compact overview instead of the full trajectory
-evaluation_output = f"{agent_answer}\n\n<TraceOverview>\n{index.overview()}\n</TraceOverview>"
+# Compose the overview into the judged output instead of the full trajectory:
+judged_output = f"{agent_answer}\n{prompt_section}"
+evaluator.evaluate(EvaluationData(input=user_prompt, actual_output=judged_output))
 ```
 
 The overview is one line per span (index, type, tool name, sizes, preview);
-`get_span` pages through oversized spans so no single tool return can overflow
-the judge's context.
+`list_spans` and `get_span` page through long traces and oversized spans so no
+single tool return can overflow the judge's context. The rubric must tell the
+judge to verify claims with the tools — otherwise it scores off the previews alone.
+
+> **Note:** this composes with `OutputEvaluator`, whose prompt is caller-controlled.
+> It does **not** work with `TrajectoryEvaluator`, which inlines the full
+> `actual_trajectory` unconditionally and would re-create the overflow this pattern
+> exists to prevent.
 
 ### Trace-based Helpfulness Evaluation
 
--- a/src/strands_evals/tools/trace_index.py
+++ b/src/strands_evals/tools/trace_index.py
@@ -45,10 +45,15 @@ from strands import tool
 from ..extractors.trace_extractor import _to_aware_utc
 from ..types.trace import (
     AgentInvocationSpan,
+    AssistantMessage,
     InferenceSpan,
     Session,
     SpanUnion,
+    TextContent,
+    ToolCallContent,
     ToolExecutionSpan,
+    ToolResultContent,
+    UserMessage,
 )
 
 _PREVIEW_CHARS = 120
@@ -92,6 +97,28 @@ def _span_text(span: SpanUnion) -> str:
     return json.dumps(span.model_dump(), default=str)
 
 
+def _render_message(msg: UserMessage | AssistantMessage) -> str:
+    """Plain-text rendering of one inference message from its fields.
+
+    Renders role + each content block by field (text, or tool name/args/result/
+    error) like the tool and agent branches, rather than dumping the pydantic dict.
+    A raw `model_dump()` repr leaks structural keys such as ``'error': None`` into the
+    text, which turns an "error" search into a false positive on every inference span
+    and hides genuine failures behind serialization noise.
+    """
+    parts = [msg.role.value]
+    for block in msg.content:
+        if isinstance(block, TextContent):
+            parts.append(block.text)
+        elif isinstance(block, ToolCallContent):
+            parts.append(f"{block.name}({json.dumps(block.arguments, default=str)})")
+        elif isinstance(block, ToolResultContent):
+            parts.append(str(block.content))
+            if block.error:
+                parts.append(f"error: {block.error}")
+    return " ".join(p for p in parts if p)
+
+
 def _search_haystack(span: SpanUnion) -> str:
     """Plain-text rendering of a span for search matching.
 
@@ -107,7 +134,10 @@ def _search_haystack(span: SpanUnion) -> str:
             str(span.tool_result.content),
         ]
         if span.tool_result.error:
-            parts.append(str(span.tool_result.error))
+            # Prefix with an "error:" token so a judge searching the overview's
+            # [ERROR] vocabulary finds the failed tool; the raw value alone
+            # (e.g. "CardDeclined: insufficient funds") carries no such token.
+            parts.append(f"error: {span.tool_result.error}")
         return "\n".join(parts)
     if isinstance(span, AgentInvocationSpan):
         parts = [span.user_prompt, span.agent_response]
@@ -116,7 +146,7 @@ def _search_haystack(span: SpanUnion) -> str:
         parts += [f"{t.name}: {t.description or ''}" for t in span.available_tools]
         return "\n".join(parts)
     if isinstance(span, InferenceSpan):
-        return "\n".join(str(m.model_dump()) for m in span.messages)
+        return "\n".join(_render_message(m) for m in span.messages)
     return str(span.model_dump())
 
 
@@ -144,8 +174,9 @@ def _describe(span: SpanUnion) -> str:
             f"-> response: {len(span.agent_response)} chars: {_preview(span.agent_response)}"
         )
     if isinstance(span, InferenceSpan):
-        size = sum(len(str(m.model_dump())) for m in span.messages)
-        preview = _preview(" ".join(str(m.model_dump()) for m in span.messages))
+        rendered = [_render_message(m) for m in span.messages]
+        size = sum(len(r) for r in rendered)
+        preview = _preview(" ".join(rendered))
         return f"INFERENCE {len(span.messages)} messages, {size} chars: {preview}"
     return f"{type(span).__name__}"
 
@@ -204,19 +235,24 @@ class TraceIndex:
         def search_spans(pattern: str, max_matches: int = 20, is_regex: bool = False) -> str:
             """Search all span content for a literal string (default) or a regex.
             Matching is case-insensitive and covers every visible field, including
-            system prompts and tool configs. Returns matching span indices with a
-            short excerpt and per-span match count. Use get_span to load a match in full.
+            system prompts and tool configs. Regex anchors ^/$ match line boundaries.
+            Returns matching span indices with a short excerpt and per-span match
+            count, capped at max_read_chars total; use get_span to load a match in full.
 
             Args:
                 pattern: Text to search for. Treated literally unless is_regex=True.
-                max_matches: Maximum number of matching spans to return.
+                max_matches: Maximum number of matching spans to return. Results are
+                    also capped at max_read_chars total, whichever comes first.
                 is_regex: Set True to treat pattern as a regular expression.
             """
             if is_regex:
                 try:
-                    rx = re.compile(pattern, re.IGNORECASE)
-                except re.error as e:
-                    return f"ERROR: invalid regex {pattern!r}: {e}. Retry with is_regex=False for a literal search."
+                    # MULTILINE so ^/$ anchor to line boundaries in the newline-joined
+                    # haystack — LLMs write anchored regexes and would otherwise read a
+                    # silent "No matches" as "claim unsupported".
+                    rx = re.compile(pattern, re.IGNORECASE | re.MULTILINE)
+                except re.error as exc:
+                    return f"ERROR: invalid regex {pattern!r}: {exc}. Retry with is_regex=False for a literal search."
                 matcher = lambda text: [(m.start(), m.end()) for m in rx.finditer(text)]  # noqa: E731
             else:
                 needle = pattern.lower()
@@ -228,10 +264,12 @@ class TraceIndex:
                         start = i + max(len(needle), 1)
                     return spans
 
-            hits, truncated = [], False
+            hits: list[str] = []
+            stop_reason: str | None = None
+            used = 0
             for i, span in enumerate(this._spans):
                 if len(hits) >= max_matches:
-                    truncated = True
+                    stop_reason = "max_matches"
                     break
                 text = _search_haystack(span)
                 positions = matcher(text)
@@ -241,11 +279,24 @@ class TraceIndex:
                 excerpt = _preview(text[max(0, s - 60) : e + 60], 160)
                 count = len(positions)
                 suffix = f" ({count} matches)" if count > 1 else ""
-                hits.append(f"[{i}]{suffix} ...{excerpt}...")
+                line = f"[{i}]{suffix} ...{excerpt}..."
+                # Bound the whole response by max_read_chars, not max_matches alone:
+                # a generous max_matches on a long trace would otherwise blow past the
+                # per-call budget every other tool honors. Always keep at least one hit.
+                if hits and used + len(line) + 1 > this.max_read_chars:
+                    stop_reason = "budget"
+                    break
+                hits.append(line)
+                used += len(line) + 1
             if not hits:
                 return f"No matches for {pattern!r}"
-            if truncated:
+            if stop_reason == "max_matches":
                 hits.append(f"[stopped at {max_matches} spans; refine the pattern or raise max_matches for more]")
+            elif stop_reason == "budget":
+                hits.append(
+                    f"[budget reached at {this.max_read_chars} chars ({len(hits)} spans shown); "
+                    f"refine the pattern to narrow results]"
+                )
             return "\n".join(hits)
 
         self.tools = [list_spans, get_span, search_spans]
@@ -268,7 +319,8 @@ class TraceIndex:
             f"Previews are truncated; call get_span/search_spans to load full content "
             f"(up to {self.max_read_chars} chars per call) and verify claims before scoring."
         )
-        lines, used, end = [], len(header), offset
+        lines: list[str] = []
+        used, end = len(header), offset
         for i in range(offset, total):
             line = self._describe_lines[i]
             if lines and used + len(line) + 1 > self.max_read_chars:
--- a/src/strands_evals/tools/trace_index.py
+++ b/src/strands_evals/tools/trace_index.py
@@ -1,0 +1,177 @@
+"""Progressive trace disclosure for judge agents.
+
+A large agent trajectory does not fit in a judge's context window. Rather than
+inlining the whole Session into the evaluation prompt (which overflows and gets
+scored as a failure), `TraceIndex` builds a small in-memory index over the trace
+and gives the judge two things:
+
+1. An `overview()` — one line per span (index, type, tool name, sizes, truncated
+   preview) — cheap enough to always fit in context, and
+2. **Lookup tools** the judge calls to load only the spans it needs to verify the
+   rubric: `list_spans`, `get_span`, `search_spans`.
+
+This is the same list / get / search shape used to query any indexed collection,
+and the same progressive-disclosure pattern skills use: the overview is the
+"name + description" line; the tools load the full content on demand.
+
+Example::
+
+    from strands_evals.evaluators import TrajectoryEvaluator
+    from strands_evals.tools.trace_index import TraceIndex
+
+    index = TraceIndex(session)
+    evaluator = TrajectoryEvaluator(
+        rubric="Every claim in the final response must be supported by a tool result.",
+        tools=index.tools,
+    )
+    # Compose the prompt with index.overview() instead of the full trajectory.
+"""
+
+import json
+import re
+
+from strands import tool
+
+from ..types.trace import (
+    AgentInvocationSpan,
+    InferenceSpan,
+    Session,
+    SpanUnion,
+    ToolExecutionSpan,
+)
+
+_PREVIEW_CHARS = 120
+_DEFAULT_MAX_READ_CHARS = 8_000
+
+
+def _flatten_spans(session: Session) -> list[SpanUnion]:
+    """Flatten all spans across traces in start_time order."""
+    spans = [span for trace in session.traces for span in trace.spans]
+    spans.sort(key=lambda s: s.span_info.start_time)
+    return spans
+
+
+def _span_text(span: SpanUnion) -> str:
+    """Full text content of a span, for search and retrieval."""
+    if isinstance(span, ToolExecutionSpan):
+        return json.dumps(
+            {
+                "tool_call": span.tool_call.model_dump(),
+                "tool_result": span.tool_result.model_dump(),
+            },
+            default=str,
+        )
+    if isinstance(span, AgentInvocationSpan):
+        return json.dumps(
+            {"user_prompt": span.user_prompt, "agent_response": span.agent_response},
+            default=str,
+        )
+    if isinstance(span, InferenceSpan):
+        return json.dumps([m.model_dump() for m in span.messages], default=str)
+    return json.dumps(span.model_dump(), default=str)
+
+
+def _preview(text: str, limit: int = _PREVIEW_CHARS) -> str:
+    text = re.sub(r"\s+", " ", text).strip()
+    return text if len(text) <= limit else text[: limit - 3] + "..."
+
+
+def _describe(span: SpanUnion) -> str:
+    """One overview line describing a span without its full payload."""
+    if isinstance(span, ToolExecutionSpan):
+        args = json.dumps(span.tool_call.arguments, default=str)
+        result_size = len(str(span.tool_result.content))
+        return (
+            f"TOOL {span.tool_call.name}({_preview(args, 80)}) "
+            f"-> result: {result_size} chars: {_preview(str(span.tool_result.content))}"
+        )
+    if isinstance(span, AgentInvocationSpan):
+        return (
+            f"AGENT prompt: {_preview(span.user_prompt, 80)} "
+            f"-> response: {len(span.agent_response)} chars: {_preview(span.agent_response)}"
+        )
+    if isinstance(span, InferenceSpan):
+        return f"INFERENCE {len(span.messages)} messages"
+    return f"{type(span).__name__}"
+
+
+class TraceIndex:
+    """Read-only list / get / search index over a Session for judge agents.
+
+    Attributes:
+        session: The Session being evaluated.
+        max_read_chars: Cap on any single tool return, so a huge span can't
+            overflow the judge's context in one call. Oversized content is
+            windowed and the tool reports how to page through it.
+    """
+
+    def __init__(self, session: Session, max_read_chars: int = _DEFAULT_MAX_READ_CHARS):
+        self.session = session
+        self.max_read_chars = max_read_chars
+        self._spans = _flatten_spans(session)
+
+        # Bind instance state into plain functions so @tool sees clean signatures.
+        # `this` (not `index`) so the public get_span(index=...) arg name is free.
+        this = self
+
+        @tool
+        def list_spans() -> str:
+            """List every span in the trace: one line per span with its index, type,
+            tool name, argument preview, and result size. Call this first to decide
+            which spans to inspect."""
+            return this.overview()
+
+        @tool
+        def get_span(index: int, offset: int = 0) -> str:
+            """Get the full content of one span by its index from the span list.
+            Large spans are windowed; the response says how to page with offset.
+
+            Args:
+                index: Span index as shown by list_spans.
+                offset: Character offset for paging through oversized spans.
+            """
+            if not 0 <= index < len(this._spans):
+                return f"ERROR: index {index} out of range (0..{len(this._spans) - 1})"
+            return this._window(_span_text(this._spans[index]), offset)
+
+        @tool
+        def search_spans(pattern: str, max_matches: int = 20) -> str:
+            """Search all span content for a regex or literal string. Returns matching
+            span indices with a short excerpt around each match. Use get_span to load
+            a matching span in full.
+
+            Args:
+                pattern: Regex (or literal text) to search for.
+                max_matches: Maximum matches to return.
+            """
+            try:
+                rx = re.compile(pattern, re.IGNORECASE)
+            except re.error:
+                rx = re.compile(re.escape(pattern), re.IGNORECASE)
+            hits = []
+            for i, span in enumerate(this._spans):
+                text = _span_text(span)
+                m = rx.search(text)
+                if m:
+                    start = max(0, m.start() - 60)
+                    hits.append(f"[{i}] ...{_preview(text[start : m.end() + 60], 160)}...")
+                if len(hits) >= max_matches:
+                    break
+            return "\n".join(hits) if hits else f"No matches for {pattern!r}"
+
+        self.tools = [list_spans, get_span, search_spans]
+
+    def overview(self) -> str:
+        """Compact one-line-per-span overview of the session."""
+        lines = [f"Trace overview: {len(self._spans)} spans (session {self.session.session_id})"]
+        lines += [f"[{i}] {_describe(span)}" for i, span in enumerate(self._spans)]
+        return "\n".join(lines)
+
+    def _window(self, text: str, offset: int) -> str:
+        if offset >= len(text):
+            return f"ERROR: offset {offset} beyond content length {len(text)}"
+        window = text[offset : offset + self.max_read_chars]
+        if offset + len(window) < len(text):
+            remaining = len(text) - offset - len(window)
+            window += f"\n[TRUNCATED: {remaining} chars remain; call again with offset={offset + len(window)}]"
+        return window
--- a/src/strands_evals/tools/trace_index.py
+++ b/src/strands_evals/tools/trace_index.py
@@ -6,25 +6,35 @@ scored as a failure), `TraceIndex` builds a small in-memory index over the trace
 and gives the judge two things:
 
 1. An `overview()` — one line per span (index, type, tool name, sizes, truncated
-   preview) — cheap enough to always fit in context, and
+   preview) — paged so it always fits in context, and
 2. **Lookup tools** the judge calls to load only the spans it needs to verify the
    rubric: `list_spans`, `get_span`, `search_spans`.
 
-This is the same list / get / search shape used to query any indexed collection,
-and the same progressive-disclosure pattern skills use: the overview is the
-"name + description" line; the tools load the full content on demand.
+This is the same list / get / search shape used to query any indexed collection:
+the overview is the "name + description" line; the tools load the full content on
+demand.
+
+`TraceIndex` composes with `OutputEvaluator`, whose prompt is caller-controlled —
+put the overview in the judged output and pass the tools. It does **not** compose
+with `TrajectoryEvaluator`, which inlines the full `actual_trajectory`
+unconditionally and so re-creates the overflow this class exists to prevent.
 
 Example::
 
-    from strands_evals.evaluators import TrajectoryEvaluator
+    from strands_evals.evaluators import OutputEvaluator
     from strands_evals.tools.trace_index import TraceIndex
 
     index = TraceIndex(session)
-    evaluator = TrajectoryEvaluator(
-        rubric="Every claim in the final response must be supported by a tool result.",
-        tools=index.tools,
+    prompt_section, tools = index.for_judge()  # overview + tools together
+    evaluator = OutputEvaluator(
+        rubric=(
+            "Every claim in the final response must be supported by a tool result. "
+            "Use the trace tools to verify each claim before scoring."
+        ),
+        tools=tools,
     )
-    # Compose the prompt with index.overview() instead of the full trajectory.
+    # Put the compact overview next to the answer instead of the full trajectory:
+    judged_output = f"{agent_answer}\n{prompt_section}"
 """
 
 import json
@@ -32,6 +42,7 @@ import re
 
 from strands import tool
 
+from ..extractors.trace_extractor import _to_aware_utc
 from ..types.trace import (
     AgentInvocationSpan,
     InferenceSpan,
@@ -45,14 +56,19 @@ _DEFAULT_MAX_READ_CHARS = 8_000
 
 
 def _flatten_spans(session: Session) -> list[SpanUnion]:
-    """Flatten all spans across traces in start_time order."""
+    """Flatten all spans across traces in start_time order.
+
+    Timestamps are normalized to timezone-aware UTC before sorting so a mix of
+    naive and aware `start_time` values (produced by different mappers) can't
+    raise `TypeError` at construction time.
+    """
     spans = [span for trace in session.traces for span in trace.spans]
-    spans.sort(key=lambda s: s.span_info.start_time)
+    spans.sort(key=lambda s: _to_aware_utc(s.span_info.start_time))
     return spans
 
 
 def _span_text(span: SpanUnion) -> str:
-    """Full text content of a span, for search and retrieval."""
+    """Full text content of a span, for retrieval via get_span."""
     if isinstance(span, ToolExecutionSpan):
         return json.dumps(
             {
@@ -63,12 +79,45 @@ def _span_text(span: SpanUnion) -> str:
         )
     if isinstance(span, AgentInvocationSpan):
         return json.dumps(
-            {"user_prompt": span.user_prompt, "agent_response": span.agent_response},
+            {
+                "user_prompt": span.user_prompt,
+                "agent_response": span.agent_response,
+                "system_prompt": span.system_prompt,
+                "available_tools": [t.model_dump() for t in span.available_tools],
+            },
             default=str,
         )
     if isinstance(span, InferenceSpan):
         return json.dumps([m.model_dump() for m in span.messages], default=str)
     return json.dumps(span.model_dump(), default=str)
+
+
+def _search_haystack(span: SpanUnion) -> str:
+    """Plain-text rendering of a span for search matching.
+
+    Unlike `_span_text`, this joins the raw field values without JSON escaping so
+    a literal a judge copies from the overview (``$150``, ``refund_amount``) matches
+    the bytes it sees, and every human-visible field — including `system_prompt`
+    and `available_tools` — is searchable.
+    """
+    if isinstance(span, ToolExecutionSpan):
+        parts = [
+            span.tool_call.name,
+            json.dumps(span.tool_call.arguments, default=str),
+            str(span.tool_result.content),
+        ]
+        if span.tool_result.error:
+            parts.append(str(span.tool_result.error))
+        return "\n".join(parts)
+    if isinstance(span, AgentInvocationSpan):
+        parts = [span.user_prompt, span.agent_response]
+        if span.system_prompt:
+            parts.append(span.system_prompt)
+        parts += [f"{t.name}: {t.description or ''}" for t in span.available_tools]
+        return "\n".join(parts)
+    if isinstance(span, InferenceSpan):
+        return "\n".join(str(m.model_dump()) for m in span.messages)
+    return str(span.model_dump())
 
 
 def _preview(text: str, limit: int = _PREVIEW_CHARS) -> str:
@@ -81,17 +130,23 @@ def _describe(span: SpanUnion) -> str:
     if isinstance(span, ToolExecutionSpan):
         args = json.dumps(span.tool_call.arguments, default=str)
         result_size = len(str(span.tool_result.content))
-        return (
+        status = "ERROR" if span.tool_result.error else "ok"
+        line = (
             f"TOOL {span.tool_call.name}({_preview(args, 80)}) "
-            f"-> result: {result_size} chars: {_preview(str(span.tool_result.content))}"
+            f"-> [{status}] result: {result_size} chars: {_preview(str(span.tool_result.content))}"
         )
+        if span.tool_result.error:
+            line += f" | error: {_preview(str(span.tool_result.error), 80)}"
+        return line
     if isinstance(span, AgentInvocationSpan):
         return (
             f"AGENT prompt: {_preview(span.user_prompt, 80)} "
             f"-> response: {len(span.agent_response)} chars: {_preview(span.agent_response)}"
         )
     if isinstance(span, InferenceSpan):
-        return f"INFERENCE {len(span.messages)} messages"
+        size = sum(len(str(m.model_dump())) for m in span.messages)
+        preview = _preview(" ".join(str(m.model_dump()) for m in span.messages))
+        return f"INFERENCE {len(span.messages)} messages, {size} chars: {preview}"
     return f"{type(span).__name__}"
 
 
@@ -100,26 +155,37 @@ class TraceIndex:
 
     Attributes:
         session: The Session being evaluated.
-        max_read_chars: Cap on any single tool return, so a huge span can't
-            overflow the judge's context in one call. Oversized content is
-            windowed and the tool reports how to page through it.
+        max_read_chars: Cap on any single tool return, so a large span or a long
+            overview can't overflow the judge's context in one call. Oversized
+            content is windowed and the tool reports how to page through it.
     """
 
     def __init__(self, session: Session, max_read_chars: int = _DEFAULT_MAX_READ_CHARS):
+        if max_read_chars < 1:
+            raise ValueError(f"max_read_chars must be >= 1, got {max_read_chars}")
         self.session = session
         self.max_read_chars = max_read_chars
         self._spans = _flatten_spans(session)
+        # Precompute per-span overview lines once (recomputing per list_spans call
+        # is measurable on large sessions).
+        self._describe_lines = [f"[{i}] {_describe(span)}" for i, span in enumerate(self._spans)]
 
         # Bind instance state into plain functions so @tool sees clean signatures.
         # `this` (not `index`) so the public get_span(index=...) arg name is free.
         this = self
 
         @tool
-        def list_spans() -> str:
-            """List every span in the trace: one line per span with its index, type,
-            tool name, argument preview, and result size. Call this first to decide
-            which spans to inspect."""
-            return this.overview()
+        def list_spans(offset: int = 0) -> str:
+            """List spans in the trace: one line per span with its index, type, tool
+            name, argument preview, and result size. Call this first to decide which
+            spans to inspect. Long traces are paged; the response says how to page with
+            offset. Previews are truncated — load a span with get_span before you rely
+            on its content to score.
+
+            Args:
+                offset: Span index to start the listing from, for paging long traces.
+            """
+            return this.overview(offset)
 
         @tool
         def get_span(index: int, offset: int = 0) -> str:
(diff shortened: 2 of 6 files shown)


poshinchen:
@strandly-the-agent Review the changes and provide consolidate feedbacks. Also, I can see that the tool contains certain existing logic in strands-evals. What's your suggestion here?

strandly-the-agent:
**TL;DR — request changes.** The list/get/search shape is the right primitive (MLflow-style progressive disclosure), the diff is tight and additive, and the paranoid read is clean. But four verified defects make the judge score **silently wrong** — the failure mode this PR exists to eliminate — and #342's actual bug (overflow reported as `score: 0`) still ships unfixed. Details, repros, and suggested fixes below; answer to `poshinchen`'s overlap question at the end.

**Verified:** head `6017825`; `pytest tests/strands_evals/tools/` → 33 passed, 1 error (sandbox missing the `langfuse` extra; CI installs it — non-finding); every 🔴/🟡 below reproduced by running code, not eyeballing. Review pipeline: 6 independent passes (correctness, API design, tests, docs, LLM-context, issue-alignment) + aggregation; the adversarial pass failed to complete (infra), but its ground — regex stalls, paging loops, mixed-tz, tool-input coercion — was covered by the correctness pass and re-verified independently.

### 🔴 Blockers

1. **`list_spans` is unbounded — `max_read_chars` doesn't cap it, contradicting the docstring and README** (`trace_index.py:117-122`, `:164-168` vs `:103-105`, README). Measured: 2,000 spans → ~347K chars (~90K tokens); ~45 spans already exceeds the 8,000-char "cap on any single tool return". It's the judge's *first documented call*, on exactly the large sessions this feature targets — so the tool re-creates the overflow it was built to prevent. Fix: page `overview()` through `_window` with an `offset` param (same protocol as `get_span`) + a `showing spans A–B of M` header.
2. **`search_spans` can't find the literal text a judge quotes — grounded claims get scored as fabricated** (`trace_index.py:148`, `:54-71`). `pattern` goes straight into `re.compile`, and `$ ( ) . ? [ |` are *valid* regex, so the `re.error` fallback never fires: `search_spans('$150')` → `No matches` on a fixture shaped like this PR's own integ test (the flagship "$150 refund" case). Compounding: `_span_text` double-encodes nested-JSON tool results, so text copied from the overview (`"refund_amount": 150`) can't match the escaped bytes; and `'error'` false-positives on `"error": null` in every successful span. Your own test hand-escapes (`test_trace_index.py:96`) — the model gets no such hint. Fix: literal by default + opt-in `is_regex`, search an unescaped rendering, count per-span matches. (Bonus: a literal default also removes the regex-stall class — a valid nested-quantifier pattern measured 15.5s on one 35K-char span.)
3. **The judge can never see `system_prompt` or `available_tools` — through any channel** (`trace_index.py:64-68`, `:88-92` vs `types/trace.py:120-121`). `get_span` returns only `user_prompt`/`agent_response` despite promising "the full content"; `search_spans` shares `_span_text`, so `search_spans("never issue a refund")` → `No matches` even when it's in the system prompt. Any instruction-following or tool-selection rubric is *actively misled* — and it's a regression vs the inline path, which shows tool configs (`evaluator.py:158-180`). Fix: drop the special case and fall through to `model_dump()` (`:71`), or add the fields explicitly.
4. **The module docstring's headline example uses the one evaluator this pattern does not work with** (`trace_index.py:19-27`). `TrajectoryEvaluator` inlines `actual_trajectory` unconditionally (`case_prompt_template.py:51`) — reproduced: followed literally it raises; with a trajectory supplied, the 24K-char tool result is inlined anyway (25,915-char prompt vs a 195-char overview). The PR body admits this limitation; the docstring demonstrates it. Fix: switch the example to `OutputEvaluator` (as the README does) + an explicit limitation note.

### On the overlap with existing strands-evals logic (your question)

Six touchpoints; **one is worth fixing now because it's a bug, two deserve a design call, three only look similar — leave them alone.** No case for a big shared-helper refactor.

- **Fix now:** `_flatten_spans` (`trace_index.py:47-51`) duplicates `detectors/utils.py:57-59` `_flatten_traces_to_spans` character-for-character, and its added sort lacks the tz-normalization the repo already wrote (`extractors/trace_extractor.py:24-28` `_to_aware_utc`) — verified crash: mixed naive/aware `start_time` → `TypeError` at `TraceIndex()` construction, reachable via the openinference/langchain/adk mappers (root cause filed as #372). Minimum: `spans.sort(key=lambda s: _to_aware_utc(s.span_info.start_time))`. Better: promote a shared `flatten_spans()` + `to_aware_utc()` helper consumed by all three call sites — justified because it fixes a crash, not for tidiness.
- **Design call:** `detectors/chunking.py` already owns "session exceeds LLM context" with *token-accurate* budgeting (`estimate_tokens`, `would_exceed_context`) while `TraceIndex` budgets in chars — different strategy (split-and-reprompt vs pull-on-demand), both legitimate, but the repo shouldn't have two disagreeing answers to "how big is this trace for a model". Reusing `estimate_tokens` is the concrete first step. Also `tools/` has no `__init__.py` and no top-level export, so the README-blessed deep import becomes de-facto public API — either export it properly or stage under `experimental/` while the composition shape settles.
- **Leave alone:** `_span_text` vs `detectors/utils._serialize_spans` (same idiom, different field-selection needs), `Evaluator._format_*` (different input types and goals), mapper-side sorts (ingest normalization, different responsibility).

<details>
<summary><b>🟡 Should-fix (6)</b></summary>

- **Composition contract: judge context is smuggled through `actual_output`, and either half fails silently** (`README.md:209-222`, `tests_integ/…:127`). `actual_output` has other owners — compared to `expected_output`, string-matched by deterministic evaluators (`deterministic/output.py:16`), persisted in reports — so in an `Experiment` where evaluators share a Case, the injected overview corrupts them. And the two setup steps are uncoupled: overview without `tools=` → judge scores off 120-char previews; `tools=` without overview → judge scores the bare answer. Neither raises. Would one atomic composition point be better — `prompt_section, tools = index.for_judge()` now, or evaluator-side `OutputEvaluator(rubric=…, trace=index)` injected in `_build_prompt` (which would also let `TrajectoryEvaluator` adopt it later)? See Questions. Also: the README snippet assigns `evaluation_output` and stops — two undefined names, no wiring to `EvaluationData`.
- **`max_read_chars` unvalidated; `0` or a negative `offset` livelocks the judge** (`trace_index.py:108`, `:170-177`). Verified: `max_read_chars=0` → `[TRUNCATED: … call again with offset=0]`; `offset=-10` → empty window advising `offset=-10`. Fix: `ValueError` in `__init__` for `< 1`; reject negative offsets with an actionable message.
- **The overview hides tool failures and inference content** (`trace_index.py:83-94`). `_describe` never reads `tool_result.error` (populated by 5 mappers), so a `ConnectionError` renders as `-> result: 0 chars:`; `INFERENCE 2 messages` has no size/preview/tool names, so the judge fetches or skips blindly. Fix: `ok`/`ERROR` status + error preview on TOOL lines; size + preview on INFERENCE lines.
- **Nothing the *model* sees says previews are truncated or that claims must be verified before scoring.** The judge system prompt (`prompt_templates.py:1-12`) never mentions tools; the integ test only works because its rubric hand-carries "Verify claims against the trace evidence" (`tests_integ/…:44`) — the README rubric doesn't, so a README user gets a judge scoring off previews. Fix: put the guidance in the tool descriptions/overview header (the only strings guaranteed to reach the model), and state `max_read_chars` so the `N chars` counts become actionable.
- **`search_spans` caps and counts wrongly, unsignalled** (`trace_index.py:151-160`): one excerpt per matching *span* (docstring says per match), hard stop at 20 with no truncation marker (`get_span` has one), `IGNORECASE` undocumented, `max_matches=0` returns 1 hit. Folds into blocker 2's rewrite.
- **The suite doesn't pin the behaviours the feature sells — mutation score 11/27 killed (41% survive).** Two tautologies (`test_trace_index.py:109` — every possible return passes; `:130` — a bare `lambda` passes); paging past page 1 unproven (the scripted loop computes `offset += max_read_chars` instead of following the returned hint, so a broken hint stays green); `max_matches` untested; the `:50` sort is deletable; every overview number unasserted. Integ A/B: the inline control arm is computed and logged but **never asserted**, the fixture is ~65K tokens (33% of a 200K window — can't overflow, so the A/B shows nothing), paging never fires (spans 6.5K < 8K default), and absolute thresholds on n=1 with an unpinned model are the flake source (`tk_grounded > tk_fabricated` is the durable assertion). Priority list available if useful.

</details>

<details>
<summary><b>Questions</b></summary>

**Blocking**
1. **Should #342 item (1) land first, on its own?** `Experiment._run_evaluator` still records overflow as `score: 0, test_pass: False`, and `detectors/utils.py:33-54` `_is_context_exceeded` already implements exactly the detection #342 describes — a handful of lines of reuse. That's the silent-corruption bug; this PR's capability only reaches users who already know they have the problem (no in-library consumer: `grep -rn TraceIndex src/` matches only the module itself).
2. **Composition shape** — caller-side-but-atomic (`index.for_judge()`) now, or evaluator-side (`trace=` param at prompt-build time) as the target? Either beats routing judge context through `actual_output`. Related: was #324's `tools=` intended as the extension point for judge context?
3. **Is `strands_evals.tools` meant to be public API** (no `__init__.py`, absent from top-level `__all__`, README deep-imports it)? Export properly, or stage under `experimental/` (the redteam precedent) while 1–2 settle? Given a new public primitive and no `needs-api-review` label in this repo, the `design` label seems like the right flag.
4. **Scope clarity:** suggest stating in the PR body's *first* line "implements #342 item (2) only; item (1) and the trajectory path remain open", and a keep-open note on #342 — the body currently reproduces #342's entire bug up top while both deferrals sit far below, which invites a skimming reader to believe the bug is fixed. ("Addresses" is not a closing keyword, so #342 will mechanically stay open — the risk is human.)

**Non-blocking**
5. Chars or tokens for `max_read_chars`? (`chunking.py:143-146` documents compact JSON underestimating 30–50% vs what's sent.)
6. Naming: `TraceIndex` indexes a `Session`, not a trace — `SessionIndex`? Cheap now, breaking later. (`list_spans`/`get_span`/`search_spans` are good names; keep.)
7. Is `_span_text` dropping `span_info` (timestamps, durations, span_id) deliberate context economy or an oversight? Latency/ordering rubrics need at least duration.
8. The 200-trace corpus / 1,344-cell matrix / accuracy numbers in the body aren't reproducible from anything in-tree — no in-tree artifact even reaches the overflow threshold. The mechanism is real; suggest labelling those numbers as external motivation so they carry the right weight.

</details>

<details>
<summary><b>Appendix — non-blocking (11)</b></summary>

- ⚪ Multi-trace sessions interleave with no trace/turn marker (`trace_index.py:47-51`, `:167`) — "in the second turn…" rubrics unanswerable; prefix `(trace N)` when >1 trace.
- ⚪ `overview()`/`_span_text` recomputed per call (0.77s per `list_spans` on 2,000×8KB spans); build once in `__init__`.
- ⚪ Empty session → `ERROR: index 0 out of range (0..-1)` (`:134`) — inverted range in model-facing text.
- ⚪ Unbounded regex work: valid nested-quantifier patterns stall a worker thread (measured ~4× per size doubling, 15.5s at 35K chars); blocker 2's literal default removes the class.
- ⚪ Paged content lacks a position header — add `[span content chars A–B of N]`.
- ⚪ `.tools` is a mutable list, no order/identity contract; two indexes on one judge collide silently (identical tool names) — tuple + "one index per judge" note.
- ⚪ `trace_index.py:13-15` "pattern skills use" — internal jargon with no referent here; the MLflow comparison from the PR body is the better anchor.
- ⚪ Model-facing tool-call vocabulary now has a fourth spelling (`TOOL …` vs `Action:`/`Tool:` and `Tool call:`/`Tool result:` in `evaluator.py`).
- ⚪ `tests_integ/…:48-57` fires a live STS call at *collection* time and `except Exception` silently skips; STS success ≠ Bedrock access — gate on an env var in a fixture like `test_cloudwatch_provider.py:40`. Also `test_trace_index_patterns.py:126`'s defensive `pytest.skip` can make a real conversion regression vanish, and `:123` reaches into private `_convert_observations`.
- Pre-existing, filed **#373**: `OutputEvaluator(tools=…)` breaks `Experiment.to_file()` (`TypeError: DecoratedFunctionTool is not JSON serializable`) — from #324, not this PR; but this PR's README is the first to document the crashing path, so one caveat sentence there is warranted.
- Pre-existing, filed **#372**: `SessionMapper.parse_timestamp` returns naive datetimes despite promising aware UTC — the upstream source of the mixed-tz crash; the one-line `_to_aware_utc` fix in this PR stands regardless.

Also checked, clean: no network/env/subprocess/telemetry anywhere in the diff — `trace_index.py` imports only `json`, `re`, `strands.tool`, and repo types; the only network call in the PR is the intended Bedrock judge invocation in `tests_integ/`. Non-int/non-str tool inputs are safely rejected through the real strands invocation path. Docstring style matches repo precedent.

</details>

*I'm an AI reviewer — treat this as prepared input for a human decision, not a gate. Happy to re-review on update.*

pdebjyot:
Thanks — this was an unusually useful review; every 🔴 reproduced against the actual code, and the fixes are pushed. Point by point:

**🔴 Blockers — all fixed**

1. **`list_spans` unbounded.** Fixed. `overview()` (and `list_spans`) now page by span index through `max_read_chars`: a `Showing spans A–B of M` banner, a `[MORE: … offset=N]` continuation marker, and `list_spans(offset=…)` to follow it — same protocol as `get_span`. The first documented call can no longer overflow. New tests walk the pages and assert every span appears exactly once.

2. **`search_spans` can't find quoted literals.** Fixed. Search is **literal by default** with opt-in `is_regex=True`. It now matches against an *unescaped* rendering of each span, counts matches per span, signals truncation at `max_matches`, and an invalid regex returns an actionable error instead of a silent miss. `search_spans("$150")` and `search_spans("refund_amount=$150")` both hit now; `"error"` no longer false-positives on `"error": null` because the haystack no longer carries JSON keys. (This also removes the regex-stall class you measured.)

3. **`system_prompt` / `available_tools` invisible.** Fixed. Both `get_span` (via `_span_text`) and search (via the new haystack) now include `system_prompt` and `available_tools` for `AgentInvocationSpan`. `search_spans("Never issue a refund")` finds the system prompt.

4. **Docstring example used the wrong evaluator.** Fixed. The module docstring and README now use `OutputEvaluator` and carry an explicit note that `TrajectoryEvaluator` inlines `actual_trajectory` unconditionally and so does not compose with this pattern.

**Overlap — the one that was a bug**

- **`_flatten_spans` tz crash.** Fixed with the minimal change you suggested: sort via the repo's existing `_to_aware_utc`, so mixed naive/aware `start_time` no longer raises `TypeError` at construction. New test covers it. I've left the shared-helper promotion and the `chunking.py` token-budget reconciliation as follow-ups rather than folding a refactor into this PR — agree they're worth a separate design call, not tidiness churn here.

**🟡 Should-fix**

- `max_read_chars < 1` now raises `ValueError`; negative offsets are rejected with an actionable message (both `get_span` and `overview`).
- Overview now surfaces tool `ok`/`ERROR` status + error preview, and INFERENCE size + preview.
- `search_spans` truncation marker + per-span counts (folded into blocker 2).
- Overview header now states previews are truncated and that claims must be verified with the tools before scoring, and names `max_read_chars` — the guidance lives in strings the model is guaranteed to see.
- Tests strengthened: the tautologies are gone, paging follows the returned hint rather than a hand-computed offset, and the new behaviors (literal/regex, counts, truncation, tz, validation, tool errors) are pinned.

**Composition contract / public-API / scope questions**

- **Scope (Q4):** done — the PR body now leads with "implements #342 item (2) only; item (1) and the trajectory path remain open," and #342 stays open.
- **Composition (Q2):** added `index.for_judge()`, which returns `(prompt_section, tools)` atomically so the "either half fails silently" footgun is gone — the docstring, README, and example all use it now. The deeper evaluator-side option (`OutputEvaluator(rubric=…, trace=index)` injected in `_build_prompt`, which would also let `TrajectoryEvaluator` adopt it) is the cleaner long-term shape but a bigger public-API commitment; I'd rather land `for_judge()` now and take the evaluator-side param as a follow-up with your call on it. The residual `actual_output` concern is real — `for_judge()` fixes the coupling, `trace=` would remove the smuggling entirely.
- **Public-API staging (Q3):** genuinely your call — export `strands_evals.tools` properly, or stage under `experimental/` (the redteam precedent) while the composition shape settles. Point me either way and I'll follow up.

I left the pre-existing #372/#373 as-is (out of scope), and the non-blocking appendix items noted for follow-up. Full suite 1763 passed, ruff clean. Re-review welcome.

poshinchen:
@strandly-the-agent review the updated changes.

strandly-the-agent:
**TL;DR — all four 🔴 blockers are fixed and verified at `44b6c1f`; two residual 🟡 in the search rewrite are worth one more small commit; the rest is maintainer design calls.** I re-ran every original repro against the new head rather than trusting the commit message — ledger below.

### Status of my previous findings

| Finding | Status |
|---|---|
| 🔴 `list_spans` unbounded | **Fixed & verified** — paged through `max_read_chars`; I walked a 2,000-span session by following only the returned `[MORE: … offset=N]` hints: 48 pages, max 8,008 chars/page, every span exactly once, terminates. New test follows the hint too (`test_trace_index.py:83-98`) — the old hand-computed-offset weakness is gone. |
| 🔴 `search_spans` regex-eats literals | **Fixed & verified** — `search_spans("$150")` and `"refund_amount=$150"` both hit; `"error"` no longer false-positives on tool spans; invalid regex returns an actionable error. **But two residuals below.** |
| 🔴 `system_prompt`/`available_tools` invisible | **Fixed & verified** — both reachable via `get_span` and searchable (`"Never issue a refund"` → hit). |
| 🔴 docstring example uses `TrajectoryEvaluator` | **Fixed** — `OutputEvaluator` + explicit limitation note in docstring and README. |
| 🟡 mixed-tz construction crash | **Fixed & verified** — sort now goes through `_to_aware_utc`; my original repro no longer raises; regression test added. |
| 🟡 `max_read_chars=0` / negative offset livelock | **Fixed & verified** — `ValueError` at construction; negative offsets rejected with actionable errors in both `get_span` and `overview`. |
| 🟡 tool errors / INFERENCE invisible in overview | **Fixed & verified** — `[ERROR] … | error: ConnectionError: refused` renders; INFERENCE lines carry size + preview. |
| 🟡 no model-facing verification guidance | **Fixed** — overview header + tool descriptions now carry it, and state `max_read_chars`. |
| 🟡 unit-test tautologies / unpinned behaviours | **Fixed** — both tautologies replaced; paging, counts, truncation, tz, validation all pinned (49 passed locally). |
| 🟡 composition footgun (Q2) | **Addressed** — `for_judge()` returns both halves atomically; docstring/README/tests use it. Evaluator-side `trace=` stays the cleaner long-term shape, as you noted. |
| 🟡 integ A/B (inline arm unasserted, fixture below overflow, STS at collection) | **Still open** — `tests_integ/` is unchanged in this delta. Fine as a follow-up, just don't read the A/B as evidence yet. |
| Q1 (#342 item 1 first), Q3 (export vs `experimental/`), tokens-vs-chars | **With maintainers** — on Q3, since `pdebjyot` asked: my suggestion is `tools/__init__.py` + top-level export rather than `experimental/` — `for_judge()` has settled the caller contract enough that a later evaluator-side `trace=` param can land without breaking these callers. |

### 🟡 New (residuals of the search fix — attack-the-fix findings)

1. **A phrase copied from an overview preview still misses when the original text has a newline** (`trace_index.py:95-120` vs `:124`). Previews collapse whitespace but the search haystack keeps raw `\n` — verified: content `"Refund approved.\nAmount: $150 total."` shows in the overview as `…approved. Amount…`, and `search_spans("approved. Amount")` → `No matches`. This is the one compounding cause from the original search blocker that didn't get fixed, and copy-from-preview is exactly how a judge forms queries. Suggestion: whitespace-normalize both haystack and needle the way `_preview` does (`re.sub(r"\s+", " ", …)`); costs exactness of offset math nothing since excerpts are re-previewed anyway.
2. **InferenceSpan haystack is `str(m.model_dump())` — Python repr with dict keys** (`trace_index.py:118-119`). Verified: `search_spans("content")`, `"text"`, and `"role"` all false-positive on every inference span (`[0] (3 matches) …{'role': <Role.USER: 'user'>, 'content': [{…`), and excerpts show repr noise. The `"error"`-artifact fix reached tool spans but not this branch. Suggestion: extract the message text content (role + text fields) for the haystack, keeping `model_dump` only in `_span_text` for `get_span`.

<details>
<summary><b>Appendix — non-blocking (5)</b></summary>

- ⚪ `is_regex=True` still allows catastrophic backtracking — `(a+)+$` against a single 3KB span hung >55s in my sandbox (`:215-220`). Opt-in now, so acceptable, but a one-line docstring warning (or a pattern-length cap) is cheap.
- ⚪ Empty session: `get_span(0)` still returns `ERROR: index 0 out of range (0..-1)` (`:199-200`) — the inverted-range message from my first review's appendix survives the rewrite.
- ⚪ `search_spans("")` matches every span with inflated counts (`:222-229`) — an explicit `ERROR: empty pattern` would be kinder to a confused judge.
- ⚪ `max_matches=0` short-circuits to `No matches for …` (`:233,:245-246`) — misleading, though nobody sane passes 0.
- ⚪ `for_judge()` returns `self.tools` by reference (`:304`) — caller mutation mutates the index; a tuple or copy would pin the contract. Also the `[MORE: … call again with offset=N]` marker inside the *prompt-embedded* overview reads as "call again" without naming `list_spans` — one word (`call list_spans with offset=N`) removes the ambiguity.

</details>

<details>
<summary><b>Evidence ledger / review shape</b></summary>

- Head reviewed: `44b6c1f` (delta `6017825..44b6c1f`, +414/−70 across `trace_index.py`, `test_trace_index.py`, README, one integration-test tweak).
- `pytest tests/strands_evals/tools/` → **49 passed, 1 error** — the error is the known `langfuse`-extra gap in my sandbox (CI installs it), same as last round. `ruff`/full-suite claims not re-run here (no hatch in sandbox); CI is the arbiter for those.
- 11-repro battery: every original 🔴/🟡 repro re-executed against the new head (results in the table above), plus new probes: 2,000-span page-walk via returned hints; `max_read_chars=1` floor (bounded ~244-char return, terminates); Unicode case-folding (`strasse`→`STRASSE` hits); cross-field newline joins (no space-vs-`\n` false positive across fields); mixed naive/aware construction.
- Review shape: per-fix verification + attack-the-fix run by me first-party (this is a fixes-only delta responding to my review); an additional independent adversarial pass was still in flight when I posted — if it surfaces anything material beyond the two 🟡 above I'll add a short follow-up, otherwise silence means it agreed.

</details>

*I'm an AI reviewer — treat this as prepared input for a human decision. With the two 🟡 addressed (or explicitly deferred), I have nothing further blocking; the open design calls (Q1/Q3, tokens-vs-chars) are `poshinchen`'s to make.*

strandly-the-agent:
**Addendum — the independent adversarial pass finished, and three things are material beyond my comment above** (it otherwise agreed: paging survived a 300-session fuzz with zero exactly-once violations, `get_span` reconstruction is byte-exact, and the tz fix holds for naive + UTC + `+05:00` mixes).

1. 🟡 **`search_spans` is the one tool `max_read_chars` doesn't bound — and its own truncation marker tells the judge to blow past it** (`trace_index.py:231-249` vs the claim at `:158`/README). Measured on an 800-span session with `max_read_chars=8000`: `max_matches=100` → 12,614 chars, `max_matches=800` → **101,269 chars** (13× the cap) — while `overview()` and `get_span` on the same session stay ≤8K. Hitting the default 20-span cap is normal on a long trace, and the marker says "raise max_matches for more". Suggestion: stop appending once the accumulated return would exceed `max_read_chars` and say "budget reached", rather than trusting `max_matches` alone.
2. 🟡 **My residual #2 is worse than I stated — on real mapper output, error search is fully inverted.** Reproduced on the repo's own ADK fixture (`adk_live_spans.json`, no failed tools): `search_spans("error")` → false positive via `'error': None` inside an inference span's repr. And in a session with a genuinely failed tool, the real failure is **unfindable**: `"error"`, `"ERROR"`, and `"[ERROR]"` (copied from the overview) all miss it, because the tool haystack carries only the error's *value* (`CardDeclined: insufficient funds`). A judge asking "did any tool fail?" gets 100% false positives and misses the one real failure. Reaches any ADK/CloudWatch/LangChain-mapped session (their inference messages carry tool content). Suggestion: render inference messages from fields (role + text + tool name/args/result) like the other two branches, and emit a searchable `error:` token in the tool branch so the overview's `[ERROR]` vocabulary and the search index speak the same language.
3. 🟡 **Anchored regexes silently never match** — the haystack is newline-joined but `is_regex=True` compiles without `re.MULTILINE`, so `"^Refunded"` / `"…TKT-1$"` → `No matches` (`:215-220`). LLMs write anchored regexes, and the judge reads that as "claim unsupported". One flag fixes it.

<details>
<summary>Smaller items + test gaps the pass found (6)</summary>

- ⚪ False `[stopped at N spans…]` marker when the scan ended exactly on the last match (`:233,:247`) — `truncated` is set before checking whether anything further matches; wastes a judge turn.
- ⚪ Literal-match offsets are computed on `text.lower()` but sliced from `text`, so excerpts drift (and can come back empty) when U+0130 `İ` precedes the match — Turkish traces. Cleanest fix also simplifies the code: replace the hand-rolled matcher with one `re.compile(re.escape(pattern), re.IGNORECASE)` path, whose offsets are always valid.
- ⚪ Overview `result: N chars` counts raw text while `get_span` returns `json.dumps(… ensure_ascii=True)` — non-ASCII content under-reports ~6×, so the judge's paging estimate is wrong (degrades gracefully via the TRUNCATED marker). Report `len(_span_text(span))` instead.
- ⚪ `max_read_chars` is a soft cap at tiny values (header counted; `Showing…` line and markers not; first line appended unconditionally): `max_read_chars=200` → 428-char page. ~1% slippage at the default — fine — but worth one docstring word ("approximate").
- Test gaps: `test_search_does_not_false_positive_on_serialization_artifacts` has **no `InferenceSpan`** in its fixture — reverting the entire inference-haystack fix passes 25/25; "Matching is case-insensitive" is unpinned (dropping both `.lower()` calls passes); the negative-offset assertion at `test_trace_index.py:172` is vacuous (the error string contains no `offset=`, so the guarded assert always takes the `True` branch); `:314`'s `assert tools is index.tools` locks in the live-list aliasing my earlier appendix flagged — prefer `==` + a copy.
- Staleness note: the index snapshots the Session at construction (consistently — fuzzing found no corruption); one docstring line saying so would complete the "read-only" claim.

</details>

Repro script + full output are uploaded as review artifacts (`adversarial-repros.py`, `adversarial-repro-output.log`). With #1/#2 folded into the same small commit as the two residuals above, I have nothing further — the verdict from my previous comment stands.

poshinchen:
@strandly-the-agent why can't this be a built-in functions when the context / traces are too large? It seems like having these as tool / plugin is not ideal? Investigate the approaches thoroughly.

strandly-the-agent:
**TL;DR — it can and should be built in. The PR conflates two things: the *mechanism* (an index the judge LLM queries through `list_spans`/`get_span`/`search_spans`) and the *wiring* (the caller composes it via `for_judge()` + `tools=` + pasting the overview into `actual_output`). The tool-calls part is intrinsic — a judge can't read what isn't in its context, so anything that isn't lossy truncation or N×-cost chunking has to give the judge retrieval, and retrieval means tools on the judge agent. The plugin/caller-composed part is the design choice, and it's the wrong one: it can never reach `TrajectoryEvaluator` or the built-in trace-level judges (Helpfulness, Faithfulness, …), it smuggles judge context through a data field, and it only helps users who already know they have the problem. The repo already has the exact built-in pattern to copy — `detect_failures` preflights with `would_exceed_context`, falls back on `_is_context_exceeded`, and the SDK can't rescue a single oversized message on its own.**

**Recommendation:** keep `TraceIndex` as an *internal* engine; add automatic disclosure in the evaluator base — inline exactly as today when the trace fits, substitute the paged overview + attach the tools to the judge `Agent` only when it wouldn't — and land #342 item (1) (report overflow as an error, not `score: 0`) first and separately. Concretely for this PR: re-scope so the README stops advertising `strands_evals.tools.trace_index` + `for_judge()` as the user story, and either add the ~60-line base-class wiring here or stage the engine under `experimental/` until the wiring PR lands. Details below.

<details>
<summary><b>What is built in today — and where oversized traces currently go (verified)</b></summary>

- **Every judge inlines unconditionally, no size check anywhere.** `compose_test_prompt` writes `<Trajectory>{actual_trajectory}</Trajectory>` (`prompt_templates/case_prompt_template.py:48-51`); `OutputEvaluator._build_prompt` writes `<Output>{actual_output}</Output>` (`output_evaluator.py:45-56`); the trace-level judges format the whole session history line-by-line (`evaluator.py:224-244 _format_trace_level_prompt`, `:182-192 _format_session_history`) and hand it to `Agent(...)` with no tools (`helpfulness_evaluator.py:64-65`). `grep -rn would_exceed_context src/strands_evals/evaluators` → nothing.
- **The SDK can't save a one-shot prompt.** `Agent` catches `ContextWindowOverflowException`, calls `conversation_manager.reduce_context`, and retries (`strands/agent/agent.py:1678-1687`), but `SlidingWindowConversationManager` trims *older messages* / truncates tool results; with a single oversized user message it raises `"Unable to trim conversation context!"` (`sliding_window_conversation_manager.py:247,262`). So the judge's one big prompt has no SDK-level escape.
- **The overflow then becomes a quality score.** `Experiment._run_evaluator` catches every exception and records `test_pass: False, score: 0, reason: "Evaluator error: …"` (`experiment.py:447-456`) — #342 item (1), still open.
- **The repo already has a built-in pattern for exactly this problem — in `detectors/`, not `evaluators/`.** `detect_failures` preflights `would_exceed_context(user_prompt)` and routes to `_detect_chunked`, and *also* falls back reactively when the call raises and `_is_context_exceeded(e)` (`failure_detector.py:84-94`); the chunked path is `split_spans_by_tokens` → per-chunk model call → `merge_chunk_failures` (`:118-150`). `analyze_root_cause` has a three-tier direct → prune → chunk fallback (`AGENTS.md:302`). Token estimation is tiktoken-accurate via the SDK (`chunking.py:32-60`). None of this is wired into the judge evaluators.

</details>

<details>
<summary><b>Approach comparison</b></summary>

| | Approach | Reaches | Fidelity | Cost | Verdict |
|---|---|---|---|---|---|
| **A** | **Built-in automatic progressive disclosure** — base evaluator measures the rendered trajectory (`estimate_tokens`); fits → inline as today; doesn't → substitute paged overview into the slot the evaluator owns (`<Trajectory>`, `# Conversation History`, the trajectory part of `<Output>`) and add the discovery tools to the `Agent` the evaluator already builds | **all** judge evaluators incl. `TrajectoryEvaluator`, trace-level, multimodal | high — judge pulls exact spans on demand; no lossy cut | tool round-trips **only when the alternative is a guaranteed failure**; one `estimate_tokens` per eval (detectors already pay this) | **Recommended** |
| **B** | Built-in chunk-and-aggregate (reuse `split_spans_by_tokens`) | all | **lossy for grounding**: claim in chunk 3, evidence in chunk 1; and "aggregate a 0–1 rubric score across chunks" has no clean semantics (mean? min? any-fail?) | N× judge calls | Right for per-span outputs (detectors merge failure *lists* naturally); wrong for one holistic score. Not the default |
| **C** | Built-in truncate/summarize | all | lossy in exactly the wrong direction for grounding rubrics (the one span that supports/contradicts a claim is what gets cut — #342's argument) | 1 call | Last-resort only, and must be flagged in `reason` |
| **D** | Status quo: caller-composed plugin (`for_judge()` + `tools=` + paste into `actual_output`) | `OutputEvaluator` family only; **never** `TrajectoryEvaluator` (`case_prompt_template.py:51`) or trace-level judges (no `tools=`, `helpfulness_evaluator.py:65`) | high when wired; **silently degraded when half-wired** | — | Keep at most as an escape hatch for custom/non-Strands judges; not the product story |
| **E** | Fix the reporting first — `_is_context_exceeded` → distinct error status in `_run_evaluator` instead of `score: 0` | all | — | ~10 lines | **Land first, regardless** |

Why D isn't ideal, in the maintainer's terms: discoverability (deep import, no export, no in-library consumer — `grep -rn TraceIndex src/` hits only the module itself), a two-step footgun (each half fails silently), and `actual_output` has other owners (compared to `expected_output`, string-matched by deterministic evaluators, persisted in reports) so the overview corrupts sibling evaluators sharing a Case.

</details>

<details>
<summary><b>Minimal built-in shape (A + E) — sketch and touch points</b></summary>

Mirror `failure_detector.py:84-94` exactly: preflight, plus reactive fallback.

```python
# evaluators/evaluator.py (base) — one helper
def _disclose(self, session: Session, inline: str) -> tuple[str, list]:
    """Inline when it fits; otherwise paged overview + discovery tools."""
    if self.disclosure == "never" or not would_exceed_context(self.system_prompt + inline):
        return inline, []
    section, tools = _TraceIndex(session).for_judge()   # internal engine
    logger.info("trajectory exceeds judge context; using progressive disclosure")
    return section, list(tools)
```

```python
# each judge, at the Agent construction it already has (output_evaluator.py:74, helpfulness_evaluator.py:65, trajectory_evaluator.py:73, multimodal_*)
prompt, disclosure_tools = self._disclose(session, rendered)
agent = Agent(model=self.model, tools=[*(self.tools or []), *disclosure_tools], system_prompt=self.system_prompt, ...)
try:
    result = agent(prompt, structured_output_model=EvaluationOutput)
except Exception as e:                       # reactive path, like failure_detector.py:87-94
    if not disclosure_tools and _is_context_exceeded(e):
        prompt, disclosure_tools = _TraceIndex(session).for_judge()
        result = Agent(..., tools=[*(self.tools or []), *disclosure_tools])(prompt, ...)
    else:
        raise
```

- **Slots:** `compose_test_prompt` gets an optional pre-rendered trajectory (or the caller passes the overview in place of `actual_trajectory`) so `<Trajectory>` is the substitution point for `TrajectoryEvaluator`; `_format_trace_level_prompt` substitutes the `# Conversation History` block for trace-level judges; `OutputEvaluator` only when a `Session` is present on the case.
- **Public API:** none required for default-on-overflow behaviour. One optional knob, `disclosure: Literal["auto", "always", "never"] = "auto"` on the base, covers opt-out and A/B testing. No behaviour change for anything that fits today.
- **Observability:** when disclosure triggers, say so in `EvaluationOutput.reason`/result metadata so a user knows the judge read via tools.
- **What `TrajectoryEvaluator` loses:** nothing — its deterministic scorers take the small tool-name lists the judge extracts; those fit and stay inline.
- **Cost when it triggers:** a few tool round-trips versus a certain overflow → `score: 0`. Not a trade-off.
- **E, independently:** `_run_evaluator` → `if _is_context_exceeded(e): status="error"` (or `score=None`) instead of `score: 0`; `detectors/utils.py:33-54` already has the predicate.

</details>

<details>
<summary><b>What this means for PR #343 and why it wasn't built in</b></summary>

- `pdebjyot`'s stated reason (08-27) was that evaluator-side integration is "a bigger public-API commitment" and #324's `tools=` was the extension point available. Fair as a scoping call, but the analysis above says the evaluator-side shape is *smaller* in public surface (one optional knob vs a new documented class + method + import path) and larger in reach.
- The engine in this PR is now solid (paging fuzzed, search fixed) and would be the internal implementation of A essentially unchanged — `for_judge()` is exactly the hook `_disclose` needs. What should change: it becomes private (`_trace_index.py` or an underscore class), the README section goes (or moves to an "advanced: custom judges" note), and the composition moves into the evaluators.
- Two ways to land it: (1) re-scope this PR to engine + base-class wiring (`OutputEvaluator` + `TrajectoryEvaluator` + trace-level base, ~60–100 lines) — my preference, since the engine is ready and the wiring is where the value is; or (2) merge the engine under `experimental/` now with no public story, and do the wiring as the next PR. Either way, E first.

</details>

*Verified against `05c5f2a` and `main`; the SDK behaviour is from the installed `strands-agents` in my sandbox. An independent design pass is still running — I'll add a note only if it disagrees materially.*

strandly-the-agent:
**Addendum — the independent design pass agrees with the recommendation above, and corrects one detail in my sketch:**

1. **Scope correction.** `OutputEvaluator` never inlines the trajectory — `compose_test_prompt` only writes `<Trajectory>` when `uses_trajectory=True` (`case_prompt_template.py:48-51`), and `OutputEvaluator` calls it without that flag (`output_evaluator.py:56-61`). Its overflow comes from a huge `actual_output` *string*, which an index over a `Session` can't help with. So automatic substitution honestly applies to **`TrajectoryEvaluator`'s `<Trajectory>` slot + the 15 trace/tool-level judges** that render via `_format_trace_level_prompt` / `_format_session_history` / `_format_tool_level_prompt` (all built with `Agent(...)` and no tools today) — which is exactly where the silent overflow lives. For `OutputEvaluator`, built-in disclosure is *additive* (it needs `actual_trajectory=Session` on the case to have anything to disclose) and should stay opt-in (`disclosure="always"`). Strengthens the case for A rather than weakening it: the evaluators D can never reach are the ones that need it most.
2. **Measured magnitude.** A 41-span session with ~1.2 MB of tool results renders to **407,203 tokens** inline (`would_exceed_context` → True) vs **2,756 tokens** as a `TraceIndex` overview — 148×. Probe + log uploaded as review artifacts (`judge-context-probe.py/.log`).
3. **Two things to know when reusing the detectors' preflight:** (a) `would_exceed_context` budgets against `DEFAULT_MAX_INPUT_TOKENS=200_000 × PREFLIGHT_SAFETY_MARGIN=0.65` → effective ~130K (`detectors/constants.py:8-11`), so an evaluator-side version should take the judge model's window rather than hard-code it; the `chunking.py:69,123` docstrings still cite 0.85/0.70 — stale vs `constants.py`, pre-existing. (b) The SDK's only automatic lever is `_truncate_tool_results` — an oversized payload *inside a toolResult block* gets middle-out truncated (4,000,000 → 452 chars, verified), which is approach C's fidelity; a plain-text user message just raises. Neither changes the recommendation.

poshinchen:
@strandly-the-agent your answer is too tedious: https://github.com/strands-agents/evals/pull/343#issuecomment-5637213617. Summarize it, be neat, simple and clear to understand. What's the recommendation and what will the devx look like. What are the recommended changes?

strandly-the-agent:
Fair — here's the short version.

**Recommendation:** make it built in. Evaluators should handle oversized traces themselves; users shouldn't have to know `TraceIndex` exists.

**DevX — today (this PR):**
```python
index = TraceIndex(session)
prompt_section, tools = index.for_judge()
evaluator = OutputEvaluator(rubric=..., tools=tools)
evaluator.evaluate(EvaluationData(input=..., actual_output=f"{answer}\n{prompt_section}"))
```
**DevX — proposed:**
```python
evaluator = TrajectoryEvaluator(rubric=...)   # or HelpfulnessEvaluator(), etc. — unchanged
evaluator.evaluate(case)                       # trace fits → inlined as today
                                               # trace too big → judge gets a paged overview + list/get/search tools, automatically
```
No new imports, no new class to learn. One optional knob if wanted: `disclosure="auto" | "always" | "never"` (default `auto`). Behaviour for traces that fit today is unchanged.

**Recommended changes, in order:**
1. **Fix the reporting (#342 item 1), separate small PR.** In `Experiment._run_evaluator`, a context overflow becomes an *error* result, not `score: 0`. The predicate already exists (`detectors/utils.py:_is_context_exceeded`).
2. **Make `TraceIndex` internal** — keep the code (it's solid now), drop the README user story and the public `strands_evals.tools.trace_index` import.
3. **Wire it into the evaluators** — in the base class: measure the rendered trace with the existing `would_exceed_context`; if it fits, inline as today; if not, substitute the paged overview and add the three tools to the judge `Agent` the evaluator already builds. Applies to `TrajectoryEvaluator` and the trace/tool-level judges (Helpfulness, Faithfulness, …) — the ones that overflow silently today and that the current plugin approach can never reach. Same preflight-then-fallback pattern `detect_failures` already uses.

**For this PR specifically:** either re-scope it to (2)+(3), or merge just the engine under `experimental/` and do (3) next. Not as a public plugin.

pdebjyot:
Agreed — this should be built in, not a plugin the caller wires up. Confirming the plan and re-scoping accordingly.

**The split:** the *mechanism* (the judge reads a large trace by calling `list_spans`/`get_span`/`search_spans`) is intrinsic — a judge is an LLM and can't score what isn't in its context, so any non-lossy path has to give the judge retrieval tools. What's wrong is the *wiring*: making the caller compose `for_judge()` + `tools=` + paste the overview into `actual_output`. That never reaches `TrajectoryEvaluator` or the built-in trace/tool-level judges (Helpfulness, Faithfulness, …) — which are exactly the ones that overflow silently today — and it only helps users who already know they have the problem.

**Plan (three steps, in order):**

1. **Fix the reporting first — separate small PR (#342 item 1), now up as #399.** In `Experiment._run_evaluator`, a context overflow becomes a distinct *could-not-evaluate* result instead of `score: 0 / test_pass: False`, so an overflow is no longer indistinguishable from a genuine quality failure and can't silently drag down `overall_score` or pass rate. It gates strictly on the existing `_is_context_exceeded` predicate, so a genuine evaluator bug still surfaces as a failure rather than being hidden. This is worth landing on its own regardless of the disclosure work.

2. **Make `TraceIndex` internal.** Keep the engine (it's solid now), drop the public `strands_evals.tools.trace_index` import and the README user story so we're not committing to a plugin surface.

3. **Wire it into the evaluators.** In the judge-building path, preflight the rendered prompt with the existing `would_exceed_context` (taking the judge model's window, not a hard-coded 200K); if it fits, inline exactly as today; if not, substitute the paged overview and attach the three tools to the judge `Agent` the evaluator already constructs. Applies to `TrajectoryEvaluator` + the trace/tool-level judges.

**DevX — unchanged for callers:**
```python
evaluator = TrajectoryEvaluator(rubric=...)   # or HelpfulnessEvaluator(), etc.
evaluator.evaluate(case)   # trace fits → inlined as today
                           # trace too big → judge auto-gets overview + list/get/search tools
```
One optional knob if you want it: `disclosure="auto" | "always" | "never"` (default `auto`). Behaviour for traces that fit today is unchanged.

**One correction to my earlier sketch:** `OutputEvaluator` doesn't inline the trajectory (`compose_test_prompt` only writes `<Trajectory>` when `uses_trajectory=True`, which it doesn't set) — its overflow comes from a large `actual_output` string, which a `Session` index can't help. So automatic disclosure honestly targets `TrajectoryEvaluator` + the trace/tool-level judges; for `OutputEvaluator` it's additive and stays opt-in.

**For this PR:** (1) is already up as #399. I'll re-scope this PR to (2)+(3) — stage the engine as internal and add the base-class wiring. Sound right before I push the changes?

poshinchen:
@strandly-the-agent review the changes based on the discussions and proposed structure.

strandly-the-agent:
**TL;DR — there's nothing new to review yet.** The PR head is still `05c5f2a` (Sep 3, the commit I reviewed last round); the re-scope `pdebjyot` laid out on Sep 14 hasn't been pushed — his comment ended with *"Sound right before I push the changes?"* and is waiting on a yes.

From my side: **yes, the plan is right** — it matches what was agreed (engine internal, wiring in the evaluator base, #399 first). `pdebjyot`, go ahead. What I'll check the push against is below so it isn't a surprise.

One heads-up: a `disclosure=` kwarg on the judge evaluators is new public surface, so the re-scoped PR should get `needs-api-review` when it lands (the current PR has no API label).

<details>
<summary><b>What the re-scoped push should contain (review checklist)</b></summary>

1. **Engine goes internal.** Move `tools/trace_index.py` → e.g. `evaluators/_trace_index.py` (or `experimental/`); drop the README section and the public `for_judge()` story. `tests_integ/test_trace_index_judge_reliability.py` can go too or stay as a follow-up — it's still unasserted on the inline arm.
2. **Wiring lives in the base class, once.** 21 evaluator files build their own `Agent(...)` (e.g. `trajectory_evaluator.py:72,94`, `helpfulness_evaluator.py:65,81`), so the preflight must be a single base-class helper, not per-evaluator copy-paste. Preflight with the existing `would_exceed_context(rendered_prompt, max_input_tokens=…)` — pass the judge model's window rather than the detector default (`DEFAULT_MAX_INPUT_TOKENS × 0.65`). Fits → prompt is byte-identical to today. Doesn't → overview replaces the trajectory slot (`_format_trace_level_prompt` `evaluator.py:254`, `_format_session_history` `:202`, `_format_tool_level_prompt` `:214`, and `TrajectoryEvaluator`'s `<Trajectory>`) and the three tools go on that `Agent`.
3. **Tests prove the switch, not the mock.** (a) fits-case: rendered prompt equals today's, no tools on the agent; (b) overflow-case: overview + exactly three tools; (c) `disclosure="never"` restores today's behaviour so the overflow reaches #399's not-applicable path; (d) `disclosure="always"` on `OutputEvaluator` with `actual_trajectory=Session`.
4. **The 4 open inline threads still apply once internal** — especially the whitespace-normalize one (`trace_index.py:150`): the judge copies phrases from the overview preview regardless of who wired the tools. Empty-pattern and empty-session messages are two-liners. The `for_judge()` by-reference thread becomes moot if the method goes private.
5. **Ordering.** #399 (`c954fe3`) implements step 1 as agreed and is independent of this PR; it should merge first. I've only skimmed it (gates on `_is_context_exceeded`, emits a `NOT_APPLICABLE` row) — not a review.

</details>

*I'm an AI reviewer — this is prepared input for a human decision.*

poshinchen:
@strandly-the-agent review the changes based on the discussions and proposed structure.

strandly-the-agent:
**TL;DR — request changes, but this push is the agreed structure and it's close.** Checklist items 1–4 from my Sep 16 comment all hold at `a8e0a87` (verified by running code, not reading the commit message); item 5 (#399) is still open. One new 🔴: the tool-level judges preflight the *smallest* row, so multi-turn sessions still overflow on later rows — the exact silent `score: 0` this feature exists to remove. One-line fix. Also: ~18 evaluators gained a public `disclosure=` kwarg and the PR still has no `needs-api-review` label.

| Sep 16 checklist | Status at `a8e0a87` |
|---|---|
| 1. Engine internal | ✅ `tools/trace_index.py` → `evaluators/_trace_index.py`; no import of the old path anywhere in `src/`, `tests/`, `tests_integ/`, README; not exported from `evaluators/__init__.py` or top level; README rewritten around `disclosure=`. |
| 2. Wiring in base class, judge window | ✅ `_render_with_disclosure` / `_resolve_disclosure_index` / `_judge_window_tokens` live once in `evaluator.py`; inline prompt is **byte-identical** to `main` for `TrajectoryEvaluator` and `HelpfulnessEvaluator` with `tools == []` (checked programmatically). |
| 3. Tests prove the switch | ✅ 13 new tests in `test_evaluator_disclosure.py` incl. byte-identical, three-tools-on-overflow, `never`, `always`, no-Session. |
| 4. Four old inline threads | ✅ all verified fixed: `"approved. Amount"` now hits across a newline; `search_spans("")` → `ERROR: empty pattern…`; empty session → `ERROR: trace has no spans`; `for_judge()` returns a copy. (Threads are `github-actions`'s to resolve.) |
| 5. #399 first | ⏳ still OPEN — with `disclosure="never"` or the 🔴 below, an overflow is `score: 0` until it lands. |

### 🔴 Tool-level judges preflight the wrong row — later rows still overflow

`_tool_level_disclosure` probes with **`tool_inputs[0]`** (`evaluator.py:363`) on the premise that *"Every tool call in a case is judged against the same session history"* (`:354`). It isn't: `ToolLevelInput.session_history` is cumulative — `extractors/trace_extractor.py:149` builds it as `list(session_history) + [prior_executions]` and appends each turn's prompt/tools/response before the next trace. Row 0 is the smallest prompt by construction. Repro (6 turns, 120K-char results, default `ToolSelectionAccuracyEvaluator()`):

```
rows: 6 | judge window: 200000
  row 0: prompt       139 chars  would_exceed=False
  row 1: prompt   240,202 chars  would_exceed=False
  row 2: prompt   480,265 chars  would_exceed=True
  …
  row 5: prompt 1,200,454 chars  would_exceed=True
_tool_level_disclosure decision: disclose = False
```

**Who hits it:** `ToolSelectionAccuracyEvaluator()` / `ToolParameterAccuracyEvaluator()` with default `disclosure="auto"` on any multi-turn `Session` via `Experiment` — rows 2–5 go to the judge inlined, overflow, and land as `score: 0` (pre-#399) or not-applicable (post-#399). Suggestion — probe the largest row (they're all rendered in the loop anyway) and fix the docstring premise at `:354`:

```suggestion
        probe = (
            max((self._format_tool_level_prompt(ti) for ti in tool_inputs), key=len)
            if self.disclosure == "auto"
            else ""
        )
```

Repro script + output: `repro-tool-level-probe.py` / `.log` (review artifacts).

<details>
<summary><b>Questions (non-blocking)</b></summary>

1. `OutputEvaluator` is deliberately not wired (agreed: additive/opt-in there). Worth one README line, since a user who reads the `disclosure` docs will try `OutputEvaluator(disclosure="always")` and get `TypeError`.
2. `_JUDGE_CONTEXT_WINDOWS` is a substring table in the evaluator base (`nova-micro`/`nova-lite`/`nova-pro`/`claude`, else `DEFAULT_MAX_INPUT_TOKENS`). Is this where the maintainers want model metadata to live, and should it be overridable without monkey-patching (e.g. a `judge_window_tokens=` kwarg or a hook)? Custom `Model` objects and OpenAI/Gemini judges all fall to the 200K default.
3. Should `disclosure` be threaded through `Evaluator.__init__` once rather than copy-pasted into ~18 signatures? Same result, one place to validate.

</details>

<details>
<summary><b>Evidence ledger</b></summary>

- Head `a8e0a87` vs merge-base `abf7b91`: +2,097/−126 across 29 files.
- Full unit run: **2,139 passed, 8 failed, 49 errors** — every failure/error is in `providers/test_langfuse_provider.py` (`module 'strands_evals.providers' has no attribute 'langfuse_provider'`, ×47) or `telemetry/test_config.py` (`module 'opentelemetry' has no attribute 'exporter'`, ×10). Both are missing optional deps in my sandbox; neither module is touched by this PR. Log uploaded as `pytest.log`.
- Byte-identical check: `TrajectoryEvaluator._render_with_disclosure(...)` == `compose_test_prompt(...)` from `main`; `HelpfulnessEvaluator` likewise; `always` yields `['list_spans','get_span','search_spans']` and the `<TraceOverview>` block in the `# Conversation History:` slot.
- Review shape: this is a follow-up to my own checklist, so the checklist verification and the 🔴 repro are first-party. The three independent passes I dispatched (correctness, API bar-raiser, adversarial) **timed out** when my sandbox was recycled mid-run and produced no output; a narrower retry of the adversarial/correctness pass is in flight. I'll add a short addendum only if it surfaces something material — if you see nothing further from me, it found nothing beyond the above.

</details>

*I'm an AI reviewer — prepared input for a human decision. With the 🔴 fixed and #399 merged first, nothing else here blocks from my side; the API questions are `poshinchen`'s call and the reason for the `needs-api-review` suggestion.*

strandly-the-agent:
**Addendum — the retry pass finished; two 🟡 worth folding into the same fix commit, and a useful "verified clean" list.**

1. 🟡 **Disclosure is a silent no-op for a non-`Session` trajectory — even with `disclosure="always"`.** `_session_of` (`evaluator.py:127-131`) returns `None` unless `actual_trajectory` is a `Session`, but the public type is `list[Any] | Session | None` (`types/evaluation.py:104`) and the README's own `TrajectoryEvaluator` example feeds a *list* from `extract_agent_tools_used_from_messages` (`README:188-196`); the skill judges document "a `Session` or a raw message list" (`README:577-578`). Measured: with a list, `always` → `tools=[]`, 2.6M-char prompt inlined, no overview. Under `auto` that equals pre-PR behaviour (not a regression), but the new README section says the judges "handle this automatically", and an explicit `always` doing nothing with no signal is the bad part. Suggestion: one `logger.warning` when disclosure is wanted, a trajectory is present, and it isn't a `Session`; one README sentence naming the `Session` requirement.
2. 🟡 **17 of the 18 new `tools=tools` wirings are untested.** Deleting `tools=tools` from both `Agent(...)` calls in `helpfulness_evaluator.py:69,87` leaves the suite green (84 passed on the relevant files); only `TrajectoryEvaluator`'s wiring is asserted (`test_evaluator_disclosure.py:204`). That's the worst failure shape this feature can have — the prompt tells the judge to read via `list_spans`/`get_span`/`search_spans` and the agent has no such tools — so it should be pinned: one parametrized test over the judge classes asserting the disclosure path passes exactly those three tools to `Agent`. (The other mutation — `_resolve_disclosure_index` always `None` — *is* caught, 4 failures.)

<details>
<summary>⚪ (2) and verified clean</summary>

- ⚪ Skill judges' probe under-counts: `skill_selection_accuracy_evaluator.py:222` probes only the tail (`inline_context[1]`), omitting the `## Task`/`## Available skills` head and per-skill block; `skill_instruction_following_evaluator.py:202` probes `invoked[0]` so a later skill with a bigger `SKILL.md` is under-counted. Bounded by the safety margin. Also `_case_context_and_tools` (`:220-224`) builds the inline context unconditionally, so `always` serialises up to 600KB it then discards — `_render_with_disclosure` deliberately avoids that; mirror it.
- ⚪ The disclosure path can still overflow on a huge single `agent_response` (kept inline by design) — verified it raises `ContextWindowOverflowException` loudly rather than mis-scoring. No action.

**Verified clean:** probe == prompt actually sent for every single-render judge (`goal_success_rate` both branches, `coherence`, `correctness`, `helpfulness`, +13); `correctness._evaluate_with_reference` correctly has no disclosure (inlines no trajectory); `tools` forwarded at all 18 call sites (static sweep); `_parse_trajectory` already hard-requires a bare `Session` for trace/tool/session-level judges, so finding 1 is confined to `TrajectoryEvaluator` + the two skill judges; `disclosure="never"` propagates `strands.types.exceptions.ContextWindowOverflowException` unchanged and `_is_context_exceeded` (`detectors/utils.py:33-53`) matches it — #399's catch will work; preflight threshold (~130K tokens) sits below `serialize_trajectory`'s 600K-char cap, so the cap doesn't mask disclosure.

Evidence uploaded as review artifacts: `probe.py`, `probe-output.txt`, `mutation-check.txt`.

</details>

Nothing here changes the verdict above: fix the 🔴 (largest-row probe), add the warning + test from this addendum, merge #399 first, and get the `needs-api-review` look at the `disclosure=` kwarg.

PR #352, opened 2026-08-05
fix(test): update ADK multi-agent integ test for single-trace design

liramon2:
## Description

Updates `test_adk_multi_agent_evaluation` to match the single-trace design introduced in #336.

PR #336 replaced per-agent trace splitting with `agent_span_id`-based tool scoping — all agents in a multi-agent ADK session now live in a single `Trace`. The integration test still asserted `len(session.traces) == 2`, causing it to fail.

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/tests_integ/test_adk_eval.py
+++ b/tests_integ/test_adk_eval.py
@@ -334,42 +334,36 @@ def test_adk_multi_agent_evaluation(telemetry, create_multi_agent_runner):
     # Tool spans are scoped to their owning agent via agent_span_id
     tool_spans = [s for s in trace.spans if isinstance(s, ToolExecutionSpan)]
 
-    # Find each agent's span_id for ownership checks
-    coordinator_span_id = next(
-        s.span_info.span_id for s in agent_spans if s.metadata.get("agent_name") == "coordinator"
-    )
-    specialist_span_id = next(
+    coordinator_span_ids = {s.span_info.span_id for s in agent_spans if s.metadata.get("agent_name") == "coordinator"}
+    specialist_span_ids = {
         s.span_info.span_id for s in agent_spans if s.metadata.get("agent_name") == "weather_specialist"
-    )
+    }
 
     # get_weather should be owned by the specialist, not the coordinator
     for tool_span in tool_spans:
         if tool_span.tool_call.name == "get_weather":
-            assert tool_span.agent_span_id == specialist_span_id, (
-                f"get_weather should be scoped to specialist ({specialist_span_id}), "
+            assert tool_span.agent_span_id in specialist_span_ids, (
+                f"get_weather should be scoped to the specialist ({specialist_span_ids}), "
                 f"got agent_span_id={tool_span.agent_span_id}"
             )
             assert tool_span.tool_call.tool_call_id is not None, (
                 f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
             )
         elif tool_span.tool_call.name == "transfer_to_agent":
-            assert tool_span.agent_span_id == coordinator_span_id, (
-                f"transfer_to_agent should be scoped to coordinator ({coordinator_span_id}), "
-                f"got agent_span_id={tool_span.agent_span_id}"
+            assert tool_span.agent_span_id in coordinator_span_ids | specialist_span_ids, (
+                f"transfer_to_agent should be scoped to a known agent span, got agent_span_id={tool_span.agent_span_id}"
             )
 
     # Coordinator tools should only be transfer_to_agent — no specialist tools leaked
-    coordinator_tools = [s for s in tool_spans if s.agent_span_id == coordinator_span_id]
+    coordinator_tools = [s for s in tool_spans if s.agent_span_id in coordinator_span_ids]
     coordinator_tool_names = {s.tool_call.name for s in coordinator_tools}
     assert "get_weather" not in coordinator_tool_names, (
         f"Specialist tool leaked onto coordinator: {coordinator_tool_names}"
     )
-    assert coordinator_tool_names <= {"transfer_to_agent"}, (
-        f"Unexpected tools on coordinator: {coordinator_tool_names}"
-    )
+    assert coordinator_tool_names <= {"transfer_to_agent"}, f"Unexpected tools on coordinator: {coordinator_tool_names}"
 
     # Specialist must have at least the get_weather call
-    specialist_tools = [s for s in tool_spans if s.agent_span_id == specialist_span_id]
+    specialist_tools = [s for s in tool_spans if s.agent_span_id in specialist_span_ids]
     assert len(specialist_tools) >= 1, "Specialist should have at least one tool span"
     specialist_tool_names = {s.tool_call.name for s in specialist_tools}
     assert "get_weather" in specialist_tool_names, f"Expected get_weather tool, got {specialist_tool_names}"
--- a/tests_integ/test_adk_eval.py
+++ b/tests_integ/test_adk_eval.py
@@ -320,33 +320,56 @@ def test_adk_multi_agent_evaluation(telemetry, create_multi_agent_runner):
     # Deserialize the trajectory from the first report case back into a Session object.
     session = Session.model_validate(report.cases[0]["actual_trajectory"])
 
-    assert len(session.traces) == 2, (
-        f"Multi-agent should produce 2 traces (coordinator + specialist), got {len(session.traces)}"
+    assert len(session.traces) == 1, (
+        f"Multi-agent should produce 1 trace (all agents share the same OTEL trace_id), got {len(session.traces)}"
     )
 
-    # Identify traces by agent name in their AgentInvocationSpan metadata
-    agent_spans = [s for t in session.traces for s in t.spans if isinstance(s, AgentInvocationSpan)]
+    trace = session.traces[0]
+
+    agent_spans = [s for s in trace.spans if isinstance(s, AgentInvocationSpan)]
     agent_names = {s.metadata.get("agent_name") for s in agent_spans}
     assert "coordinator" in agent_names, f"Expected coordinator agent, got {agent_names}"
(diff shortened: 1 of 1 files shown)


liramon2:
@strandly-the-agent Review this PR real quick

tests_integ/test_adk_eval.py
```diff
@@ -320,33 +320,56 @@ def task_function(case: Case) -> dict:
...
+                f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
+            )
+        elif tool_span.tool_call.name == "transfer_to_agent":
+            assert tool_span.agent_span_id == coordinator_span_id, (
+                f"transfer_to_agent should be scoped to coordinator ({coordinator_span_id}), "
+                f"got agent_span_id={tool_span.agent_span_id}"
+            )
+
+    # Coordinator tools should only be transfer_to_agent — no specialist tools leaked
+    coordinator_tools = [s for s in tool_spans if s.agent_span_id == coordinator_span_id]
```
strandly-the-agent: 🟡 **A valid ADK trace fails here.** When `weather_specialist` transfers back to `coordinator` (ADK's default — `disallow_transfer_to_parent=False`, and its injected instruction tells the sub-agent to do it), the specialist emits its own `transfer_to_agent` span, correctly attributed to the *specialist*. Line 356 hard-asserts it must equal `coordinator_span_id`, so the test fails on correct mapper output:

```
agent coordinator          span_id=coord-1
tool  transfer_to_agent    agent_span_id=coord-1
agent weather_specialist   span_id=spec
tool  get_weather          agent_span_id=spec
tool  transfer_to_agent    agent_span_id=spec   <-- legitimately the specialist's
agent coordinator          span_id=coord-2      <-- coordinator re-entered

AssertionError: transfer_to_agent should be scoped to coordinator (coord-1), got agent_span_id=spec
```

Same root cause makes the leak check at `:362` narrower than intended: filtering on a single `coordinator_span_id` means a tool attributed to `coord-2` escapes it (a mapper regression that leaks a non-`get_weather` tool onto the second coordinator span passes today).

Set-based lookup fixes both. Verified across 9 scenarios (2 baseline shapes, no-delegation, specialist-never-calls-tool, transfer-back, 4 injected mapper regressions): current block wrong on 2/9, this on 0/9 — every regression case still fails, and it's `ruff check` + `ruff format` clean.

```suggestion
    # An agent can be invoked more than once per trace (ADK re-enters the coordinator when a
    # sub-agent transfers back), so collect every span_id per agent name.
    coordinator_span_ids = {s.span_info.span_id for s in agent_spans if s.metadata.get("agent_name") == "coordinator"}
    specialist_span_ids = {
        s.span_info.span_id for s in agent_spans if s.metadata.get("agent_name") == "weather_specialist"
    }

    # get_weather should be owned by the specialist, not the coordinator
    for tool_span in tool_spans:
        if tool_span.tool_call.name == "get_weather":
            assert tool_span.agent_span_id in specialist_span_ids, (
                f"get_weather should be scoped to the specialist ({specialist_span_ids}), "
                f"got agent_span_id={tool_span.agent_span_id}"
            )
            assert tool_span.tool_call.tool_call_id is not None, (
                f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
            )
        elif tool_span.tool_call.name == "transfer_to_agent":
            # ADK also grants transfer_to_agent to the specialist (transfer back to the parent),
            # so require attribution to one of the known agent spans, not to the coordinator.
            assert tool_span.agent_span_id in coordinator_span_ids | specialist_span_ids, (
                f"transfer_to_agent should be scoped to a known agent span, got agent_span_id={tool_span.agent_span_id}"
            )

    # Coordinator tools should only be transfer_to_agent — no specialist tools leaked
    coordinator_tools = [s for s in tool_spans if s.agent_span_id in coordinator_span_ids]
```

The #336 invariant survives and gets slightly stronger: `get_weather` must belong to a specialist span and to *no* coordinator span.

tests_integ/test_adk_eval.py
```diff
@@ -320,33 +320,56 @@ def task_function(case: Case) -> dict:
...
+    coordinator_tool_names = {s.tool_call.name for s in coordinator_tools}
+    assert "get_weather" not in coordinator_tool_names, (
+        f"Specialist tool leaked onto coordinator: {coordinator_tool_names}"
+    )
+    assert coordinator_tool_names <= {"transfer_to_agent"}, (
+        f"Unexpected tools on coordinator: {coordinator_tool_names}"
+    )
+
+    # Specialist must have at least the get_weather call
+    specialist_tools = [s for s in tool_spans if s.agent_span_id == specialist_span_id]
```
strandly-the-agent: Companion to the suggestion above — same reason (the specialist could also be invoked more than once).

```suggestion
    specialist_tools = [s for s in tool_spans if s.agent_span_id in specialist_span_ids]
```

strandly-the-agent (commented):
**One should-fix:** the rewrite is sound and CI-green, but the new `transfer_to_agent` assertion will fail on a *valid* ADK trace — whenever the specialist transfers back to the coordinator, that span is legitimately owned by the specialist. Intermittent, model-dependent flake; fix suggested inline (verified 9/9 scenarios).

**Verified at `3a767c1`**
- ✅ `ruff check tests_integ/test_adk_eval.py` → clean; all CI checks green on this head, incl. `Run integration tests`
- ✅ Mutation-tested the new assertions against 3 plausible regressions of `Trace.model_post_init` (`src/strands_evals/types/trace.py:162-206`) — coordinator swallows specialist subtree / everything→root / no-op — **all three fail loudly**, baseline passes. The invariant #336 added is genuinely guarded, and more directly than the old `len(traces) == 2` bucketing.
- 🟡 Reproduced the transfer-back failure offline against the real `ADKOtelSessionMapper`: `AssertionError: transfer_to_agent should be scoped to coordinator (coord-1), got agent_span_id=spec`
- ⚪ `ruff format --check` wants two of the new lines rewrapped (not CI-blocking — `test-lint` only runs the linter, `pyproject.toml:102-104` → `lint-check`); `hatch run prepare`'s formatter step would rewrite them.

<details><summary>Why the transfer-back path is real, not contrived</summary>

ADK gives the sub-agent `transfer_to_agent` with the parent as a target by default (`disallow_transfer_to_parent` defaults to `False`) and its injected instruction explicitly says to transfer back to the parent when the sub-agent isn't the best fit. It's an ordinary `FunctionTool`, so it emits its own `execute_tool` span — owned by `weather_specialist`, which is correct. The coordinator is then re-entered, producing a **second** `invoke_agent coordinator` span, so `next(...)` at `:338-343` resolves only the first one.

The fixture's own prompt invites exactly this: `"Summarize the specialist's response in one sentence"` (`tests_integ/test_adk_eval.py:140-142`) is only reachable after control returns to the coordinator. Whether Gemini takes the path is nondeterministic — it didn't on this CI run — which is why this reads as a latent flake rather than a hard break, and the failure message would wrongly accuse the mapper.

Comparison across 9 scenarios (baseline sibling + legacy nested shapes, no-delegation, specialist-never-calls-tool, transfer-back, and 4 injected mapper regressions): the current block is wrong on 2/9, the suggested version on 0/9 — it still fails every regression case, so the fix doesn't cost teeth.
</details>

<details><summary>How this was reviewed</summary>

Context-build + three independent fresh-context passes (correctness, test-quality/mutation, adversarial), then de-duped and severity-gated. Notably the correctness and adversarial passes *disagreed*: the correctness pass reproduced the two-coordinator-span topology but concluded it was harmless, having not traced the specialist's own `transfer_to_agent` span through the `elif` branch. I re-ran the adversarial repro myself and it holds. Couldn't run the integ test itself (no live `GOOGLE_API_KEY`), so all repros go through the real mapper + `Session.model_validate` with the assertion block extracted verbatim from the file.
</details>

<details><summary>Appendix — non-blocking (3)</summary>

- ⚪ **`tool_call_id` isn't asserted for `transfer_to_agent`** (`:355-358`) — the `get_weather` branch checks it, the `elif` doesn't; setting it to `None` still passes. Parity with `main` (that span used to live on the coordinator trace, whose branch didn't check either) and the mapper docstring claims `ToolExecutionSpan` always carries `gen_ai.tool.call.id` (`src/strands_evals/mappers/adk_otel_session_mapper.py:21-22`) — but I have no captured multi-agent trace to confirm that for the delegation span, so I deliberately left the assertion *out* of the suggestion rather than risk a new failure. Add it if you know it holds.
- ⚪ **Formatter drift**: `ruff format` collapses `assert coordinator_tool_names <= {"transfer_to_agent"}, (...)` (`:367-369`) to one line. The inline suggestion is already `ruff format`-clean; this second spot isn't covered by it.
- ⚪ **`next()` without `default=`** (`:338-343`) would raise `StopIteration` rather than assert — provably unreachable given the `agent_names` guards at `:331-332`, and moot with the suggested set comprehensions. Mentioned only so it doesn't get re-raised later.

Nothing here is pre-existing debt worth a separate issue.
</details>

**Question (non-blocking):** do you *want* this test to exercise the transfer-back topology? If yes, the inline fix is the right shape. If you'd rather the trace shape be deterministic, `disallow_transfer_to_parent=True` on the specialist pins it — at the cost of never covering the transfer-back attribution the mapper now has to handle. Your call; it decides which fix to take.

---
*Reviewed by `strandly-the-agent` — an experimental AI agent. Solid work for a human to approve, not a gate; push back if I've got something wrong.*

tests_integ/test_adk_eval.py
liramon2: Addressed in 62ae1dce3fdf78c0f34916e594462e47491581a7

tests_integ/test_adk_eval.py
liramon2: Addressed in 62ae1dce3fdf78c0f34916e594462e47491581a7

opieter-aws (approved):

poshinchen (approved):

Issue #329, opened 2026-07-27
[FEATURE] Add Claude Agents SDK support to OpenInference session mapper


liramon2:
### Problem Statement

Teams building AI agents with Claude Agents SDK cannot use strands-evals to evaluate their agent traces (such as from OpenInference). Without a supported mapper, Claude Agents users must either manually transform their traces into Session format or forgo trace-based evaluation entirely.

### Use Case

- As a Claude Agents user, I want to evaluate my agent with strands-evals without rewriting my trace collection code

PR #340, opened 2026-07-31
feat: add support for Claude agents to OpenInference mapper

liramon2:
## Description
The evals framework currently cannot evaluate agents built with Claude Agent SDK, which may be instrumented with OpenInference and produce spans with the scope `openinference.instrumentation.claude_agent_sdk`. This PR modifies `OpenInferenceSessionMapper` to map Claude Agent spans into Sessions. It includes support for unified ADOT spans and in-memory spans.

This also includes producer agnostic fixes for the following OpenInference bugs:
- If `output.value.content` on `execute_tool` spans contained a list of text blocks, it was converted into a repr string instead having its text joined. Similarly, if output.value itself was a JSON list of content blocks, it was stored as the raw JSON string without text extraction.
- Agent spans with an error code but no output value were rejected from conversion instead of surfaced to the judge.

## Related Issues

https://github.com/strands-agents/evals/issues/329

## Type of Change

New feature

## Testing

How have you tested the change? Verify that the changes do not break functionality or introduce new warnings.

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works
- [x] My change is focused and reasonably small; I have split unrelated work into separate PRs
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #353, opened 2026-08-05
feat: add Claude Agents OpenInference integration tests

liramon2:
## Description
Add integration tests for the Claude Agents SDK mapper.

This also fixes a minor bug caught by the integration tests, where Claude Agents could emit tool spans with status `async_launched` and those spans were wrongfully treated as errors.

## Related Issues

https://github.com/strands-agents/evals/issues/329

## Type of Change

New feature

## Testing

- [x] I ran the integration tests locally and it succeeded
- [x] I ran the integration test suite with the CI pipeline and it succeeded

## Checklist

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,10 @@ langchain = [
     "opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0",
 ]
 adk = ["google-adk>=2.0.0,<3.0.0"]
+claude = [
+    "claude-agent-sdk>=0.2.0,<0.3.0",
+    "openinference-instrumentation-claude-agent-sdk>=0.1.8,<0.2.0",
+]
 
 [tool.ruff]
 line-length = 120
@@ -74,7 +78,7 @@ include = ["src/**/*.py", "tests/**/*.py", "tests_integ/**/*.py"]
 [tool.hatch.envs.hatch-test]
 installer = "uv"
 extra-args = ["-n", "auto", "-vv"]
-features = ["otel", "langfuse", "langchain", "opensearch", "adk"]
+features = ["otel", "langfuse", "langchain", "opensearch", "adk", "claude"]
 dependencies = [
     "pytest>=8.0.0,<10.0.0",
     "pytest-cov>=7.0.0,<8.0.0",
--- a/tests_integ/test_claude_openinference_eval.py
+++ b/tests_integ/test_claude_openinference_eval.py
@@ -29,12 +29,11 @@ from strands_evals.mappers import OpenInferenceSessionMapper, detect_otel_mapper
 from strands_evals.telemetry import StrandsEvalsTelemetry
 from strands_evals.types.trace import AgentInvocationSpan, Session, ToolExecutionSpan
 
-# Bedrock model for the Claude Agent SDK
-DEFAULT_MODEL = os.environ.get("STRANDS_CLAUDE_TEST_MODEL", "us.anthropic.claude-sonnet-4-20250514-v1:0")
+DEFAULT_MODEL_ID = "us.anthropic.claude-sonnet-4-20250514-v1:0"
 
 BEDROCK_ENV = {
     "CLAUDE_CODE_USE_BEDROCK": "1",
-    "ANTHROPIC_MODEL": DEFAULT_MODEL,
+    "ANTHROPIC_MODEL": DEFAULT_MODEL_ID,
     "AWS_REGION": os.environ.get("AWS_REGION", "us-west-2"),
 }
 
@@ -83,7 +82,6 @@ async def _run_claude_agent(prompt: str, *, agents: dict | None = None) -> str:
     """Run a Claude Agent SDK query and return the final response text."""
     query = _get_query()
 
-    # allowed_tools is the permission allow-list for the entire session including sub-agents.
     allowed_tools = ["Agent", "Bash"] if agents else ["Bash"]
 
     options = ClaudeAgentOptions(
@@ -239,6 +237,9 @@ def test_claude_multi_agent_evaluation(telemetry):
         assert tool_span.tool_call.tool_call_id is not None, (
             f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
         )
+        # The Agent tool reports "async_launched" as an operational status, not a failure.
+        if tool_span.tool_call.name == "Agent" and tool_span.tool_result.error == "async_launched":
+            continue
         assert tool_span.tool_result.error is None, (
             f"tool '{tool_span.tool_call.name}' did not execute successfully: "
             f"error={tool_span.tool_result.error!r} content={tool_span.tool_result.content!r}"
--- a/tests_integ/test_claude_openinference_eval.py
+++ b/tests_integ/test_claude_openinference_eval.py
@@ -237,13 +237,16 @@ def test_claude_multi_agent_evaluation(telemetry):
         assert tool_span.tool_call.tool_call_id is not None, (
             f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
         )
-        # The Agent tool reports "async_launched" as an operational status, not a failure.
-        if tool_span.tool_call.name == "Agent" and tool_span.tool_result.error == "async_launched":
-            continue
         assert tool_span.tool_result.error is None, (
             f"tool '{tool_span.tool_call.name}' did not execute successfully: "
             f"error={tool_span.tool_result.error!r} content={tool_span.tool_result.content!r}"
         )
+
+    executed = [s for s in tool_spans if s.tool_call.name != "Agent" and s.tool_result.error is None]
+    assert executed, (
+        "Expected at least one successful sub-agent tool execution, got "
+        f"{[(s.tool_call.name, s.tool_result.error) for s in tool_spans]}"
+    )
 
     agent_spans = [s for t in session.traces for s in t.spans if isinstance(s, AgentInvocationSpan)]
     assert len(agent_spans) >= 1, "Expected at least one AgentInvocationSpan"
--- a/tests_integ/test_claude_openinference_eval.py
+++ b/tests_integ/test_claude_openinference_eval.py
@@ -1,0 +1,252 @@
+"""Integration tests for Claude Agent SDK evaluation with OpenInference instrumentation.
+
+Creates Claude Agent SDK agents, runs them against Bedrock, captures OpenInference
+traces (scope: openinference.instrumentation.claude_agent_sdk) in memory, and
+evaluates using strands-evals via the OpenInferenceSessionMapper.
+
+Requirements:
+    pip install strands-agents-evals[claude]
+    AWS credentials configured for Amazon Bedrock access.
+
+Run with: pytest tests_integ/test_claude_openinference_eval.py -v
+"""
+
+import asyncio
+import os
+import threading
+
+import pytest
+from claude_agent_sdk import AgentDefinition, ClaudeAgentOptions
+from openinference.instrumentation.claude_agent_sdk import ClaudeAgentSDKInstrumentor
+
+from strands_evals import Case, Experiment
+from strands_evals.evaluators import (
+    CorrectnessEvaluator,
+    GoalSuccessRateEvaluator,
+    ToolSelectionAccuracyEvaluator,
+)
+from strands_evals.mappers import OpenInferenceSessionMapper, detect_otel_mapper, readable_spans_to_dicts
+from strands_evals.telemetry import StrandsEvalsTelemetry
+from strands_evals.types.trace import AgentInvocationSpan, Session, ToolExecutionSpan
+
+# Bedrock model for the Claude Agent SDK
+DEFAULT_MODEL = os.environ.get("STRANDS_CLAUDE_TEST_MODEL", "us.anthropic.claude-sonnet-4-20250514-v1:0")
+
+BEDROCK_ENV = {
+    "CLAUDE_CODE_USE_BEDROCK": "1",
+    "ANTHROPIC_MODEL": DEFAULT_MODEL,
+    "AWS_REGION": os.environ.get("AWS_REGION", "us-west-2"),
+}
+
+
+# =============================================================================
+# Fixtures
+# =============================================================================
+
+
+@pytest.fixture(scope="module")
+def telemetry():
+    """Setup OpenTelemetry with in-memory exporter and Claude Agent SDK instrumentation."""
+    telemetry = StrandsEvalsTelemetry().setup_in_memory_exporter()
+
+    instrumentor = ClaudeAgentSDKInstrumentor()
+    instrumentor.instrument()
+
+    yield telemetry
+
+    instrumentor.uninstrument()
+
+
+# =============================================================================
+# Helpers
+# =============================================================================
+
+# Import query lazily after instrumentation is set up (module-level import
+# would bypass instrumentation). The fixture ensures instrumentation runs first.
+_query_func = None
+_query_lock = threading.Lock()
+
+
+def _get_query():
+    """Lazily import the query function after instrumentation is applied."""
+    global _query_func
+    if _query_func is None:
+        with _query_lock:
+            if _query_func is None:
+                from claude_agent_sdk import query
+
+                _query_func = query
+    return _query_func
+
+
+async def _run_claude_agent(prompt: str, *, agents: dict | None = None) -> str:
+    """Run a Claude Agent SDK query and return the final response text."""
+    query = _get_query()
+
+    # allowed_tools is the permission allow-list for the entire session including sub-agents.
+    allowed_tools = ["Agent", "Bash"] if agents else ["Bash"]
+
+    options = ClaudeAgentOptions(
+        allowed_tools=allowed_tools,
+        max_turns=10,
+        agents=agents or {},
+        env=BEDROCK_ENV,
+    )
+
+    async def _run_query() -> str:
+        out = ""
+        async for message in query(prompt=prompt, options=options):
+            if hasattr(message, "result"):
+                if message.result:
+                    out = message.result
+                elif getattr(message, "is_error", False):
+                    raise RuntimeError(
+                        f"Claude Agent SDK run failed: subtype={getattr(message, 'subtype', None)!r} "
+                        f"terminal_reason={getattr(message, 'terminal_reason', None)!r}"
+                    )
+        return out
+
+    final_output = await asyncio.wait_for(_run_query(), timeout=180)
+
+    return final_output
+
+
+# =============================================================================
+# Tests — Single Agent
+# =============================================================================
+
+
+def test_claude_single_query(telemetry):
+    """Spans are captured, mapper is auto-detected, session is valid, and tool_call_ids are populated."""
+    telemetry.in_memory_exporter.clear()
+    response = asyncio.run(_run_claude_agent("Use bash to calculate: echo $((15 * 37)). Just give me the number."))
+    spans = readable_spans_to_dicts(telemetry.in_memory_exporter.get_finished_spans())
+
+    assert len(spans) > 0, "Should have captured OTEL spans"
+
+    mapper = detect_otel_mapper(spans)
+    assert isinstance(mapper, OpenInferenceSessionMapper), (
+        f"Expected OpenInferenceSessionMapper but got {type(mapper).__name__}"
+    )
+
+    session = mapper.map_to_session(spans, session_id="test-single")
+
+    assert session.session_id == "test-single"
+    assert len(session.traces) > 0, "Should have at least one trace"
+    assert "555" in response, f"Expected 555 in response, got: {response}"
+
+    # Verify tool spans have tool_call_id populated and succeeded
+    tool_spans = [s for t in session.traces for s in t.spans if isinstance(s, ToolExecutionSpan)]
+    assert len(tool_spans) >= 1, "Expected at least one tool execution span"
+    for tool_span in tool_spans:
+        assert tool_span.tool_call.tool_call_id is not None, (
+            f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
+        )
+        assert tool_span.tool_result.error is None, (
+            f"tool '{tool_span.tool_call.name}' did not execute successfully: "
+            f"error={tool_span.tool_result.error!r} content={tool_span.tool_result.content!r}"
+        )
+
+
+def test_claude_single_agent_evaluation(telemetry):
+    """Single-agent session evaluates correctly via the full experiment pipeline."""
+    test_cases = [
+        Case[str, str](
+            name="bash-calculation",
+            input="Use bash to calculate: echo $((100 + 200)). Just give me the number.",
+            expected_output="300",
+            expected_assertion="The agent used the Bash tool to compute 100+200 and responded with 300.",
+        ),
+    ]
+
+    def task_function(case: Case) -> dict:
+        telemetry.in_memory_exporter.clear()
+        response = asyncio.run(_run_claude_agent(case.input))
+
+        spans = readable_spans_to_dicts(telemetry.in_memory_exporter.get_finished_spans())
+        mapper = detect_otel_mapper(spans)
+        session = mapper.map_to_session(spans, session_id=case.session_id)
+        return {"output": response, "trajectory": session}
+
+    experiment = Experiment(cases=test_cases, evaluators=[GoalSuccessRateEvaluator()])
+    report = experiment.run_evaluations(task_function)
+
+    assert len(report.scores) == 1
+    assert all(report.test_passes), f"Some evaluations failed: {report.reasons}"
+
+
+# =============================================================================
+# Tests — Multi Agent
+# =============================================================================
+
+
+def test_claude_multi_agent_evaluation(telemetry):
+    """Multi-agent delegation + tool call spans evaluate correctly."""
+    agents = {
+        "math-specialist": AgentDefinition(
+            description=(
+                "Math specialist agent. Delegate any mathematical calculations "
+                "including arithmetic, algebra, square roots, and numeric computations."
+            ),
+            prompt=(
+                "You are a math specialist. Use Bash ONLY to run `python3 -c 'import math; print(...)'` "
+                "for calculations. Never use Bash for anything else. "
+                "Be precise and return just the numeric result."
+            ),
+            tools=["Bash"],
+        ),
+    }
+
+    test_cases = [
+        Case[str, str](
+            name="multi-agent-math",
+            input="Calculate the square root of 1764, then multiply that result by 3.",
+            expected_output="126",
+        ),
+    ]
+
+    def task_function(case: Case) -> dict:
+        telemetry.in_memory_exporter.clear()
+        response = asyncio.run(_run_claude_agent(case.input, agents=agents))
+
+        spans = readable_spans_to_dicts(telemetry.in_memory_exporter.get_finished_spans())
+        mapper = detect_otel_mapper(spans)
+        session = mapper.map_to_session(spans, session_id=case.session_id)
+        return {"output": response, "trajectory": session}
+
+    # One evaluator per extraction level: SESSION, TRACE, TOOL
+    evaluators = [
+        GoalSuccessRateEvaluator(),
+        CorrectnessEvaluator(),
+        ToolSelectionAccuracyEvaluator(),
+    ]
+
+    experiment = Experiment(cases=test_cases, evaluators=evaluators)
+    report = experiment.run_evaluations(task_function)
+
+    assert len(report.scores) == 3
+    assert all(report.test_passes[:2]), f"Some evaluations failed: {report.reasons[:2]}"
+    assert report.scores[2] >= 0.5, f"Tool selection accuracy too low: {report.reasons[2]}"
+
+    session = Session.model_validate(report.cases[0]["actual_trajectory"])
+
+    assert len(session.traces) >= 1, f"Multi-agent should produce at least 1 trace, got {len(session.traces)}"
+
+    # Verify tool spans succeeded
+    tool_spans = [s for t in session.traces for s in t.spans if isinstance(s, ToolExecutionSpan)]
+    assert len(tool_spans) >= 1, "Expected at least one tool execution span from sub-agent"
+    for tool_span in tool_spans:
+        assert tool_span.tool_call.tool_call_id is not None, (
+            f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
+        )
+        assert tool_span.tool_result.error is None, (
+            f"tool '{tool_span.tool_call.name}' did not execute successfully: "
+            f"error={tool_span.tool_result.error!r} content={tool_span.tool_result.content!r}"
+        )
+
+    agent_spans = [s for t in session.traces for s in t.spans if isinstance(s, AgentInvocationSpan)]
+    assert len(agent_spans) >= 1, "Expected at least one AgentInvocationSpan"
+
+    # Prove delegation actually happened
+    delegations = [s.tool_call.arguments.get("subagent_type") for s in tool_spans if s.tool_call.name == "Agent"]
+    assert "math-specialist" in delegations, f"Expected delegation to math-specialist, got {delegations}"
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -53,6 +53,12 @@ _LLM_METADATA_KEYS = (
     "llm.token_count.total",
 )
 
+# Tool statuses that indicate a successful execution.
+# - "success": smolagents / LangChain default
+# - "completed": Claude Agent SDK synchronous tool completion
+# - "async_launched": Claude Agent SDK async sub-agent delegation
+_TOOL_SUCCESS_STATUSES = frozenset(("success", "completed", "async_launched"))
+
 
 class OpenInferenceSessionMapper(SessionMapper):
     """Maps OpenInference traces to Session format.
@@ -558,8 +564,7 @@ class OpenInferenceSessionMapper(SessionMapper):
         tool_call = ToolCall(name=tool_name, arguments=tool_parameters or {}, tool_call_id=tool_call_id)
         tool_result = ToolResult(
             content=tool_output_content or "",
-            # "success" = smolagents/LangChain default; "completed" = Claude Agent SDK
-            error=None if tool_status in ("success", "completed") else tool_status,
+            error=None if tool_status in _TOOL_SUCCESS_STATUSES else tool_status,
             tool_call_id=tool_call_id,
         )
 
--- a/tests/strands_evals/mappers/test_openinference_session_mapper.py
+++ b/tests/strands_evals/mappers/test_openinference_session_mapper.py
@@ -1717,6 +1717,29 @@ class TestClaudeAgentSdkScopeSupport:
         # Non-ASCII must be preserved literally, not escaped
         assert "\\u" not in tool_spans[0].tool_result.content
 
+    def test_claude_async_launched_tool_status_is_not_an_error(self):
+        """async_launched is an operational status for async sub-agent delegation, not a failure."""
+        span = make_span(
+            name="Agent",
+            scope_name=CLAUDE_SDK_SCOPE_NAME,
+            attributes={
+                "openinference.span.kind": "TOOL",
+                "tool.id": "toolu_bdrk_async1",
+                "tool.name": "Agent",
+                "input.value": json.dumps(
+                    {"description": "Delegate", "subagent_type": "math-specialist", "prompt": "sqrt(1764)*3"}
+                ),
+                "output.value": json.dumps(
+                    {"status": "async_launched", "content": [{"type": "text", "text": "launched"}]}
+                ),
+            },
+        )
+        session = self.mapper.map_to_session([span], "sess-1")
+
+        tool_spans = [s for t in session.traces for s in t.spans if isinstance(s, ToolExecutionSpan)]
+        assert len(tool_spans) == 1
+        assert tool_spans[0].tool_result.error is None
+
     @pytest.mark.parametrize(
         "status,span_events,expected_error",
         [


liramon2:
@strandly-the-agent Review this PR. Only include changes in `tests_integ/test_claude_openinference_eval.py` and `pyproject.toml`. The commits beforehand are prerequisites for the integration test.

tests_integ/test_claude_openinference_eval.py
```diff
@@ -0,0 +1,267 @@
...
+        agents: Optional dict of sub-agent definitions for multi-agent mode.
+
+    Returns:
+        The agent's final text response.
+    """
+    query = _get_query()
+
+    allowed_tools = tools or ["Bash"]
+    if agents:
+        allowed_tools = ["Agent"]
```
strandly-the-agent: 🔴 **`allowed_tools=["Agent"]` drops `Bash`, so the sub-agents' tool calls are denied — this test fails ~83% of the time.**

`AgentDefinition(tools=["Bash"])` is a *restriction*, not a grant. `allowed_tools` is the actual permission allow-list and it governs sub-agent tool calls too, so with only `"Agent"` here every specialist Bash call is rejected.

Verified against live Bedrock, 15+ runs:

| config | result | denied Bash calls |
|---|---|---|
| as written (`["Agent"]`) | **10/12 FAIL** | 1–3 every run |
| `["Agent", "Bash"]` | **3/3 PASS** | 0 |

The trajectory from a failing run:

```python
'tool_call':   {'name': 'Bash', 'arguments': {'command': 'python3 -c "import math; print(math.sqrt(1764))"'}},
'tool_result': {'content': 'This command requires approval', 'error': 'error'}
```

Two things make this worse than a plain failure. First, the failure is *misleading*: the model retries with `dangerouslyDisableSandbox: True`, and the judge scores that as a safety-sandbox bypass — so CI reads like a Claude quality regression rather than a test-config bug. Second, even on the 2 runs that passed, the test never exercised a successful sub-agent tool execution — the specialist just did the arithmetic in its head (*"Since the Bash tool is unavailable in this environment, I'll work through this directly"*), so the multi-agent tool path isn't actually being covered today.

```suggestion
    allowed_tools = tools or ["Bash"]
    if agents:
        # allowed_tools is the *permission* allow-list and covers sub-agent tool
        # calls too — the specialists' AgentDefinition(tools=["Bash"]) does not
        # grant permission on its own. Omitting Bash here gets their calls denied
        # with "This command requires approval".
        allowed_tools = ["Agent", *(tools or ["Bash"])]
```

tests_integ/test_claude_openinference_eval.py
```diff
@@ -0,0 +1,267 @@
...
+        agents=agents or {},
+        env=BEDROCK_ENV,
+    )
+
+    final_output = ""
+    async for message in query(prompt=prompt, options=options):
+        if hasattr(message, "result") and message.result:
+            final_output = message.result
+
+    return final_output
```
strandly-the-agent: 🟡 **Every failure mode collapses into an empty string, and nothing bounds the call.**

`ResultMessage` carries `is_error`, `subtype` (e.g. `"error_max_turns"`) and `terminal_reason`, but this loop only checks `.result` for truthiness. On turn exhaustion or a Bedrock API error, `final_output` silently stays `""` and the run ends without an exception — so `assert "555" in response` reports `got: ` and a maintainer reads "model answered wrong" when the truth was "the run never completed".

I measured the no-timeout half with deliberately bad credentials:

```
$ pytest tests_integ/test_claude_openinference_eval.py::test_claude_single_query -x -q
E  Exception: Claude Code returned an error result: success
1 failed, 1 warning in 182.66s (0:03:02)
```

183s per test to surface a self-contradictory message, with no wall-clock cap here, none in `ClaudeAgentOptions`, and no `timeout-minutes` in `integration-test.yml` — GitHub's 360-minute default is the only backstop. Happy path for the whole module is ~53s, so a 180s cap is generous.

```suggestion
    async def _drain() -> str:
        out = ""
        async for message in query(prompt=prompt, options=options):
            if hasattr(message, "result"):
                if message.result:
                    out = message.result
                elif getattr(message, "is_error", False):
                    raise RuntimeError(
                        f"Claude Agent SDK run failed: subtype={getattr(message, 'subtype', None)!r} "
                        f"terminal_reason={getattr(message, 'terminal_reason', None)!r}"
                    )
        return out

    # Cap the live call so a wedged or retrying CLI subprocess (observed: ~183s on a
    # bad-credentials run) can't quietly eat the CI job's budget.
    return await asyncio.wait_for(_drain(), timeout=180)
```

tests_integ/test_claude_openinference_eval.py
```diff
@@ -0,0 +1,267 @@
...
+        GoalSuccessRateEvaluator(),
+        CorrectnessEvaluator(),
+        ToolSelectionAccuracyEvaluator(),
+    ]
+
+    experiment = Experiment(cases=test_cases, evaluators=evaluators)
+    report = experiment.run_evaluations(task_function)
+
+    assert len(report.scores) == 3
+    assert all(report.test_passes), f"Some evaluations failed: {report.reasons}"
```
strandly-the-agent: 🟡 **`all(report.test_passes)` requires N independent LLM judges to agree unanimously, where N is chosen by the model at runtime.**

`ToolSelectionAccuracyEvaluator` spawns one judge per tool call and its aggregator does `all(o.test_pass for o in outputs)`. So the gate is `p^N`, and N grows with however much work the agent decided to do — I saw 2, 3 and 4 judge calls across live runs, and 6 against your captured span fixture.

I'm not calling this a blocker: it's the same pattern already shipping in `test_adk_eval.py:306-311`, and once the `allowed_tools` fix lands it went 3/3 green with every judge saying Yes. But the residual flakiness is real and the fix is one line — the structural assertions below already pin the mapper's tool-level output, so this evaluator is more useful as a score than as a gate.

```suggestion
    # GoalSuccessRate / Correctness are the pass-fail gates. ToolSelectionAccuracy runs
    # one judge per tool call and all() would need N independent judges to agree, where
    # N is chosen by the model at runtime — score it, don't gate on it.
    assert all(report.test_passes[:2]), f"Some evaluations failed: {report.reasons[:2]}"
    assert report.scores[2] >= 0.5, f"Tool selection accuracy too low: {report.reasons[2]}"
```

strandly-the-agent (commented):
**Changes requested** (posted as a comment since this is a draft). Reviewed **only** `tests_integ/test_claude_openinference_eval.py` and `pyproject.toml`, per your scope request. You force-pushed mid-review — I checked, and both files are byte-identical between `8ec5a00` and the current head `838d153`, so everything below applies as-is.

**One 🔴: the multi-agent test fails ~83% of the time as written.** `allowed_tools=["Agent"]` drops `Bash`, so the specialists' Bash calls come back *"This command requires approval"*. I ran the real tests against live Bedrock 15+ times — **10 of 12 runs failed**, and **3/3 passed** with a one-line fix. Details inline at `_run_claude_agent`. Five 🟡 follow (error/timeout handling, judge-unanimity flakiness, assertion strength, dependency bounds, env-var collision); nits and PR hygiene are in the appendix.

The skeleton itself is good — it mirrors `test_adk_eval.py`'s idiom correctly and the `tool_call_id` assertions are genuine end-to-end coverage of the new mapper line.

<details><summary>✅ What was actually verified (and what wasn't)</summary>

- ✅ **Live Bedrock runs**: `test_claude_multi_agent_evaluation` executed 15+ times (`AWS_REGION=us-west-2`, `pytest -n auto -vv`). As written: **10/12 FAIL**, 1–3 denied Bash calls every run. With `allowed_tools=["Agent", "Bash"]`: **3/3 PASS**, `denied=0`, every judge returns Yes. `test_claude_single_query` and `test_claude_single_agent_evaluation` pass as written (5/5 returned `555`).
- ✅ **Mutation testing** against the real mapper + your own `claude_live_spans.json` fixture (10 spans / 3 delegations): reverting `tool.id` extraction and the CLAUDE-scope agent-span check *are* caught by your assertions. Reverting Claude's tool success/error mapping is **not** caught — `tool_result.error` is never read by any assertion here, nor by any evaluator or judge prompt (grep-confirmed across `evaluators/` and `extractors/`).
- ✅ **`us.anthropic.claude-sonnet-4-6` is valid and invocable** in both us-east-1 and us-west-2. I went in doubting this id and was wrong — it's the *better* choice, see the appendix.
- ✅ **`ClaudeAgentOptions.env` merges with `os.environ`** (`subprocess_cli.py:786-797`), so your 3-key `BEDROCK_ENV` doesn't strip the CI role's credentials. Confirmed live.
- ✅ **The lazy `_get_query()` import is load-bearing**, not cosmetic — a top-level `from claude_agent_sdk import query` permanently binds the *unwrapped* function, since the instrumentor patches via `wrapt` + `setattr` after collection. Reproduced in-process.
- ✅ **xdist isolation is fine**: 3 tests across `gw0`/`gw1` as separate processes, fixture per worker, no span bleed, no throttling in 15+ runs.
- 🔴 **Not verified**: whether the CI role has model access for this id, and whether 3 concurrent workers on a 4-CPU runner throttle. The job logs aren't retrievable with my token, so the link between the 428s CI failure and a credentials problem is a hypothesis, not a diagnosis.

</details>

<details><summary>❓ Questions (non-blocking)</summary>

- `conversion_specialist` is fully specced — formulas, three unit categories — but no test input ever mentions km/miles/°C/°F/kg/lbs, so nothing can route to it. Is a second case planned, or is math-only delegation the intended scope? As it stands it's ~15 lines of prompt text whose only effect is giving the coordinator a second delegation option to weigh on every run.
- `CorrectnessEvaluator`'s reference prompt only sees `USER QUERY` / `AGENT RESPONSE` / `EXPECTED RESPONSE` — no tool trace at all — so it can't observe the "delegated to a math specialist" half of your `expected_assertion`. `GoalSuccessRateEvaluator` *can* (via `subagent_type` in the tool-call args). Would splitting the compound assertion — the delegation claim for `GoalSuccessRate`, a plain answer-only claim for `Correctness` — make the grading mean what it says?
- The checklist says "I ran the integration tests locally and it succeeded", but `Run integration tests` fails on this head. Different creds/CLI availability locally, or was that run against an earlier revision? Worth reconciling before this leaves draft — with the caveat that the integ job was *already* failing before your test file existed (appendix).
- `#340`, the mapper this test exercises, is still open with no completed review. Is the plan for it to land on `main` first and this PR to rebase onto it (making this genuinely test-only), or is this PR the vehicle that lands the mapper too? Merging as-is ships both, which changes who needs to sign off.

</details>

<details><summary>Appendix — non-blocking (4 nits + 2 pre-existing bugs)</summary>

- ⚪ **Stale docstring** (`:10`): "Node.js + Claude Code CLI" is true at the `>=0.1.0` floor but false for the `0.2.130` that actually resolves — it bundles its own CLI binary. Suggest: *"claude-agent-sdk bundles its own CLI as of 0.2.x; Node.js/npm is only a fallback."*
- ⚪ **`expected_output="300"` (`:160`) is dead** — none of the three evaluators reads `Case.expected_output`; they all key off `expected_assertion`. Harmless, but it reads like a second independent check. Remove it or comment it as documentation.
- ⚪ **`tools=` on `_run_claude_agent` is unused** and silently discarded when `agents` is also passed. Drop the parameter or document the precedence in the docstring.
- ⚪ **The double-checked lock in `_get_query`** is correct but unnecessary — tests run sequentially within a worker and xdist workers are separate processes. A bare `from claude_agent_sdk import query` inside the function is equally correct.
- **Pre-existing, filed as #354**: the `Run integration tests` job has been failing since before your test file existed. Live testing points at the Legacy Bedrock model id pinned by the two langchain integ tests, and the timing matches. Not yours to fix, but it means a green integ run needs that fixed first.
- **Pre-existing, filed as #355**: `CorrectnessEvaluator` renders a blank `USER QUERY` for any trace with a tool call before the final answer. Already live via the merged `test_adk_eval.py`, so it's a library bug, not a defect in your file — but it's why the second question above matters.
- Not findings against you: the leaked span processor and the collection-error-on-missing-extra both already exist in all three sibling `tests_integ` modules.

</details>

*Automated review — solid work for a human to approve, not a gate. I can be wrong; push back where I am.*

tests_integ/test_claude_openinference_eval.py
```diff
@@ -0,0 +1,267 @@
...
+    tool_spans = [s for t in session.traces for s in t.spans if isinstance(s, ToolExecutionSpan)]
+    assert len(tool_spans) >= 1, "Expected at least one tool execution span from sub-agent"
+    for tool_span in tool_spans:
+        assert tool_span.tool_call.tool_call_id is not None, (
+            f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
+        )
+
+    # Verify agent invocation span is present
+    agent_spans = [s for t in session.traces for s in t.spans if isinstance(s, AgentInvocationSpan)]
+    assert len(agent_spans) >= 1, "Expected at least one AgentInvocationSpan"
```
strandly-the-agent: 🟡 **These assertions don't check that the tools succeeded, and don't prove delegation happened.**

Two gaps, both cheap to close:

**Nothing reads `tool_result.error`.** I mutation-tested this: reverting the mapper line that maps Claude's tool success/error status ships **green through all three tests, 100% of the time** — `tool_result.error` isn't read by any assertion here, nor by any evaluator or judge prompt (grep-confirmed across `evaluators/` and `extractors/`). The live runs prove the blindness the other way too: the tests went green while carrying two denied Bash calls with `error='error'`. One assertion would have caught the `allowed_tools` bug above instantly, with a message naming the actual cause.

**`len(agent_spans) >= 1` is satisfied by a single-agent run.** A real Claude multi-agent run is one OTel trace with exactly one `AgentInvocationSpan` no matter how many specialists ran — verified live and against your fixture (3 delegations → `traces=1 agent_spans=1`). So neither this nor `len(session.traces) >= 1` distinguishes delegation from no delegation. `agent_name` isn't available (the instrumentation sets it from the parent *tool's* name, so it's always `"Agent"`), but the delegation target does live in the tool-call arguments — confirmed present in every live run and in the fixture.

```suggestion
    # Verify tool spans are present and actually succeeded
    tool_spans = [s for t in session.traces for s in t.spans if isinstance(s, ToolExecutionSpan)]
    assert len(tool_spans) >= 1, "Expected at least one tool execution span from sub-agent"
    for tool_span in tool_spans:
        assert tool_span.tool_call.tool_call_id is not None, (
            f"tool_call_id should be populated, got None for tool '{tool_span.tool_call.name}'"
        )
        assert tool_span.tool_result.error is None, (
            f"tool '{tool_span.tool_call.name}' did not execute successfully: "
            f"error={tool_span.tool_result.error!r} content={tool_span.tool_result.content!r}"
        )

    # Verify agent invocation span is present
    agent_spans = [s for t in session.traces for s in t.spans if isinstance(s, AgentInvocationSpan)]
    assert len(agent_spans) >= 1, "Expected at least one AgentInvocationSpan"

    # A single-agent run also satisfies the checks above (one trace, one agent span), so
    # assert the delegation itself — the target is carried in the Agent tool's arguments.
    delegations = [
        s.tool_call.arguments.get("subagent_type") for s in tool_spans if s.tool_call.name == "Agent"
    ]
    assert "math-specialist" in delegations, f"Expected delegation to math-specialist, got {delegations}"
```

tests_integ/test_claude_openinference_eval.py
```diff
@@ -0,0 +1,267 @@
...
+    CorrectnessEvaluator,
+    GoalSuccessRateEvaluator,
+    ToolSelectionAccuracyEvaluator,
+)
+from strands_evals.mappers import OpenInferenceSessionMapper, detect_otel_mapper, readable_spans_to_dicts
+from strands_evals.telemetry import StrandsEvalsTelemetry
+from strands_evals.types.trace import AgentInvocationSpan, Session, ToolExecutionSpan
+
+# Bedrock model for the Claude Agent SDK
+DEFAULT_MODEL = os.environ.get("ANTHROPIC_MODEL", "us.anthropic.claude-sonnet-4-6")
```
strandly-the-agent: 🟡 **`ANTHROPIC_MODEL` is the one env var this file shouldn't read from.**

It's the standard variable Claude Code CLI users export in their shell profile (`export CLAUDE_CODE_USE_BEDROCK=1; export ANTHROPIC_MODEL=...`) — i.e. exactly the audience running this test locally. Anyone with it already set silently redirects the test's model, possibly to an inference-profile ARN or a non-Bedrock-format id that then breaks inside `BEDROCK_ENV`, with nothing in the test aware of it. CI provisions no such variable, so this is a local-dev footgun rather than a CI risk — but it's free to avoid, and line 38 still needs to *set* `ANTHROPIC_MODEL` for the CLI, which is what makes reading the same name here confusing.

The default id itself is fine — I verified `us.anthropic.claude-sonnet-4-6` is invocable in both us-east-1 and us-west-2, and it's a better choice than the dated id the langchain siblings pin (that one now returns a Legacy-model error; see the filed issue).

```suggestion
DEFAULT_MODEL = os.environ.get("STRANDS_CLAUDE_TEST_MODEL", "us.anthropic.claude-sonnet-4-6")
```

pyproject.toml
```diff
@@ -66,6 +66,10 @@ langchain = [
     "opentelemetry-instrumentation-langchain>=0.40.0,<0.62.0",
 ]
 adk = ["google-adk>=2.0.0,<3.0.0"]
+claude = [
+    "claude-agent-sdk>=0.1.0",
+    "openinference-instrumentation-claude-agent-sdk>=0.1.0",
+]
```
strandly-the-agent, line 72: 🟡 **This is the only optional-dependency group in the file with no upper bound.**

`langfuse<4`, `otel<2.0.0`, `adk<3.0.0` all cap theirs. The `>=0.1.0` range here isn't academic — it spans releases with materially different runtime requirements: `0.1.0` is a 32KB pure-Python wheel that genuinely needs Node.js and a separately-installed CLI (which is what the test's docstring describes), while what actually resolves today (`0.2.130`) is a ~286MB wheel bundling its own CLI binary and needing no Node at all. Pinning to what you've actually validated makes the docstring true again and stops a future release silently changing the shape of the dependency.

Worth knowing about line 81 too: `features` is env-level, so adding `"claude"` pulls that 286MB wheel into all 11 unit-test matrix legs, where nothing under `tests/` uses it. That's pre-existing debt (`adk` and `langchain` do the same), and `hatch` does support scoping features per matrix leg via `[tool.hatch.envs.hatch-test.overrides]` — but that's a repo-wide change, not this PR's job.

```suggestion
claude = [
    "claude-agent-sdk>=0.2.0,<0.3.0",
    "openinference-instrumentation-claude-agent-sdk>=0.1.8,<0.2.0",
]
```

liramon2:
@strandly-the-agent Review this PR (integration test only). The plan is to land #340 first and then rebase this on top.

tests_integ/test_claude_openinference_eval.py
```diff
@@ -0,0 +1,250 @@
...
+            tools=["Bash"],
+        ),
+    }
+
+    test_cases = [
+        Case[str, str](
+            name="multi-agent-math",
+            input="Calculate the square root of 1764, then multiply that result by 3.",
+            expected_output="126",
+            expected_assertion="The agent delegated to the math-specialist sub-agent to perform the calculation.",
```
strandly-the-agent: 🔴 **A behavioural assertion can't be graded by `CorrectnessEvaluator` — this now fails deterministically, and nothing checks the answer any more.**

Two consequences of narrowing the assertion to delegation-only:

**1. `CorrectnessEvaluator` fails every run.** In reference mode it treats `expected_assertion` as the *expected response text* and compares it against the actual response — it never sees the tool trace. So it compares "the agent delegated to the math-specialist sub-agent" against `"126"`, finds them unrelated, and concludes the agent didn't delegate. Live at this head:

```
E  AssertionError: Some evaluations failed: [
     "...called Agent({... 'subagent_type': 'math-specialist' ...}), which explicitly
      delegates to a math-specialist sub-agent. The assertion is satisfied.",     # GoalSuccessRate: True
     "The user query is empty/blank... The expected response describes a process/behavior
      (delegation), not a specific numerical answer... The agent did not follow the
      expected pattern of delegating to a sub-agent."                             # Correctness: False
   ]
E  assert False
E   +  where False = all([True, False])
```

Note the irony: the agent *did* delegate — `GoalSuccessRate` saw it in the tool call and passed. Only `Correctness`, which can't see tool calls, got it wrong. (The blank user query is the pre-existing bug in #355; your rewrite turned it from latent into fatal.)

**2. `expected_output="126"` is graded by nothing.** None of these three evaluators reads `Case.expected_output` — only the deterministic `Equals`/`Contains`, `InteractionsEvaluator` and the multimodal evaluators do. Previously the compound assertion checked the answer; now `126` sits in a field no evaluator in this test consults, so the *correctness of the result* is unverified.

The fix solves both at once: let the judges grade the **answer** (which both can see) and prove **delegation** structurally with the assert at `:249`. That's the right tool for each job, and it makes the `CorrectnessEvaluator` in the "one evaluator per extraction level" trio actually meaningful. Verified live: **2/2 green** (37.25s, 41.90s) with this plus the `:249` fix.

```suggestion
            expected_output="126",
            expected_assertion="The final response reports 126 as the answer.",
```

tests_integ/test_claude_openinference_eval.py
```diff
@@ -0,0 +1,250 @@
...
+        assert tool_span.tool_result.error is None, (
+            f"tool '{tool_span.tool_call.name}' did not execute successfully: "
+            f"error={tool_span.tool_result.error!r} content={tool_span.tool_result.content!r}"
+        )
+
+    agent_spans = [s for t in session.traces for s in t.spans if isinstance(s, AgentInvocationSpan)]
+    assert len(agent_spans) >= 1, "Expected at least one AgentInvocationSpan"
+
+    # Prove delegation actually happened
+    delegations = [s.tool_call.arguments.get("agent_name") for s in tool_spans if s.tool_call.name == "Agent"]
```
strandly-the-agent: 🔴 **The key is `subagent_type`, not `agent_name` — so `delegations` is always `[None]` and this assert can never pass.**

The `Agent` tool's arguments carry exactly four keys, and `agent_name` isn't one of them. Verified two ways.

Against your own captured fixture (`tests/strands_evals/mappers/fixtures/claude_live_spans.json`, deterministic — no model needed):

```
Agent-tool span argument keys: ['description', 'prompt', 'run_in_background', 'subagent_type']
  .get("agent_name")     -> None            x3
  .get("subagent_type")  -> 'research-specialist', 'research-specialist', 'math-specialist'

as shipped -> delegations = [None, None, None]   assert "math-specialist" in delegations -> False
with subagent_type -> ['research-specialist', 'research-specialist', 'math-specialist'] -> True
```

And live at this head, with the `:202` issue fixed so the run reaches this line:

```
E  AssertionError: Expected delegation to math-specialist, got [None]
E  assert 'math-specialist' in [None]
```

Worth noting *why* this assert is still worth having despite the trouble: `agent_name` is genuinely the name the OpenInference instrumentation uses elsewhere — but it sets that from the parent *tool's* name, so it's always the literal string `"Agent"`, never the specialist. `subagent_type` in the tool-call arguments is the only place the delegation target survives into the mapped session.

```suggestion
    delegations = [s.tool_call.arguments.get("subagent_type") for s in tool_spans if s.tool_call.name == "Agent"]
```

strandly-the-agent (commented):
**Round 2 at `ca8f340` — changes requested.** Every finding from round 1 is genuinely fixed, and the `allowed_tools` blocker is confirmed dead: I re-ran the multi-agent test live and the specialist's Bash executes cleanly now. Thanks for landing #340 first and rebasing — that answers the merge-path question, nothing further needed there.

But **the rewrite of the multi-agent case introduced two new failures, and the test now fails 100% of the time** — earlier than before. Both are one-liners, both verified live:

1. 🔴 `expected_assertion` is now a pure *behavioural* claim, and `CorrectnessEvaluator` grades it as expected **response text** — so it compares "the agent delegated…" against `"126"`, concludes the agent didn't delegate, and fails. Live: `test_passes=[True, False]`.
2. 🔴 `arguments.get("agent_name")` — the key is `subagent_type`. Live: `got [None]`.

I verified a two-line fix takes it to **2/2 green**. Details in the two inline comments.

<details><summary>✅ Round-1 findings: all six fixed — verified, not assumed</summary>

I re-ran the tests against live Bedrock at this head rather than eyeballing the diff.

| Round-1 finding | Status |
|---|---|
| 🔴 `allowed_tools=["Agent"]` dropped `Bash` | **Fixed and confirmed live.** `["Agent", "Bash"] if agents else ["Bash"]` — the specialist's Bash now executes, zero denials, and the run reaches the new `tool_result.error is None` assert cleanly. |
| 🟡 silent `""` on error / no timeout | **Fixed.** `_run_query()` raising on `is_error` + `asyncio.wait_for(..., timeout=180)`. Both single-agent tests pass in 14.42s combined, so the refactor didn't disturb the happy path. |
| 🟡 `all(report.test_passes)` judge unanimity | **Fixed**, and I verified the index assumption my own suggestion depended on — `report.scores` really is in evaluator order `[GoalSuccessRate, Correctness, ToolSelection]`, so `[:2]` and `scores[2]` slice the right things. |
| 🟡 no `tool_result.error` check | **Fixed** in the multi-agent test (see the nit below about the single-agent one). |
| 🟡 unbounded `claude` extra | **Fixed** exactly as suggested. |
| 🟡 `ANTHROPIC_MODEL` collision | **Fixed** — now `STRANDS_CLAUDE_TEST_MODEL`. |
| ⚪ dead `tools=` param, `conversion_specialist` | **Both removed.** Dropping the unused specialist also removes a delegation ambiguity the coordinator had to resolve on every run. |

Evidence: `pytest tests_integ/test_claude_openinference_eval.py::test_claude_single_query ::test_claude_single_agent_evaluation` → `2 passed in 14.42s`. Multi-agent, patched with the two fixes below → passed in 37.25s and 41.90s (2/2). Multi-agent as shipped → failed in 41.54s and 38.43s (2/2), with the two distinct failures isolated one at a time.

</details>

<details><summary>⚪ Two small leftovers</summary>

- **`:10` — the Node.js requirement is now unambiguously wrong.** It was merely stale before; with the floor raised to `claude-agent-sdk>=0.2.0`, *every* version in range bundles its own CLI binary, so no reader of this file ever needs `npm install -g @anthropic-ai/claude-code`. Suggest: `(claude-agent-sdk bundles its own CLI binary — no separate Node.js/npm install needed.)`
- **`:143` — `test_claude_single_query`'s tool loop still lacks the `tool_result.error is None` assert** you added to the multi-agent test at `:237`. Worth mirroring: that test is the one that would catch a `Bash` permission or tool-result-mapping regression first and most cheaply, and it's the exact assert that would have surfaced the round-1 blocker with an accurate message.

</details>

*Automated review. The two 🔴 are reproduced live and the fix is verified green, but push back if you read it differently.*

liramon2:
@strandly-the-agent Review this PR (integration test only).

strandly-the-agent (commented):
**Round 3 at `62ffc8f` — all clear from me. No new findings, nothing blocking.**

Both round-2 🔴 are fixed and confirmed live, and the whole module is green and stable: **3 passed on two full-module runs** (51.8s, 48.6s, `-n auto` like CI) plus the multi-agent test **3/3 more times on its own — 5/5 total** against live Bedrock.

One of your changes wasn't what I suggested, so I checked it instead of assuming: you **deleted** `expected_assertion` rather than repointing it at the answer. **That's the better call and I'm withdrawing my concern.** It puts both judges in *basic* mode, which reads the full conversation history rather than the blank-`USER QUERY` reference path from #355 — so it sidesteps that bug entirely instead of living next to it. I verified the answer is still genuinely graded (mutation evidence below), so there's no coverage hole.

**On the red X:** `Run integration tests` still fails on this head, and the evidence says that is **not** your test. Your module passes locally; `tests_integ/test_langchain_openinference_eval.py` fails in 1.4s with the LLM call erroring, and the model id it pins is rejected outright in my account — that's #354. Attribution is strong but not proven: I still can't read the CI job logs.

<details><summary>✅ Previous findings — status, verified not assumed</summary>

| Finding | Round | Status |
|---|---|---|
| 🔴 `arguments.get("agent_name")` always `None` | 2 | **Fixed** — now `subagent_type`. Confirmed live: the delegation assert is load-bearing (it fails on an empty list) and it passes, so delegation genuinely happened on every run. |
| 🔴 behavioural `expected_assertion` misgraded by `CorrectnessEvaluator` | 2 | **Fixed** by removing it — see the basic-mode analysis below. |
| ⚪ stale Node.js docstring line | 2 | **Fixed** — line removed. |
| ⚪ missing `tool_result.error` assert in `test_claude_single_query` | 2 | **Fixed** — mirrored at `:138`. Passes live, and round-1 mutation testing already proved this assert discriminates. |
| 🔴 `allowed_tools` dropped `Bash` | 1 | Still fixed — 5/5 runs with zero denied tool calls. |
| 🟡 error handling + 180s timeout, judge-unanimity slicing, `tool_result.error`, dep bounds, env-var collision | 1 | All still in place and exercised green. |

**Why deleting the assertion is fine.** `_has_assertion` / `_has_reference` both key off `expected_assertion`, so removing it routes `GoalSuccessRateEvaluator` and `CorrectnessEvaluator` to `_evaluate_basic`. Basic mode formats the judge prompt via `_format_trace_level_prompt` (`evaluator.py:234-254`), which includes the full conversation history *and* the response — unlike reference mode's `_extract_user_prompt`, which is the one that goes blank (#355).

I mutation-tested that this still grades the answer, using your captured span fixture:

```
CORRECT   ("...difference is 17°F")   CorrectnessEvaluator  pass=True   score=1.0  label='Perfectly Correct'
WRONG(42) ("...difference is 42F")     CorrectnessEvaluator  pass=False  score=0.5  label='Partially Correct'
```

So a wrong answer still fails the test. `expected_output="126"` remains unread by these three evaluators, but it now costs nothing — the judges cover the answer and the structural assert covers the delegation, which is the right split.

</details>

<details><summary>🔍 Evidence, and what I did not verify</summary>

Ran against live Bedrock (us-west-2), `claude-agent-sdk` 0.2.130 with its bundled CLI:

```
full module, -n auto:  3 passed in 51.78s
full module, -n auto:  3 passed in 48.59s
multi-agent alone:     1 passed in 40.23s / 41.43s / 43.51s
```

The #354 attribution:

```
$ pytest tests_integ/test_langchain_openinference_eval.py
  Error in OpenInferenceTracer.on_llm_error callback: TracerException(...)
  3 failed in 1.39s          # LLM call errored; too fast to be a real invocation

$ converse(us.anthropic.claude-sonnet-4-20250514-v1:0)   # pinned at test_langchain_openinference_eval.py:28
  ResourceNotFoundException: ... marked by provider as Legacy ...
$ converse(us.anthropic.claude-sonnet-4-6)               # what this PR uses
  OK
```

**Not verified:** whether CI's account behaves the same (the Legacy rejection is account-scoped and I can't read the job logs), so treat the CI-red attribution as strongly-supported rather than proven. My 5 runs are one account, one region, one day — good evidence of stability, not a calibrated flake rate. I also can't rule out that my local langchain failure is partly an artifact of the unbounded `langchain>=0.3.0` pin resolving to v1.x (there's a `create_react_agent` deprecation warning in the output) — which, if so, is the same class of problem as the bound you just added for `claude`.

**Review shape, disclosed:** this round was a self-run follow-up pass, not the full specialist fan-out — the 10-line delta only touches findings I filed. I verified each fix against the diff rather than the commit message, re-ran the repros, and attacked the fixes themselves (that's how the round-2 `agent_name` regression turned up). If you push behaviour changes beyond these fixes, say so and I'll run the full pipeline.

</details>

One status note, not a request: this PR still carries #340's mapper commits, so merging it today would land the mapper too — consistent with your plan to land #340 first and rebase, just flagging that the rebase hasn't happened yet.

*Automated review. Green here means green in my account on 5 runs — a human should still approve.*

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into 1 comment. I rebased over main and removed the Claude mapper changes from this PR (now on main).

strandly-the-agent (commented):
**Round 4 at `df8e629` — one 🔴, and it's a regression the rebase introduced. Everything from rounds 1–3 is still fixed.** Consolidated into this single comment as you asked.

The PR is genuinely test-only now (2 files, +257/−1) and the merged mapper on main is compatible with every assertion the test makes — I checked that separately, it isn't the problem.

### 🔴 The rebase swapped the model id to the dead one

`tests_integ/test_claude_openinference_eval.py:33` went from `us.anthropic.claude-sonnet-4-6` to **`us.anthropic.claude-sonnet-4-20250514-v1:0`** — the exact legacy id behind #354, which is still open. That id is rejected outright; the one you had works. Verified today, both regions, back to back:

```
us-west-2  us.anthropic.claude-sonnet-4-20250514-v1:0   ResourceNotFoundException: ... marked by provider as Legacy ...
us-west-2  us.anthropic.claude-sonnet-4-6               OK
us-east-1  us.anthropic.claude-sonnet-4-20250514-v1:0   ResourceNotFoundException: ... marked by provider as Legacy ...
us-east-1  us.anthropic.claude-sonnet-4-6               OK
```

And the module itself, at this exact head — the **only** difference between the two runs is the model id:

```
RUN A  as shipped (default id)                          3 failed in 6.71s
       'An error occurred: Claude Code returned an error result: success'  (x2 evaluators)

RUN B  same code, STRANDS_CLAUDE_TEST_MODEL=...-4-6      3 passed in 58.08s
```

This same module was **5/5 green at `62ffc8f`** with the previous id, so this is new. I think I can see the reasoning — `test_langchain_openinference_eval.py:28` pins that id under the comment *"Use the same model that works in CI (same as evaluators use)"* — but that comment is stale on both counts: the evaluators' actual default is `global.anthropic.claude-sonnet-4-6` (`evaluators/evaluator.py:24`), and those langchain tests are precisely the ones that have been failing. Reverting line 33 to `us.anthropic.claude-sonnet-4-6` is the whole fix.

```suggestion
DEFAULT_MODEL = os.environ.get("STRANDS_CLAUDE_TEST_MODEL", "us.anthropic.claude-sonnet-4-6")
```

**Caveat, stated plainly:** my sandbox account is not CI's, and the Legacy rejection is account- and recency-scoped ("you have not been actively using the model in the last 30 days"), so I can't prove CI hits it. `Run integration tests` is red on `df8e629`, but that tells us nothing about your file either way — the langchain tests would redden it regardless, which is #354. If CI's account *does* still have access to the legacy id, the change is still a downgrade to a model AWS has marked for retirement.

<details><summary>✅ Rounds 1–3: all findings still fixed, and re-exercised at this head</summary>

RUN B above executes the whole module, so every one of these was exercised green, not merely eyeballed:

| Finding | Round | Status at `df8e629` |
|---|---|---|
| 🔴 `allowed_tools=["Agent"]` dropped `Bash` | 1 | Fixed — sub-agent Bash executes, zero denials |
| 🟡 silent `""` on error / no timeout | 1 | Fixed — and it's why RUN A fails in 6.7s instead of hanging |
| 🟡 judge-unanimity gate | 1 | Fixed — `test_passes[:2]` + `scores[2] >= 0.5` |
| 🟡 no `tool_result.error` assert | 1 | Fixed — present in both tests |
| 🟡 unbounded `claude` extra | 1 | Fixed — bounds intact after the rebase |
| 🟡 `ANTHROPIC_MODEL` env collision | 1 | Fixed — still `STRANDS_CLAUDE_TEST_MODEL` |
| 🔴 `arguments.get("agent_name")` → always `None` | 2 | Fixed — `subagent_type` |
| 🔴 behavioural `expected_assertion` misgraded | 2 | Fixed — removed; basic-mode judges still grade the answer |
| ⚪ Node.js docstring, ⚪ mirrored error assert | 2/3 | Both fixed |

**Merged-mapper compatibility (main now carries #340).** The mapper on main is not byte-identical to the branch version I originally verified against, so I re-checked rather than assumed — mapped your captured fixture through main's merged mapper and ran the test's assertion bodies against the result:

```
mapper: OpenInferenceSessionMapper  traces: 1  tool_spans: 6  agent_spans: 1
all tool_call_id populated : True
all tool_result.error None  : True
subagent_type delegations   : ['research-specialist', 'research-specialist', 'math-specialist']
```

All three assertion families still hold. Combined with RUN B passing, the merged mapper is compatible — the model id is the sole cause of RUN A's failure.

</details>

<details><summary>⚪ Appendix — non-blocking: the AWS_REGION change is dead code</summary>

`:38` changed the fallback from `us-east-1` to `us-west-2`, but that fallback is unreachable in a pytest run: `conftest.py:37-38` does `os.environ.setdefault("AWS_REGION", "us-east-1")` in `pytest_sessionstart`, which runs *before* collection imports this module, so `os.environ.get("AWS_REGION", "us-west-2")` always finds `us-east-1` already set. CI also sets it explicitly (`integration-task — integration-test.yml:141`). Verified:

```
AWS_REGION after conftest sessionstart : us-east-1
-> BEDROCK_ENV AWS_REGION would be     : us-east-1
```

Harmless as-is. But if the intent was to run this test in `us-west-2` (e.g. for model availability), it doesn't do that — you'd need to override `AWS_REGION` itself, and the region isn't the reason the legacy id fails anyway (it's rejected in both).

</details>

*Automated review. Everything above is reproduced in my sandbox against live Bedrock; the CI-account caveat is real, so a human should confirm which model ids the integ role can actually reach.*

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into 1 comment. The model id is intentionally Sonnet 4 to match the other integration tests.

strandly-the-agent (commented):
**Round 5 at `05fe93f` — taking the model id as settled. One 🟡 on the `async_launched` exemption, which I think weakens the multi-agent test.** Single comment as asked.

> **Updated after posting** — I found a way to run this at your head after all (patching the model constant locally), so the `async_launched` finding below now carries live evidence and is sharper than when first posted: **the exemption never fires in the happy path, and the assert I'm suggesting passes 4/4.** The original version said this round had no runtime evidence; that no longer holds.

**On the model id: your call, I'm not re-raising it.** Matching the sibling integ tests is a legitimate consistency argument and you own that decision. Two consequences for the record, then I'll drop it: (1) that id returns `ResourceNotFoundException ... marked by provider as Legacy` in my account in both us-east-1 and us-west-2, and `05fe93f` also removed the `STRANDS_CLAUDE_TEST_MODEL` escape hatch, so running this test now requires editing the source; (2) `Run integration tests` is red here, but that proves nothing about your file — the langchain tests redden it regardless (#354). If the integ role can invoke `us.anthropic.claude-sonnet-4-20250514-v1:0`, my last round is moot; that's a one-command check in CI's account and you're better placed to run it than I am.

### 🟡 The `async_launched` exemption lets the multi-agent test pass with no successful tool execution

First, what I measured. I ran the multi-agent scenario 4x at this head with **only** the model constant swapped for one my account can reach:

```
run 1: tool_spans=[('Bash', None), ('Bash', None), ('Agent', None)]  async_launched=False  successful non-Agent tools=2
run 2: tool_spans=[('Bash', None), ('Bash', None), ('Agent', None)]  async_launched=False  successful non-Agent tools=2
run 3: tool_spans=[('Bash', None), ('Bash', None), ('Agent', None)]  async_launched=False  successful non-Agent tools=2
run 4: tool_spans=[('Bash', None), ('Bash', None), ('Agent', None)]  async_launched=False  successful non-Agent tools=2
```

Two things follow. **The `Agent` span reports success (`error=None`), not `async_launched`, in 4/4 runs** — so on this path `:241` is dead code, and whatever produced `async_launched` for you was a different condition (a backgrounded launch? a different model or turn count?). Worth knowing which, because that's the path the test is now blind on. And **the sub-agent's `Bash` spans are present and successful** — 2 per run, every run.

That second fact is what makes the gap worth closing rather than theoretical. `continue` removes the only check that the delegated work ran, and every remaining assertion is satisfiable by the `Agent` span **alone**:

| assertion | satisfied by the `Agent` span by itself? |
|---|---|
| `:235` `len(tool_spans) >= 1` | yes — the `Agent` call is a tool span |
| `:238` `tool_call_id is not None` | yes |
| `:243` `tool_result.error is None` | **skipped** by the new `continue` |
| `:249` `len(agent_spans) >= 1` | yes — that's the root agent span |
| `:252` `"math-specialist" in delegations` | yes — read from the `Agent` call's own arguments |

So when the exemption does fire, the test can go green having verified only that a delegation was *requested* — the same shape of hole as the round-1 blocker, where the specialist's `Bash` was silently denied and the test still looked fine.

The fix keeps your exemption and restores the guarantee. Given the measurements above it passes in the normal path (2 successful non-`Agent` tools, 4/4), so it adds no flakiness:

```suggestion
        # The Agent tool reports "async_launched" as an operational status, not a failure.
        if tool_span.tool_call.name == "Agent" and tool_span.tool_result.error == "async_launched":
            continue
        assert tool_span.tool_result.error is None, (
            f"tool '{tool_span.tool_call.name}' did not execute successfully: "
            f"error={tool_span.tool_result.error!r} content={tool_span.tool_result.content!r}"
        )

    # The exemption above can hide a delegation whose sub-agent work never ran, so require
    # at least one non-Agent tool to have actually succeeded.
    executed = [s for s in tool_spans if s.tool_call.name != "Agent" and s.tool_result.error is None]
    assert executed, (
        "Expected at least one successful sub-agent tool execution, got "
        f"{[(s.tool_call.name, s.tool_result.error) for s in tool_spans]}"
    )
```

**One question I couldn't answer myself:** when you saw `async_launched`, were the sub-agent's `Bash` spans still in the trace? I never reproduced that state, so I don't know whether it coexists with real sub-agent work. If it does, the assert above costs nothing and locks the behaviour in; if it doesn't, the test needs it more, not less.

<details><summary>🟡 The real fix probably belongs in the mapper, not here</summary>

The mapper decides this at `src/strands_evals/mappers/openinference_session_mapper.py:562`:

```python
# "success" = smolagents/LangChain default; "completed" = Claude Agent SDK
error=None if tool_status in ("success", "completed") else tool_status,
```

Anything outside that two-value set becomes an `error`, so `async_launched` is reported as a tool failure to every consumer — not just this test, but any judge or report reading `tool_result.error`. Grepping the tree, `async_launched` appears **only** in your new test line; the mapper (now on main) has no notion of it, so every future caller hits the same trap and needs the same per-call-site workaround.

If `async_launched` is a real Claude Agent SDK status, it arguably belongs in that non-error set, or `ToolResult` needs a status field distinct from `error`. Happy to file that against main if you agree it's a mapper gap — out of scope for this PR either way, and your test-level workaround is a fine stopgap.

</details>

<details><summary>✅ Rounds 1–4 status, and two ⚪ notes on this commit</summary>

All prior findings are still in place at `05fe93f` (verified against the diff, not the commit message): `allowed_tools = ["Agent", "Bash"]`, the `is_error` raise + 180s `wait_for`, `test_passes[:2]` / `scores[2] >= 0.5`, `tool_result.error` asserts in both tests, `subagent_type`, no `expected_assertion`, dependency bounds, and the Node.js docstring line gone. The 4 runs above exercise all of them green.

One changed shape rather than regressing: round 1's `ANTHROPIC_MODEL` collision is now resolved *more* firmly — nothing is read from the environment, so ambient shell config can't redirect the test. The cost is the lost escape hatch.

Two ⚪ observations, both non-blocking:

- **`:85` lost the comment explaining why `Bash` is in `allowed_tools`.** That line looks redundant (`AgentDefinition(tools=["Bash"])` appears to grant it) and it isn't — dropping `Bash` here is what made this test fail ~83% of the time in round 1, because `allowed_tools` is the permission allow-list and sub-agent calls are checked against it. A future reader is fairly likely to "simplify" it back; one line of comment is cheap insurance.
- **No override for the model id.** With the constant hardcoded, a contributor whose account can't reach a Legacy-marked model can't run this test without editing the source — which is exactly what I had to do to produce the evidence above. Same tradeoff as the sibling tests, so it may be the consistency you want; just noting it's now the only integ module whose model is unreachable from at least one real account.

</details>

*Automated review. The runs above are my sandbox account against live Bedrock on a locally-patched model constant — not your CI configuration, so weight them accordingly and push back where you have runtime knowledge I don't.*

liramon2:
Update 8/11:
- Prerequisite (#329) has been merged as of today
- Currently refining with Strandly

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into 1 comment.

strandly-the-agent (commented):
**Round 6 at `359d93b` — the mapper fix is the right call. One 🟡: it ships without a unit test.** Single comment as asked.

> **Updated after posting** with the runs I said I'd add: **mapper unit suite 359 passed**, and the **integ module 3 passed live** at this head. Both confirm the change works — and the 359 make the finding below sharper, not softer: a thorough suite, zero coverage of the new status.

You fixed the root cause instead of the symptom — `_TOOL_SUCCESS_STATUSES` at `openinference_session_mapper.py:60` used at `:567`, with the per-test `continue` deleted and the `executed` guard added at `:245`. That's better than what I suggested: every consumer of `tool_result.error` now sees `async_launched` correctly, not just this test.

But that also changes what's in scope. This is no longer a test-only PR — it's a **behaviour change to a mapper on `main` that serves smolagents, LangChain and Claude** — and it has no unit test. Its only coverage is a live integration test that, by your own model-id decision, can't run outside CI's account. The repo already has the exact pattern to copy.

<details><summary>🟡 Add a unit test for the new status — paste-ready</summary>

`tests/strands_evals/mappers/test_openinference_session_mapper.py:1684`'s `test_claude_tool_span_content_blocks_output` is the template (`make_span` at `:29`, `CLAUDE_SDK_SCOPE_NAME` at `:26`), and `:1775` already covers the failure direction. A near-copy pins the new behaviour:

```python
    def test_claude_async_launched_tool_status_is_not_an_error(self):
        """async_launched is an operational status for async sub-agent delegation, not a failure."""
        span = make_span(
            name="Agent",
            scope_name=CLAUDE_SDK_SCOPE_NAME,
            attributes={
                "openinference.span.kind": "TOOL",
                "tool.id": "toolu_bdrk_async1",
                "tool.name": "Agent",
                "input.value": json.dumps(
                    {"description": "Delegate", "subagent_type": "math-specialist", "prompt": "sqrt(1764)*3"}
                ),
                "output.value": json.dumps({"status": "async_launched", "content": [{"type": "text", "text": "launched"}]}),
            },
        )
        session = self.mapper.map_to_session([span], "sess-1")

        tool_spans = [s for t in session.traces for s in t.spans if isinstance(s, ToolExecutionSpan)]
        assert len(tool_spans) == 1
        assert tool_spans[0].tool_result.error is None
```

Why it's worth the five minutes: I ran the mapper suite at this head and **359 tests pass, none of which exercises `async_launched`** — so the suite is thorough everywhere except the line you just changed. This is also the third consecutive round where a change to that exact status-mapping line had a consequence nobody predicted (round 1: nothing read `tool_result.error` at all; round 5: `async_launched` misfiled as an error; now: the success set has grown). A unit test is what stops the next change from silently undoing this one — and unlike the integ test, it runs on every PR in every account.

</details>

<details><summary>⚪ Two smaller notes, and one risk in my own suggestion</summary>

- **The success set is global across producers.** `async_launched` is Claude-specific, but `:60` whitelists it for smolagents and LangChain traces too. If either ever emits that status meaning something else, it would now be silently treated as success. The comment block documents provenance per value, which mitigates it, and per-scope sets would be over-engineering for three strings — I'd leave it, just noting the tradeoff is now baked in.
- **Title and label no longer describe the PR.** Title is `feat: add Claude Agents OpenInference integration tests` and the label is `chore`, but the diff now includes a src/ bug fix. Cosmetic; `fix:` would match the actual content, and the commit message ("fix success status bug caught by integration test") already says it well.
- **A risk I introduced, owning it:** `assert executed` (`:245`) requires a *non-`Agent`* tool span to have succeeded, so the test now depends on the specialist actually calling `Bash` rather than doing the arithmetic in its head. It passed in the live run below, and 4/4 runs last round produced 2 successful `Bash` spans, so the risk looks small — but it is model-dependent, and if you ever see this assert fail on an otherwise-correct run, that's the reason. Loosening it would trade the guarantee away, so I'd keep it and treat a failure as signal.

</details>

<details><summary>✅ What I verified this round</summary>

- **Read the diff, not the commit message.** `05fe93f → 359d93b` touches exactly two files: the mapper and the integ test. **No file under `tests/` was modified** — that's the basis for the finding above.
- **Mapper unit suite at this head: `359 passed` in 3.02s.** No existing test regressed from widening the success set. I confirmed the code under test really was your fix rather than a stale install — the imported module contains `_TOOL_SUCCESS_STATUSES` with `async_launched`, and the old `in ("success", "completed")` form is gone. CI agrees so far: green on all five Windows legs (Python 3.10–3.14) when I looked, with linux/macOS still queued.
- **Full integ module live at this head: `3 passed` in 57.69s** (my account, with the model constant patched locally so it can run at all). So the mapper fix and the new `executed` assert both hold in the happy path.
- **The exemption removal is safe *because* of the mapper fix** — the loop's `assert tool_result.error is None` at `:240` now passes for an `async_launched` Agent span, where before the fix it would have failed. The two halves of this commit depend on each other; neither is correct alone.
- **`executed` placement is right** — after the loop, before the agent-span checks, and it still closes the degenerate case I raised in round 5 (a trace containing only the `Agent` span).
- **All rounds 1–5 findings remain in place:** `allowed_tools = ["Agent", "Bash"]`, the `is_error` raise + 180s `wait_for`, `test_passes[:2]` / `scores[2] >= 0.5`, `tool_call_id` + `tool_result.error` asserts in both tests, `subagent_type`, no `expected_assertion`, dependency bounds, Node.js line gone.
- **Caveat that stands:** the live run required patching the hardcoded model id, so it exercised `us.anthropic.claude-sonnet-4-6`, not the id CI will use. It validates the logic, not your CI configuration.

**Review shape:** self-run follow-up pass — this delta implements the two changes I proposed, so I checked it myself rather than re-running the specialist fan-out, and I tried to be as sceptical of my own `executed` suggestion as of your code (hence the third note above).

</details>

*Automated review — solid work for a human to approve. The mapper change is small but it's on a shared path; a maintainer should sanity-check the global-success-set tradeoff.*

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into 1 comment.

strandly-the-agent (commented):
**Round 7 at `d26fa4f` — all clear. No findings, and I owe you a correction.** Single comment as asked.

The unit test closes the last open item, and I confirmed it's a real guard rather than a passenger: **mutation-tested it**, and it discriminates.

```
baseline (as shipped)                                        360 passed in 3.03s
mutant: drop "async_launched" from _TOOL_SUCCESS_STATUSES    1 failed  — test_claude_async_launched_tool_status_is_not_an_error
                                                             AssertionError at test_...mapper.py:1741
restored                                                     1 passed
```

That's the whole delta since my last review (`359d93b → d26fa4f` touches only the unit-test file), so there's nothing else to review.

### I was wrong about the model id, and you were right

**`Run integration tests` is `SUCCESS` on this head** — the first green integ run across all seven rounds. Since that job runs `hatch test tests_integ` over the whole directory, it proves the integ role *can* invoke `us.anthropic.claude-sonnet-4-20250514-v1:0`, which is exactly the fact I said I couldn't verify. So my round-4 🔴 was wrong for the environment that actually matters, and "intentionally Sonnet 4 to match the other integration tests" was the correct call. The Legacy rejection I kept hitting is specific to my sandbox account.

Two consequences I'll own rather than leave lying around:

- It also means the langchain integ tests pass in CI, so the premise of #354 (that this id is what's been reddening the job) is refuted for CI's account. I'll update that issue with this evidence rather than leave a misleading bug report open in your tracker.
- Worth knowing *why* I couldn't confirm it: my token can't read the integ job's logs, so for six rounds "red" was an opaque signal I could only reason around. If you ever want a faster loop from me on integ failures, that permission is the bottleneck.

<details><summary>✅ Full status — everything raised across 7 rounds</summary>

| Round | Finding | Status at `d26fa4f` |
|---|---|---|
| 1 | 🔴 `allowed_tools=["Agent"]` denied sub-agent `Bash` (10/12 live failures) | fixed |
| 1 | 🟡 silent `""` on error, no timeout | fixed (`is_error` raise + 180s `wait_for`) |
| 1 | 🟡 judge-unanimity gate | fixed (`test_passes[:2]`, `scores[2] >= 0.5`) |
| 1 | 🟡 nothing asserted `tool_result.error` | fixed in both tests |
| 1 | 🟡 unbounded `claude` extra | fixed (`>=0.2.0,<0.3.0`) |
| 1 | 🟡 `ANTHROPIC_MODEL` env collision | fixed (nothing read from env now) |
| 2 | 🔴 `arguments.get("agent_name")` always `None` | fixed (`subagent_type`) |
| 2 | 🔴 behavioural `expected_assertion` misgraded by `CorrectnessEvaluator` | fixed (removed; basic mode still grades the answer) |
| 2/3 | ⚪ stale Node.js docstring, mirrored error assert | fixed |
| 4 | 🔴 model id → legacy | **withdrawn — I was wrong, see above** |
| 5 | 🟡 `async_launched` exemption hollowed out the multi-agent test | fixed at the root, in the mapper |
| 6 | 🟡 mapper change had no unit test | fixed, and mutation-verified above |

Also green at this head: unit tests on all 10 linux+windows legs (Python 3.10–3.14) and Lint; macOS was still queued. Locally, 360 mapper tests pass, and the integ module was 3/3 green at `359d93b` (the integ file is byte-identical here, so I didn't spend another live run re-proving it).

**Two standing non-blockers**, unchanged and not worth another round: the success set at `openinference_session_mapper.py:60` is global across producers (documented tradeoff, I'd leave it), and the PR title/label (`feat:` / `chore`) don't quite describe a diff that now includes a src/ fix.

**One risk I introduced and still own:** `assert executed` makes the multi-agent test depend on the specialist actually calling `Bash`. Green in CI and in 5 local runs, but model-dependent — if it ever fails on an otherwise-correct run, that's why.

**Review shape:** self-run follow-up pass; the delta implemented my own suggestion, so I verified it by mutation rather than re-running the specialist fan-out.

</details>

Nothing further from me — this looks ready for a human to approve.

poshinchen (approved):

PR #356, opened 2026-08-08
fix: CorrectnessEvaluator honors expected_output for reference-mode grading

### LINKED ISSUE #334: [BUG] CorrectnessEvaluator ignores Case.expected_output and silently falls back to trajectory grading
### Summary

`CorrectnessEvaluator` never reads `Case.expected_output`. When a case sets `expected_output`
but not `expected_assertion`, the evaluator silently switches to trajectory-based grading and
can return CORRECT for an answer that contradicts the expected output.

`_has_reference()` gates reference mode on `expected_assertion`, and
`_format_reference_prompt()` interpolates `expected_assertion` into `EXPECTED RESPONSE:`.
`grep expected_output src/strands_evals/evaluators/correctness_evaluator.py` returns nothing
(verified on `main` @ 0bf1252).

This matters because `Case.expected_output` is documented as *"The expected response given the
input"*, which reads like the natural reference for a correctness evaluator. Setting it does
not fail, warn, or log - the evaluation silently becomes a different, weaker check.

### Steps to Reproduce

An agent has a structured lookup tool and a document-reading tool. The structured tool returns
a coarse value; the document contains a more precise one. The case asserts the precise version.

```python
from strands_evals.evaluators import CorrectnessEvaluator

case = Case(
    name="warranty-period",
    input="What is the warranty period for my device?",
    expected_output="24 months from date of purchase, excluding batteries and consumables.",
    # expected_assertion deliberately unset
)

def task(case):
    # get_product() returned {"warranty": "2 years"}; the agent repeated it without
    # opening the manual, which is the error the case exists to catch.
    return {
        "output": "Your device has a warranty of 2 years.",
        "trajectory": mapped_session,  # StrandsInMemorySessionMapper output
    }

report = Experiment(
    cases=[case],
    evaluators=[CorrectnessEvaluator(model="<an inference profile>")],
).run_evaluations(task)
```

### Expected Behavior

One of:

1. Reference mode uses `expected_output` when it is set, comparing the answer against it; or
2. The evaluator raises or logs a warning that `expected_output` is ignored, so the silent
   downgrade to basic mode is visible.

Either way, an answer that contradicts `expected_output` should not score 1.0.

### Actual Behavior

`_has_reference()` is False, so `_evaluate_basic()` runs. The basic prompt includes the trace
context, so the judge grades the answer against the tool results and returns CORRECT because
the answer faithfully repeats them.

The judge's `reason` cited agreement with the tool output as its justification - in substance
(paraphrased, values replaced):

> The candidate response matches the tool data exactly: warranty 2 years. This correctly
> answers the user's question about the warranty period.

Score 1.00, `test_pass: true`. The report row for the case shows `expected_output` present
with the expected value, so the data did reach the evaluator and was discarded. Three
deterministic evaluators on the same output failed it correctly, which is how we noticed.

The consequence is that a suite relying on `CorrectnessEvaluator` + `expected_output` can
report a passing agent while the agent is wrong - and specifically, it validates the answer
against the very tool data that made it wrong.

### Additional Context

Documentation reachable from the quickstart describes the evaluator as *"Factual correctness
with optional reference comparison"*, which does not indicate that the reference is
`expected_assertion` rather than `expected_output`.

Versions: `strands-agents-evals` 1.0.3 (latest). `strands-agents` pinned at 1.47.0 in our
project rather than 1.50.2; the behaviour was verified directly against evals `main` @ 0bf1252
and the code path is internal to evals, independent of the `strands-agents` version.

Workaround: use `GoalSuccessRateEvaluator` with `expected_assertion`, stating the failure
condition explicitly in the assertion text.

### Possible Solution

Prefer `expected_assertion` and fall back to `expected_output` in both places:

```python
def _reference(self, evaluation_case):
    return evaluation_case.expected_assertion or evaluation_case.expected_output
```

If ignoring `expected_output` is intended, a warning when it is set with no
`expected_assertion`, plus a note in the evaluator docs, would remove the silent failure.

AmirF194:
## Description

`CorrectnessEvaluator` only enters reference-mode grading when `expected_assertion` is set on the evaluation case. `expected_output` is a separate, documented field on `EvaluationData` ("the expected response given the input"), but nothing in `correctness_evaluator.py` ever reads it: `_has_reference()` checks `expected_assertion` alone, and `_format_reference_prompt()` interpolates only `expected_assertion` into the judge prompt. A case that sets `expected_output` without also setting `expected_assertion` silently falls through to `_evaluate_basic()`'s trajectory-only grading, which can score a response CORRECT even when it contradicts the expected output, with no warning to the caller.

Fix: `_has_reference()` now treats either field as a reference, and `_format_reference_prompt()` prefers `expected_assertion` when both are set (unchanged behavior for existing callers) and falls back to `expected_output` when only that is present.

## Related Issues

Fixes #334

## Documentation PR

None; docstring updated in place, no separate docs page describes this behavior.

## Type of Change

Bug fix

## Testing

- `tests/strands_evals/evaluators/test_correctness_evaluator.py::test_has_reference_true_with_expected_output_only`, `test_format_reference_prompt_uses_expected_output_when_no_assertion`, and `test_evaluate_with_expected_output_only_uses_reference_mode` fail on main and pass with this change.
- Full suite passes: `hatch test tests --cover` (1760 passed).
- `hatch run test-lint` (ruff check + mypy) is clean.
- Not checked: behavior against a live model call in reference mode; the new tests mock the judge model the same way the existing reference-mode tests do.

- [ ] I ran `hatch run prepare` (ran its constituent checks separately: `hatch run test-lint` and `hatch test tests --cover`, not the formatter step, to avoid reformatting unrelated pre-existing files)

## Checklist

### DIFF
--- a/src/strands_evals/evaluators/correctness_evaluator.py
+++ b/src/strands_evals/evaluators/correctness_evaluator.py
@@ -51,9 +51,9 @@ class CorrectnessEvaluator(Evaluator[InputT, OutputT]):
     - **Basic mode**: Evaluates correctness based on the conversation context and the
       assistant's response alone. Uses a 3-level scoring rubric
       (Perfectly Correct=1.0, Partially Correct=0.5, Incorrect=0.0).
-    - **Reference mode**: When `expected_assertion` is provided on the evaluation case,
-      evaluates whether the agent's response is correct by comparing it to the reference.
-      Uses a binary CORRECT/INCORRECT scoring rubric (CORRECT=1.0, INCORRECT=0.0).
+    - **Reference mode**: When `expected_assertion` or `expected_output` is provided on the
+      evaluation case, evaluates whether the agent's response is correct by comparing it to
+      the reference. Uses a binary CORRECT/INCORRECT scoring rubric (CORRECT=1.0, INCORRECT=0.0).
     """
 
     evaluation_level = EvaluationLevel.TRACE_LEVEL
@@ -88,8 +88,9 @@ class CorrectnessEvaluator(Evaluator[InputT, OutputT]):
         self.model = model
 
     def _has_reference(self, evaluation_case: EvaluationData[InputT, OutputT]) -> bool:
-        """Check if the evaluation case contains an expected_assertion for reference-based evaluation."""
-        return bool(evaluation_case.expected_assertion)
+        """Check if the evaluation case contains an expected_assertion or expected_output for
+        reference-based evaluation."""
+        return bool(evaluation_case.expected_assertion) or bool(evaluation_case.expected_output)
 
     def evaluate(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
         parsed_input = self._get_last_turn(evaluation_case)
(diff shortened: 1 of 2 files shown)


AmirF194:
Checking in after a week, no rush. Two of the checks are held pending an internal approval gate on your side (integration tests for external PRs), everything else that has run is green.

AmirF194:
Closing this after a month with no review capacity here. The CI approval gate for external PRs never got triggered and the deployment protection rule expired on its own. Reopening later is fine if this becomes relevant.

PR #357, opened 2026-08-10
feat: add could-not-evaluate status so non-gradable cases are excluded from scores

### LINKED ISSUE #346: [FEATURE] Could-Not-Evaluate verdict for evaluators that cannot score a case
### Problem Statement

When an evaluator cannot produce a meaningful score for a given case, it has no way to say so. The only output shape is `EvaluationOutput(score=..., test_pass=..., reason=...)`, which forces every result into the pass/fail binary. An evaluator that legitimately cannot grade a case (missing input data, inapplicable rubric, harness error) must pick a score anyway, and that score pollutes aggregates.

Take an error-handling evaluator. If the agent's trajectory contains no tool errors, there is nothing to evaluate. Today the evaluator has three bad options:

1. Return `score=1.0, test_pass=True` and inflate the pass rate with a hollow pass.
2. Return `score=0.0, test_pass=False` and penalize a correct agent for something that never happened.
3. Encode "could not evaluate" into the `reason` string and hope downstream consumers parse it out.

Option 3 is what I currently do. It works, but it's fragile and invisible to the framework. The `_default_aggregator` on `Evaluator` computes a straight average:

```python
avg_score = sum(o.score for o in outputs) / len(outputs)
```

There is no way to exclude a result from this sum without the aggregator parsing `reason` strings. That coupling should not exist.

The same problem shows up at the `Experiment` level. When building the `EvaluationReport`, scores are averaged directly:

```python
overall_score=sum(scores) / len(scores) if scores else 0,
```

A non-gradable result at `score=0.0` drags the average down. A non-gradable result at `score=1.0` inflates it. Either way the aggregate lies.

### Proposed Solution

Add an optional `status` field to `EvaluationOutput` that defaults to `"graded"`:

```python
class EvaluationOutput(BaseModel):
    score: float
    test_pass: bool
    reason: str | None = None
    label: str | None = None
    status: str = "graded"  # "graded" | "could_not_evaluate" | "informational"
```

Three values:

- `"graded"` (default) means the score and test_pass are real verdicts. Backward compatible with every existing evaluator and consumer.
- `"could_not_evaluate"` means the evaluator tried to grade but could not. Preconditions not met, harness failure, missing data. Aggregators should exclude this from pass-rate and score math.
- `"informational"` means the evaluator does not pass/fail by design. It surfaces content for human review but should never count toward a numeric aggregate.

The field is optional and defaults to `"graded"`, so existing code keeps working with zero changes.

The `_default_aggregator` in `evaluator.py` would filter before computing:

```python
@staticmethod
def _default_aggregator(outputs: list[EvaluationOutput]) -> tuple[float, bool, str]:
    graded = [o for o in outputs if o.status == "graded"]
    if not graded:
        return (0.0, False, "No gradable evaluation outputs produced")
    avg_score = sum(o.score for o in graded) / len(graded)
    all_pass = all(o.test_pass for o in graded)
    combined_reason = " | ".join(o.reason for o in graded if o.reason)
    return avg_score, all_pass, combined_reason
```

The `Experiment` report assembly in `experiment.py` would also need to exclude non-graded rows when computing `overall_score`. Today it does `sum(scores) / len(scores)`. With this feature, it would filter on the status carried through from the aggregated outputs. Concretely, the `_run_evaluator` method already calls `evaluator.aggregator(outputs)`, and since non-graded outputs would be excluded at the aggregator level, the per-evaluator score would already be correct. The remaining gap is when multiple evaluators are flattened via `EvaluationReport.flatten()`, which also averages all scores. Adding `status` to `EvaluationOutput` (and propagating it into the report's `cases` dict) would let `flatten()` skip non-graded rows too.

An evaluator producing a non-gradable result would look like:

```python
def evaluate(self, evaluation_case: EvaluationData) -> list[EvaluationOutput]:
    if not self._has_errors_in_trajectory(evaluation_case):
        return [EvaluationOutput(
            score=0.0,
            test_pass=False,
            reason="No tool errors occurred in trajectory.",
            status="could_not_evaluate",
        )]
    # ... actual grading logic
```

### Use Case

1. An error-handling evaluator that only applies when tool errors exist. No errors means nothing to score. Today it must fake a pass or a fail. With `status="could_not_evaluate"` the result gets excluded from aggregates and surfaced separately in reports.

2. A faithfulness evaluator where the agent declined to answer (empty output). There is no claim to ground-check. Instead of scoring 1.0 or 0.0, it returns CNE and the aggregate reflects only cases where faithfulness was actually testable.

3. A "referenced tickets" evaluator that lists which tickets the agent mentioned but does not pass/fail. It produces `status="informational"` results that show up in a report for human review without counting toward the suite's pass rate.

4. Context-window overflow during judging. Issue #342 describes judge-based evaluators scoring 0 when a trace exceeds the context window. The error-isolation catch block could set `status="could_not_evaluate"` instead of recording `score=0, test_pass=False`, giving downstream consumers a clean signal that the 0 is a harness failure, not a quality failure.

### Alternatives Considered

A convention on `score` value (e.g. score=-1.0 means skip). I do this today. It works until someone writes an evaluator that legitimately scores on a scale including negative numbers, or a consumer forgets to filter. A dedicated field is explicit and impossible to misinterpret.

A convention on `reason` prefix (e.g. `"[Could Not Evaluate] ..."`). Also what I do today. Brittle, requires string parsing, and invisible to typed consumers. The framework should not require regex on a reason string to compute a correct average.

A separate `EvaluationSkip` type or exception. More disruptive. Evaluators return `list[EvaluationOutput]`, and that contract is baked into every evaluator subclass and into `_run_evaluator`'s aggregation flow. Adding a union type or exception would require every consumer to handle a new branch. An optional field on the existing model is the smallest possible change.

### Additional Context

- Issue #342 identifies the same root cause from the harness-error angle (context-window overflow producing `score=0`). This feature gives #342 a clean landing spot on the model layer. The two are complementary.
- Issue #247 (skill evaluation) discusses aggregation of paired runs. Aggregation that correctly excludes non-gradable results is relevant there too, though #247 does not strictly depend on this landing first.
- The field is a plain string rather than an enum to allow future extension without a breaking change. A `Literal["graded", "could_not_evaluate", "informational"]` type annotation is fine for documentation and tooling while remaining forward-compatible.

pdebjyot:
## What

Adds an `EvaluationOutput.status` field so an evaluator can signal that it could **not** produce a real verdict for a case, and excludes those non-gradable outputs from every score/pass aggregate. Implements the design in #346.

`status` is one of:
- `"graded"` (default) — `score`/`test_pass` are real verdicts and count toward aggregates.
- `"could_not_evaluate"` — the evaluator tried but couldn't grade the case (preconditions not met, missing data, harness error). Excluded from aggregates.
- `"informational"` — surfaces content for human review; never counts toward a numeric aggregate.

The default is `"graded"`, so existing evaluators, consumers, and previously-saved reports are unaffected.

## Why

Today, when an evaluator fails — a judge context-window overflow, missing data, unmet preconditions — the case is recorded as `score=0, test_pass=False` and that `0` is thrown into `overall_score`. A correct agent gets silently punished because the *judge* couldn't score the case, not because the agent was wrong. A capability failure of the harness is reported as a quality failure of the agent, with no way to tell the two apart.

## How

- **`types/evaluation.py`** — add `GRADED` / `COULD_NOT_EVALUATE` / `INFORMATIONAL` constants (plain `str`, not an `Enum`, so new values can be added without a breaking change) and the `status` field on `EvaluationOutput` (`Literal[...]`, default `graded`).
- **`evaluators/evaluator.py`** — `_default_aggregator` averages graded outputs only. When every output is non-graded there is nothing to score, so it returns a non-failure sentinel with the evaluator's own reason string.
- **`experiment.py`** — roll per-output statuses up to an evaluator-level status; tag both error-isolation paths (`RetryError` and generic `Exception`) as `could_not_evaluate`; compute `overall_score` from graded rows only.
- **`types/evaluation_report.py`** — `flatten()` excludes non-graded case rows from `overall_score`; rows with no `status` key default to `graded` so legacy reports load identically.

Raw per-case `scores` / `test_passes` lists are left untouched — only the *aggregate* changes, so anything reading individual rows is unaffected.

## Behavior change (illustration)

A correct agent, one evaluator grading it `1.0`, and one evaluator that raises (simulating a judge context-window overflow):

| evaluator | score | status |
|---|---|---|
| GoodEvaluator | 1.0 | `graded` |
| OverflowEvaluator | 0.0 | `could_not_evaluate` |

- **Before:** `overall_score = 0.500` (naive average punishes the correct agent).
- **After:** `overall_score = 1.000` (reflects only what was gradable).

## Tests

18 tests across the three layers:
- Aggregator (`_default_aggregator`): graded-only average, `could_not_evaluate`/`informational` exclusion, all-skipped non-failure, empty-list unchanged.
- Report (`flatten()`): non-graded row excluded, missing-status defaults to graded, all-non-graded → 0.
- `EvaluationOutput.status` default + settable.
- Experiment layer: error-isolation path tags `could_not_evaluate`; mixed-status run excludes the non-graded rows from `overall_score`.

Full suite: **1753 passed** (baseline + net-new tests, no regressions). `ruff check` / `ruff format --check` clean.

Closes #346

### DIFF
--- a/src/strands_evals/evaluators/evaluator.py
+++ b/src/strands_evals/evaluators/evaluator.py
@@ -6,7 +6,7 @@ from strands.models.model import Model
 from typing_extensions import Any, Generic, TypeGuard
 
 from ..extractors import TraceExtractor
-from ..types.evaluation import EvaluationData, EvaluationOutput, InputT, OutputT
+from ..types.evaluation import GRADED, EvaluationData, EvaluationOutput, InputT, OutputT
 from ..types.trace import (
     AssistantMessage,
     Context,
@@ -79,13 +79,24 @@ class Evaluator(Generic[InputT, OutputT]):
 
     @staticmethod
     def _default_aggregator(outputs: list[EvaluationOutput]) -> tuple[float, bool, str]:
-        # Handle empty outputs list to avoid division by zero
-        if not outputs:
-            return (0.0, False, "No evaluation outputs produced")
+        # Only "graded" outputs contribute to the score/pass aggregates. Outputs
+        # marked "could_not_evaluate" or "informational" are excluded so a
+        # non-gradable result (e.g. a harness overflow, or an evaluator whose
+        # preconditions weren't met) doesn't silently drag the average up or down.
+        graded = [o for o in outputs if o.status == GRADED]
 
-        avg_score = sum(o.score for o in outputs) / len(outputs)
-        all_pass = all(o.test_pass for o in outputs)
-        combined_reason = " | ".join(o.reason for o in outputs if o.reason)
+        # No gradable outputs (empty list, or every output was non-graded): there
+        # is nothing to score. Report it as a non-failure so a fully-skipped
+        # evaluator doesn't read as a quality failure.
+        if not graded:
+            if not outputs:
+                return (0.0, False, "No evaluation outputs produced")
+            combined_reason = " | ".join(o.reason for o in outputs if o.reason)
+            return (0.0, False, combined_reason or "No gradable evaluation outputs produced")
+
+        avg_score = sum(o.score for o in graded) / len(graded)
+        all_pass = all(o.test_pass for o in graded)
+        combined_reason = " | ".join(o.reason for o in graded if o.reason)
         return avg_score, all_pass, combined_reason
 
     def evaluate(self, evaluation_case: EvaluationData[InputT, OutputT]) -> list[EvaluationOutput]:
--- a/src/strands_evals/experiment.py
+++ b/src/strands_evals/experiment.py
@@ -45,7 +45,7 @@ from .evaluators.trajectory_evaluator import TrajectoryEvaluator
 from .telemetry import get_tracer, serialize
 from .telemetry._cloudwatch_logger import _send_to_cloudwatch
 from .types.detector import DiagnosisConfig
-from .types.evaluation import EvaluationData, InputT, OutputT
+from .types.evaluation import COULD_NOT_EVALUATE, GRADED, EvaluationData, EvaluationOutput, InputT, OutputT
 from .types.evaluation_report import EvaluationReport
 from .types.trace import Session
 from .utils import is_throttling_error
@@ -81,6 +81,19 @@ def _get_label_from_score(evaluator: Evaluator, score: float) -> str:
 
     # Otherwise, return YES/NO based on score
     return "YES" if score >= 0.5 else "NO"
+
+
+def _aggregate_status(outputs: list[EvaluationOutput]) -> str:
+    """Roll a list of per-output statuses up to a single evaluator-level status.
+
+    The evaluator result is "graded" if at least one output was graded; if every
+    output was non-graded, the whole result is non-gradable and is reported as
+    "could_not_evaluate" so downstream aggregates exclude it. Mirrors the
+    graded-only filter in ``Evaluator._default_aggregator``.
+    """
+    if any(o.status == GRADED for o in outputs):
+        return GRADED
+    return COULD_NOT_EVALUATE
 
 
 class Experiment(Generic[InputT, OutputT]):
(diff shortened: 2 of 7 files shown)


Unshure:
Thanks for opening this pull request. I am going to close this in favor of the pull request that was opened by the user who opened this issue initially: https://github.com/strands-agents/evals/pull/359

PR #359, opened 2026-08-10
feat: add status field to EvaluationOutput

max-rattray-aws:
## Summary

Adds an optional `status` field to `EvaluationOutput` that lets evaluators signal when a result should not count toward aggregate scores.

Three values are supported:
- `"graded"` (default) - real verdict, included in all aggregations
- `"could_not_evaluate"` - evaluator tried but could not grade (preconditions not met, harness failure, missing data)
- `"informational"` - surfaces content for human review, not pass/fail

The `_default_aggregator`, `Experiment` report builder, and `EvaluationReport.flatten()` all filter to graded outputs before computing `overall_score`. The field defaults to `"graded"`, making this fully backward compatible.

## Usage

```python
from strands_evals.types import EvaluationOutput

# Evaluator that only applies when tool errors exist
def evaluate(self, evaluation_case):
    if not self._has_errors(evaluation_case):
        return [EvaluationOutput(
            score=0.0,
            test_pass=False,
            reason="No tool errors occurred in trajectory",
            status="could_not_evaluate",
        )]
    # ... actual grading logic
```

## What's tested

- `EvaluationOutput.status` field: default value, serialization roundtrip, backward compat
- `_default_aggregator`: filters non-graded outputs, handles all-CNE, mixed statuses, empty list
- `EvaluationReport.flatten()`: excludes non-graded from `overall_score`, preserves statuses
- File roundtrip: statuses survive `to_file`/`from_file`
- End-to-end: evaluators returning `could_not_evaluate`, `informational`, mixed outputs
- Custom aggregator override still works

All 190 relevant tests pass (24 new + 166 existing). Ruff formatting and linting clean.

Related to #346

### DIFF
--- a/src/strands_evals/evaluators/evaluator.py
+++ b/src/strands_evals/evaluators/evaluator.py
@@ -79,13 +79,16 @@ class Evaluator(Generic[InputT, OutputT]):
 
     @staticmethod
     def _default_aggregator(outputs: list[EvaluationOutput]) -> tuple[float, bool, str]:
-        # Handle empty outputs list to avoid division by zero
-        if not outputs:
-            return (0.0, False, "No evaluation outputs produced")
+        # Filter to only graded outputs for aggregation
+        graded = [o for o in outputs if o.status == "graded"]
 
-        avg_score = sum(o.score for o in outputs) / len(outputs)
-        all_pass = all(o.test_pass for o in outputs)
-        combined_reason = " | ".join(o.reason for o in outputs if o.reason)
+        # Handle empty graded list to avoid division by zero
+        if not graded:
+            return (0.0, False, "No gradable evaluation outputs produced")
+
+        avg_score = sum(o.score for o in graded) / len(graded)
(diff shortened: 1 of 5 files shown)


pdebjyot:
Nice — this matches the design in #346 closely. I'd independently put up #357 for the same issue a few hours earlier; happy to close mine as superseded by this one since it's already through review. Cross-linking for provenance.

One thing worth a follow-up (applies to this implementation as merged, not a blocker): the graded-only filter is applied to overall_score, but the raw scores / test_passes lists still carry the 0.0 / False placeholders for could_not_evaluate rows. Any consumer that computes a pass-rate directly off test_passes — rather than reading overall_score — will still count skipped/overflowed cases as hard failures. Might be worth either a docstring note steering consumers to overall_score/statuses, or a small pass_rate helper that applies the same graded-only filter, so the two aggregation paths can't drift.

Good to see the _worker task-crash path tagged too, not just the _run_evaluator catches — that's the case that most often bites in practice.

PR #365, opened 2026-08-11
feat(mappers): add OpenAI Agents support to GenAI session mapper

liramon2:
## Description
Add support for OpenAI Agent SDK + OpenLLMetry Traceloop instrumentation to `GenericGenAISessionMapper`. This combination supports GenAI conventions with some deviations. The PR handles both Cloudwatch ADOT and in-memory exporter formats. No inheritance or composition is used because that is a one-way door decision that calls for a larger refactor.

This PR also fixes parent span ids in `GenericGenAISessionMapper` so that they point to converted spans. This preserves the agent-tool scopes when converting from OpenAI Agent traces (or other GenAI traces) to evaluator inputs.

## Related Issues

https://github.com/strands-agents/evals/issues/319

## Type of Change

New feature

## Testing

## Checklist

----


liramon2:
@strandly-the-agent Review this PR. Consolidate findings into 1 comment.

strandly-the-agent:
**TL;DR** — draft, so comments rather than approve. No 🔴 blockers: the mapper does the right thing on both shipped fixtures and I verified the per-agent tool scoping end to end. Five 🟡s worth addressing before this leaves draft, the top two being that **the PR's core new behaviour isn't pinned by any test** (tool lists can be swapped with the suite green) and that it **quietly changes a shared base class** other mappers depend on. Nice work overall — the per-agent `available_tools` scoping is a genuine improvement over back-filling everything onto one agent.

✅ Reviewed at `10abb57` · `pytest tests/strands_evals/mappers/ -q` → **373 passed** (4.57s) · misattribution repro confirmed · candidate fix verified (373 passed, both fixtures byte-identical) · ⚠️ the adversarial/repro pass **timed out and was not re-run**, so the deliberate fuzzing angle is a gap in this review; the API pass timed out once and was re-run successfully.

---

### 🟡 1. The per-agent tool attribution — the whole point of the `_convert_trace` override — isn't tested

`openai_agents_otel_session_mapper.py:69-75`, tests at `test_openai_agents_otel_session_mapper.py:59-161`.

All 14 tests are existence checks (`len(...) > 0`, `any(...)`, "every X has a truthy Y"). Mutation testing: I deleted the per-agent scoping so every tool lands on every agent → **14/14 still pass**. I then *swapped* ownership so `coordinator → ['multiply_numbers','divide_numbers']` and `math_specialist → ['ask_math_specialist']` — exactly backwards — → **14/14 still pass**. The correct values (which I confirmed by running the mapper) are:

```
coordinator      (91578cd4f1232ef0) -> ['ask_math_specialist']
math_specialist  (8c1028e91e04dfa2) -> ['multiply_numbers', 'divide_numbers']
```

Worth asserting those literal values, plus the exact agent-span count and names, instead of `any(...)`.

### 🟡 2. The `bridge_parent_gaps` call added to the shared base class is a real fix for *other* consumers — shipped undocumented and untested

`generic_gen_ai_session_mapper.py:119-124` (raw parent map) and `:160` (the call). To be precise about what changed: `bridge_parent_gaps` itself is pre-existing at `utils.py:282`; `utils.py` only gained an import and a routing branch here. The new *call site* is the behaviour change, and it's load-bearing for this mapper — the base class classifies on `gen_ai.operation.name` (`:128`) and converts only `chat`/`execute_tool`/`invoke_agent`, so the 7 live / 4 ADOT spans whose op-name is the literal `"unknown"` get dropped, and those are the **parents** of the chat/tool spans.

Two things follow. It's a genuine bug fix for plain `GenericGenAISessionMapper` users too (without it, a tool nested under an unconverted span is misattributed to the wrong root agent) — so base is the right home, not the subclass. But it changes `parent_span_id` and downstream tool ownership for every consumer of `detect_otel_mapper`'s dict-span fallback, and **deleting the call leaves all 373 mapper tests passing**. It deserves its own test:

```suggestion
class TestBridgeParentGapsIntegration:
    """Regression test for bridge_parent_gaps being wired into _convert_trace."""

    def setup_method(self):
        self.mapper = GenericGenAISessionMapper()

    def test_tool_span_reparented_across_unconverted_intermediate(self):
        spans = [
            make_span(trace_id="t1", span_id="agent-1", operation_name="invoke_agent"),
            make_span(trace_id="t1", span_id="wrapper-1", parent_span_id="agent-1", operation_name="unknown"),
            make_span(
                trace_id="t1", span_id="tool-1", parent_span_id="wrapper-1", operation_name="execute_tool",
                attributes={"gen_ai.tool.name": "calc", "gen_ai.tool.input": "{}", "gen_ai.tool.output": "4"},
            ),
        ]
        session = self.mapper.map_to_session(spans, "sess-1")
        assert len(session.traces[0].spans) == 2  # wrapper-1 dropped
        tool = next(s for s in session.traces[0].spans if isinstance(s, ToolExecutionSpan))
        assert tool.span_info.parent_span_id == "agent-1"  # bridged, not "wrapper-1"
```

### 🟡 3. A nested nameless `invoke_agent` span makes its tool bubble up to the wrong agent

`openai_agents_otel_session_mapper.py:86-89`. The skip is the only thing in the repo that can drop a real agent span (the base `_convert_agent_invocation_span` never returns `None`). Drop one and it's no longer there to stop `Trace.model_post_init`'s subtree walk at the agent boundary (`types/trace.py:195-196` checks `isinstance`, so a *missing* boundary is invisible), `bridge_parent_gaps` re-parents the tool past it, and the back-fill then credits the wrong agent. Reproduced:

```
root(coordinator) -> gap -> sub(no agent.name, no input.messages) -> gap -> tool(divide_numbers)
=> AGENT root tools=['divide_numbers']   TOOL divide_numbers owner=root   sub: absent
```

**Reachability, honestly:** this needs a *nested* `invoke_agent` span carrying neither attribute. Traceloop demonstrably does emit nameless `invoke_agent` spans — the root `Agent workflow` in both fixtures is one — but in every capture here only the root is nameless, so I can't name a real trace that nests one. That's why this is 🟡 and not a blocker.

One caution: the intuitive fix ("also require the span be parentless") is **wrong** — I checked, and in the ADOT fixture `Agent workflow`'s parent `6b49d758067b2298` *is* in the trace (the starlette `POST /invocations` span), so that condition would stop skipping the wrapper there. Gating on whether any *ancestor* is itself an `invoke_agent` span does hold — with it the repro is fixed, both fixtures produce byte-identical output, and the suite stays at 373 passed:

```suggestion
        if not attrs.get("gen_ai.agent.name") and not attrs.get("gen_ai.input.messages"):
            # Only a ROOT wrapper is safe to drop. A nested nameless agent must be kept:
            # it is the boundary that stops tool spans being credited to an ancestor agent.
            cur = self._parent_of.get(span.get("span_id"))
            seen: set[str] = set()
            while cur and cur not in seen:
                seen.add(cur)
                if cur in self._agent_op_ids:
                    return super()._convert_agent_invocation_span(span, span_info)
                cur = self._parent_of.get(cur)
            return None
```

(`_agent_op_ids` / `_parent_of` built from the trace's raw spans in `_convert_trace`, which already receives them.)

### 🟡 4. The module docstring documents an export format that has no code path

`openai_agents_otel_session_mapper.py:6-8` claims format 1 is *"ADOT/CloudWatch: Messages in `span_events[].body.input/output.messages`"*. Every span in **both** fixtures has `span_events: []` and no `body` key; messages are in `gen_ai.input.messages` attributes in both (7 live / 5 ADOT). The inherited base class has no `body` handling at all — its `span_events` paths (`generic_gen_ai_session_mapper.py:180`, `:264`, `:346`) key on `event_name`. The wording looks copy-pasted from `langchain_otel_session_mapper.py:7`, a standalone mapper that really does implement that path.

```suggestion
Supports two attribute-based export shapes — both carry messages in
gen_ai.input.messages / gen_ai.output.messages span attributes (this instrumentor
populates no span_events):
1. Live SDK export: bare openai_agents-scoped spans.
2. ADOT/CloudWatch export: the same spans plus AWS-specific `aws.*` attributes,
   with httpx/starlette spans mixed into the same trace.
```

### 🟡 5. `agent_handoff` spans are dropped silently, and in the ADOT fixture that loses the only link between the two agents

The base class classifies only `chat`/`execute_tool`/`invoke_agent` (`generic_gen_ai_session_mapper.py:140-151`), so the fixture's `agent_handoff OpenAI Agent` span (`span_id=42963e788a6291be`, carrying `gen_ai.handoff.from_agent=coordinator` / `to_agent=math_specialist`) vanishes. Both ADOT `AgentInvocationSpan`s then end up `parent_span_id=None`, so nothing in the output says the two agents are related. `BaseSpan.metadata` already exists if you want somewhere cheap to stash it — or a follow-up issue is fine, just not silence.

---

### Questions (all non-blocking)

1. Would it be cleaner to split the base-class `bridge_parent_gaps` change into its own commit/PR with its own test? It's a fix for consumers who have nothing to do with OpenAI Agents, and right now it's invisible in a PR titled "add OpenAI Agents mapper".
2. Is the user-facing vocabulary settled? The branch says `openllmetry`, the commit says `traceloop`, the class says `openai_agents_otel`, the docstring says "Traceloop/OpenLLMetry". The exported class name is the one that's hard to change later — and "OpenAI" alone may read as "OpenAI models generally" or collide with OpenInference's own `openai` instrumentation.
3. Is dropping `agent_handoff` a deliberate scope call for now?

<details>
<summary>Per-pass breakdown (7 passes routed; 1 did not complete)</summary>

| Pass | Verdict | Notes |
|---|---|---|
| Triage | accept | routine mapper addition implementing #319 |
| Context build | — | #319 explicitly specifies subclassing `GenericGenAISessionMapper`, so the inheritance is sanctioned, not ad-hoc. This is the only mapper in the repo that subclasses another. The `bridge_parent_gaps` usage is idiomatically consistent with all three existing callers. |
| Correctness/safety | issues | findings 3, 5, 2 |
| Test quality | issues | finding 1 — 4 mutants, 3 survive |
| Docs accuracy | docs-issues | finding 4, plus appendix items |
| LLM-context (LLMINFO) | model-context-gaps | all pre-existing/upstream — see appendix |
| API / DevX | minor | naming + extension-point docs; no `needs-api-review` label exists in this repo, so no label gate applies |
| Adversarial / repro | **did not complete** | timed out and was not re-run — disclosed as a gap |

Aggregation was done inline by the orchestrator rather than by a separate pass.

</details>

<details>
<summary>Appendix — non-blocking (8)</summary>

**⚪ Nits in this diff**

- `openai_agents_otel_session_mapper.py:80-85` — the docstring says it "Skips the root 'Agent workflow' span", but the condition never inspects the span name. That coincidence holds in both fixtures; describing the *mechanism* would stop a future maintainer being surprised when a differently-named span vanishes.
- `utils.py:159` — the fallback `known_scopes` set wasn't extended with `SCOPE_OPENAI_AGENTS`. Harmless today only because the first loop returns as soon as any span matches; it's safe by ordering, not by the set being right.
- `openai_agents_otel_session_mapper.py` scope filter also admits scope-less (`""`) spans and returns an empty `Session` with no log line on a producer mismatch. Not a new gap — `adk_otel_session_mapper.py:85` does exactly the same, so there's no convention here being skipped.
- The PR checklist has "I have updated the documentation accordingly" ticked, but no docs file was touched. Worth unticking. For what it's worth this is a repo-wide gap rather than yours: neither #309 nor #340 updated `SKILL.md` either, so its mapper list at `SKILL.md:143-144` is already missing `GenericGenAISessionMapper`.
- `GenericGenAISessionMapper` doesn't document `_convert_trace` (`:115`) or `_convert_agent_invocation_span` (`:325`) as override points. Now that a subclass overrides both, a one-line note on each would help.

**Pre-existing / upstream — explicitly not this PR's problem, listed so they're not lost**

- `trace_extractor.py:107-115,149` — tool-level `session_history` takes one root agent's `user_prompt` and reuses it for every tool span. Live fixture: `math_specialist`'s own input is `"42 * 17 / 3"` but its tool calls render the coordinator's `"What is 42 multiplied by 17, then divided by 3?"`. Benign here (paraphrase), and untouched by this PR — but these are the first fixtures in the repo with genuinely distinct multi-agent prompts, so they're what exposes it.
- Upstream instrumentation bug: ADOT's coordinator turn has a `type:"text"` part whose content is a Python `repr` — `ResponseFunctionToolCall(arguments='{}', call_id='call_51Gc…', name='transfer_to_math_specialist', …)`. I confirmed that repr is already in the raw fixture attribute, so no mapper creates it, but it reaches correctness/coherence judges verbatim with internal ids. It's also why the ADOT coordinator's `available_tools` is `[]` despite it having called two handoff tools — they were emitted as text, not tool calls.
- The `available_tools` back-fill is call-tautological: a tool is only listed if it was called, so the tool under evaluation always appears in its own "available tools" list and a foregone alternative never can. Same pattern in `cloudwatch_session_mapper.py:209`, `opensearch_session_mapper.py:93`, `strands_in_memory_session_mapper.py:415-419`. I checked that neither fixture carries a `gen_ai.agent.tools`-style attribute, so no schema is being left on the table — the source format genuinely lacks it.

Happy to file follow-up issues for any of these three if useful — I didn't want to open tickets unasked.

</details>

<details>
<summary>What I ran</summary>

```
branch pr365 @ 10abb5736a24303410b2e18701d5cc00bb62a407 (base main, merge-base e82933c)

pytest tests/strands_evals/mappers/ -q                      -> 373 passed, 4.57s
pytest .../test_openai_agents_otel_session_mapper.py -q      -> 14 passed
mapper -> TraceExtractor(TOOL_LEVEL) on the live fixture     -> per-agent available_tools confirmed
mutants M1/M4 (tool scoping deleted / swapped)               -> 14/14 pass  (gap)
mutant  M2 (bridge_parent_gaps call removed)                 -> 373 pass    (gap)
mutant  M3 (wrapper skip removed)                            -> 1 failure   (covered)
candidate fix for finding 3                                  -> 373 passed, fixtures unchanged
```

One correction to my own working notes, in case it saves someone the detour: an early probe of mine appeared to show empty trace-level tool lists. `TraceLevelInput` has no `available_tools` field at all (`['span_info', 'agent_response', 'session_history']`), so that was my own `getattr` default, not a finding. Dropped it.

*Automated review — solid work for a human to approve, not a gate. Worth a maintainer's eye on findings 1 and 2 in particular.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate findings into 1 comment. Accepting the risk of finding 3 because it is difficult to reach.

strandly-the-agent:
**TL;DR** — re-reviewed at `5b1142b`. Findings 1, 2 and 4 are genuinely fixed, and I confirmed it the hard way: the three mutants that previously survived the suite are now all killed. Finding 3's risk is yours to accept — noted, not re-raised. Only finding 5 (`agent_handoff` dropped silently) is still open, and it needs a yes/no from you rather than more work from me. Still a draft so this stays comments, but from my side there's nothing blocking left.

✅ `5b1142b` · mapper suite **378 passed** (was 373) · mappers+extractors+types **510 passed** · `ruff check` + `ruff format --check` clean · the only source change in this delta is one docstring line, so zero behavioural risk from the commit itself.

### Previous findings — status

| | Finding | Status |
|---|---|---|
| 🟡1 | per-agent tool attribution untested | **fixed** — `TestPerAgentToolAttribution` |
| 🟡2 | base-class `bridge_parent_gaps` untested | **fixed** — `TestUnconvertedSpanReparenting` |
| 🟡3 | nested nameless `invoke_agent` span | **risk accepted by you** — not re-raised |
| 🟡4 | docstring documented a non-existent format | **fixed** (one cosmetic leftover below) |
| 🟡5 | `agent_handoff` spans dropped silently | **still open** |

### The mutation evidence, re-run against your new tests

This is the part I'd want to see as the author, because passing tests alone don't prove the tests *bite*:

| Mutant | Before | Now |
|---|---|---|
| M1 — drop per-agent scoping, back-fill every tool onto every agent | 14/14 passed ❌ | **3 failures** ✅ |
| M4 — swap tool ownership to the wrong agent | 14/14 passed ❌ | **2 failures** ✅ |
| M2 — delete the `bridge_parent_gaps` call from the shared base class | 373 passed ❌ | **3 failures** ✅ |

One thing worth knowing about M2: it's caught by your new base-class test *and* by both `test_coordinator_tools` / `test_math_specialist_tools`. So the live fixture's per-agent attribution genuinely depends on the bridging — the two changes are coupled, and that coupling is now pinned from both ends. That's better coverage than I asked for.

I also checked the back-fill is idempotent, since your new tests assert exact tool lists and `Trace.model_post_init` mutates spans in place — mapping the same spans three times with the same mapper instance gives identical lists, no double-append.

### Still open

**🟡 5 — `agent_handoff` spans are still dropped silently** (`generic_gen_ai_session_mapper.py:140-151`). In the ADOT fixture, span `42963e788a6291be` carries `gen_ai.handoff.from_agent=coordinator` / `to_agent=math_specialist` and is the only thing linking the two agents; both `AgentInvocationSpan`s come out `parent_span_id=None`, so the relationship is unrecoverable downstream. Entirely reasonable to defer — I'd just rather it be a deliberate "not now, tracked in #NNN" than silence. Happy to file the issue if you want it.

<details>
<summary>Cosmetic leftover on finding 4, and two carried-over ⚪ nits (all optional)</summary>

**Finding 4's fix is correct but the list is now self-cancelling** — `openai_agents_otel_session_mapper.py:6-8`:

```
Supports two trace formats:
1. ADOT/CloudWatch: Messages in gen_ai.* attributes
2. Live instrumentation: Messages in gen_ai.* attributes
```

The misleading `span_events[].body` claim is gone, which was the actual problem. But both entries now say the same thing, so the list tells a reader nothing about how the two differ. The real difference is the extra `aws.*` attributes and the `httpx`/`starlette` spans mixed into the same trace:

```suggestion
Both supported export shapes carry messages in gen_ai.input.messages /
gen_ai.output.messages attributes (this instrumentor populates no span_events):
1. Live SDK export: bare openai_agents-scoped spans.
2. ADOT/CloudWatch export: the same spans plus AWS-specific `aws.*` attributes,
   with httpx/starlette spans mixed into the same trace.
```

**⚪** `test_tools_not_shared_across_agents` compares only `tool_sets[0] & tool_sets[1]`, so it would quietly ignore a third agent if one ever appeared (and `IndexError` rather than assert if there were fewer than two — though `test_live_session_has_exactly_two_agent_spans` covers that). `assert not set.intersection(*tool_sets)` over all agents would generalise for free.

**⚪** Per-agent attribution is asserted on the live fixture only. The ADOT coordinator's `available_tools` is `[]` (because its handoff calls were emitted as text — the upstream `repr` issue from my last comment), and nothing pins that, so if finding 5 ever gets addressed no test will notice the change.

Still open from the previous appendix, unchanged and all optional: the `_convert_agent_invocation_span` docstring at `:80-85` still describes name-matching it doesn't do; `utils.py:159`'s `known_scopes` set still omits `SCOPE_OPENAI_AGENTS` (safe only by loop ordering); and the PR checklist's "I have updated the documentation accordingly" box is still ticked with no docs file touched.

</details>

<details>
<summary>Review shape and what I ran</summary>

This was a **self-run follow-up**, not the full multi-pass pipeline: the delta since `10abb57` is 71 lines touching two test files plus one docstring line, all of it addressing findings I filed, so a fresh fan-out would mostly have rediscovered its own earlier output. I read the new commit, verified each fix against the diff rather than the commit message, re-ran the original mutants, and attacked the fix itself (idempotency, lint, wider suite). No adversarial/repro pass ran this round — same gap I disclosed last time, and with no behavioural source change in this delta it's a smaller gap than it was.

```
head 5b1142bd6de489b2034bdd641f0486cb7773c5ca (delta from 10abb57: +71/-1, 3 files)

pytest tests/strands_evals/mappers/ -q                     -> 378 passed, 3.72s
pytest mappers/ extractors/ types/ -q                      -> 510 passed
pytest TestPerAgentToolAttribution TestUnconvertedSpanReparenting -> 5 passed
ruff check <3 changed files>                               -> All checks passed
ruff format --check <3 changed files>                       -> 3 files already formatted
mutants M1 / M4 / M2                                       -> 3 / 2 / 3 failures (all now caught)
repeated map_to_session x3, same mapper instance            -> identical tool lists (idempotent)
```

Tree restored to PR head after each mutant; nothing left modified.

*Automated review — solid work for a human to approve, not a gate.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.

strandly-the-agent:
**TL;DR** — follow-up review of just the `5b1142b..9fb48f5` delta. **All five earlier findings are now resolved**, including finding 5 which this commit fixes properly. The new `_apply_handoff_reparenting` is the right idea and works on the real fixture; it has three edge cases around agent *names* being used as identity. None of them blocks — I went looking for a hang or corrupted tool ownership and found neither. Still a draft, so comments rather than approve.

✅ `9fb48f5` · mapper suite **381 passed** (was 378) · `ruff check` + `ruff format --check` clean · finding 5 verified fixed: ADOT `math_specialist` (`c2828d53371818f0`) now parents under `coordinator` (`1afcb5ca9fe89dda`) · cycle-guard patch below verified (381 still passing, fixtures unchanged).

### Previous findings — all closed

| | Finding | Status |
|---|---|---|
| 🟡1 | per-agent tool attribution untested | fixed at `5b1142b` |
| 🟡2 | base-class `bridge_parent_gaps` untested | fixed at `5b1142b` |
| 🟡3 | nested nameless `invoke_agent` span | **risk accepted by you** — not re-raised |
| 🟡4 | docstring documented a non-existent format | fixed, including the cosmetic leftover — the two formats now actually differ |
| 🟡5 | `agent_handoff` dropped silently | **fixed by this commit** — verified above |

You also took the `set.intersection(*tool_sets)` nit, and added the ADOT per-agent assertions I'd flagged as missing. Nothing from my earlier comments is outstanding.

---

All three new findings are in `_apply_handoff_reparenting` (`openai_agents_otel_session_mapper.py:72-101`), and they share one root cause: **`gen_ai.agent.name` is treated as a unique key for an agent span.**

### 🟡 6. Duplicate agent names — only the last span with a given name is reparented

`openai_agents_otel_session_mapper.py:89` — `agent_spans_by_name[name] = agent_spans[span_id]` is a plain dict, so when the same sub-agent is invoked twice in one trace (ordinary in a multi-turn conversation) earlier invocations are overwritten and never reparented. Repro — one `coordinator → math_specialist` handoff, two `math_specialist` invocations:

```
AGENT coord  parent=None
AGENT ms1    parent=None     <- stranded, never reparented
AGENT ms2    parent=coord
```

`ms1` keeps whatever parent it had, so the tree is half-built and `_find_root_agent_span` sees two roots where there should be one. This is the one I'd most want addressed, and it needs a small shape change rather than a one-liner: key by name to a *list* and resolve each handoff edge to a specific invocation (nearest following `start_time`, or nearest common ancestor) instead of "the last one seen". Happy to prototype it if useful.

### 🟡 7. Handoff edges can build a cyclic or self-referential parent tree

`openai_agents_otel_session_mapper.py:97-101` assigns unconditionally, with no check that the edge is acyclic. Mutual handoff — which is a normal OpenAI Agents pattern, a specialist handing control back to a triage/coordinator agent — produces `A.parent=B` **and** `B.parent=A`. `from_agent == to_agent` produces `A.parent=A`.

**Being straight about severity:** I tried to make this hurt and couldn't. `TraceExtractor` at both TRACE_LEVEL and TOOL_LEVEL extracts fine, re-constructing a `Trace` from the cyclic spans completes, and tool ownership stays correct — every parent-walker I could find (`bridge_parent_gaps`, `Trace.model_post_init`) has a `visited`/`seen` guard. So this is a data-integrity wart, not a blocker. The real consequence is that with a cycle *no* agent is parentless, so `_find_root_agent_span` (`types/trace.py:135-149`) silently falls through to its earliest-`start_time` branch — "which agent is the root" stops being derived from structure. Guarding is cheap, and I verified this keeps both fixtures identical at 381 passed:

```suggestion
        agents_by_id = {
            a.span_info.span_id: a for a in agent_spans_by_name.values() if a.span_info.span_id
        }

        def _would_cycle(child_id: str, new_parent_id: str) -> bool:
            cur: str | None = new_parent_id
            seen: set[str] = set()
            while cur and cur not in seen:
                if cur == child_id:
                    return True
                seen.add(cur)
                ancestor = agents_by_id.get(cur)
                cur = ancestor.span_info.parent_span_id if ancestor else None
            return False

        # Re-parent sub-agents to parent agents
        for from_name, to_name in handoffs:
            from_span = agent_spans_by_name.get(from_name)
            to_span = agent_spans_by_name.get(to_name)
            if not (from_span and to_span and from_span.span_info.span_id and to_span.span_info.span_id):
                continue
            if from_span is to_span:
                logger.debug("agent=<%s> | skipping self-handoff", to_name)
                continue
            if _would_cycle(to_span.span_info.span_id, from_span.span_info.span_id):
                logger.debug("from=<%s> to=<%s> | skipping cycle-inducing handoff", from_name, to_name)
                continue
            to_span.span_info.parent_span_id = from_span.span_info.span_id
```

With that, `A↔B` resolves to `A.parent=None, B.parent=A` and self-handoff leaves `A.parent=None`.

### 🟡 8. Two handoffs to the same target make the tree depend on raw span order

Same lines — last-write-wins. Given `alpha → gamma` and `beta → gamma` in one trace, `gamma`'s parent is whichever handoff span appears later in the input list. Identical logical input, two different trees:

```
span order [alpha->gamma, beta->gamma]  =>  C.parent=B
span order [beta->gamma, alpha->gamma]  =>  C.parent=A
```

Since span order isn't guaranteed stable across exporters, the same trace can evaluate differently on two runs. Worth at least a documented tie-break (earliest handoff wins, say) so it's a decision rather than an accident.

<details>
<summary>Appendix — non-blocking (3)</summary>

**⚪ The new logic has no edge-case tests.** `test_handoff_reparents_math_specialist_under_coordinator` covers the ADOT happy path only — nothing exercises duplicate names, a mutual handoff, a self-handoff, or a handoff naming an agent whose span was dropped by the wrapper-skip. Findings 6–8 are all invisible to the suite as it stands.

**⚪ Docstrings that explained non-obvious ordering were deleted in this commit.** The class docstring lost its explanation of why the overrides exist, and `_convert_trace`'s docstring (`:52`) lost "After base conversion (which sets `agent_span_id` via `Trace.model_post_init`)". That sentence was load-bearing: reparenting at `:59` runs *after* `super()._convert_trace()` has already constructed the `Trace` and assigned tool ownership, so the back-fill at `:62-68` reads a pre-reparenting `agent_span_id`. Both happen to agree today, but the next person to touch the ordering has lost the note that told them why it matters. `_apply_handoff_reparenting`'s own docstring is also silent on the name-keyed / last-write-wins behaviour behind findings 6 and 8.

**⚪ Consequence of a nit I raised last round.** `set.intersection(*tool_sets)` is the right generalisation, but it raises `TypeError` on zero agents and, with exactly one agent, returns that agent's own set so `assert not ...` fails confusingly. `test_live_session_has_exactly_two_agent_spans` covers the fixture, so this is theoretical — flagging it only because it came from my suggestion.

</details>

<details>
<summary>Review shape, and one thing I got wrong</summary>

**Shape:** a self-run follow-up scoped to the `5b1142b..9fb48f5` delta (+71/-19 across the mapper and its test file). Because this delta carries genuinely new behaviour rather than just tests, I ran an independent adversarial pass over the new logic as well — but it exceeded its time budget and **did not complete**, so it contributed nothing here and the findings above are all from my own repros. Aggregation was done inline. Net: findings 6–8 are single-reviewer results, so a second pair of eyes on the reparenting logic wouldn't be wasted.

**Something I got wrong, in case the diff misleads anyone else:** reading the unified diff for the new test file, it looked like it contained two syntax errors (a stray `)` and an unclosed `next(`). That was a diff-rendering artifact — the file on disk is fine, `python -m py_compile` exits 0, and all 381 tests collect and pass. No syntax problem exists; I'd rather say so than leave a phantom blocker implied.

```
head 9fb48f50e2bd234f16d3c05c7b07e6cc2c8995c3 (delta from 5b1142b: +71/-19, 2 files)

pytest tests/strands_evals/mappers/ -q                  -> 381 passed, 3.93s
ruff check / ruff format --check (2 changed files)      -> All checks passed / already formatted
py_compile on the new test file                         -> exit 0 (no syntax error)
ADOT fixture through the mapper                         -> math_specialist parent=coordinator (finding 5 fixed)
repro: duplicate agent name                             -> ms1 parent=None, ms2 parent=coord
repro: mutual handoff A<->B                             -> A.parent=B, B.parent=A (cycle)
repro: self-handoff                                     -> A.parent=A
repro: two handoffs -> same target, order swapped        -> C.parent=B vs C.parent=A
hang hunt: TRACE_LEVEL, TOOL_LEVEL, Trace re-construction under a cycle -> no hang, ownership intact
cycle-guard patch above                                 -> 381 passed, ruff clean, fixtures unchanged
```

Tree restored to PR head after every experiment; nothing left modified.

*Automated review — solid work for a human to approve, not a gate.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.  The 3 findings found in the last round are theoretical, not realistic.

strandly-the-agent:
**No new findings.** Reviewed the `9fb48f5..ebc5c5a` delta — it's a docstring only, functionally inert, and **accurate**. Findings 6–8 withdrawn per your call; nothing outstanding from any round. Still a draft so I won't mark approve, but there's nothing blocking from my side.

✅ `ebc5c5a` · mapper suite **381 passed** · `ruff check` + `ruff format --check` clean · diff contains no functional lines, docstring only.

**Your call on 6–8 is a fair one, and my own evidence supports it.** All three needed a multi-agent topology that appears in neither capture, and when I went looking for consequences I found none worth the name: no hang anywhere, tool ownership stayed correct, and the only reachable effect was `_find_root_agent_span` falling through to its earliest-`start_time` tiebreak. I tiered them 🟡 out of caution; ⚪ would have been the better call.

One thing worth recording, since it's the useful outcome here: the new docstring on `_apply_handoff_reparenting` (`openai_agents_otel_session_mapper.py:77-81`) — *"Uses agent name as span identity (last-write-wins for duplicate names or competing handoffs)"* — is exactly right, and I re-ran my repros against it to confirm rather than assume. Competing handoffs `alpha→gamma` then `beta→gamma` give `gamma.parent=beta`; reverse the span order and you get `gamma.parent=alpha`. So the docstring now states the real contract, which turns the behaviour I flagged into a documented design decision. That's a legitimate resolution and, for a limitation you judge unreachable in practice, the cheaper one.

<details>
<summary>What I ran</summary>

```
head ebc5c5ad4038298bed6b844fa436fc390d077e5e (delta from 9fb48f5: docstring only, +4)

pytest tests/strands_evals/mappers/ -q            -> 381 passed, 3.44s
ruff check / ruff format --check                  -> All checks passed / already formatted
diff filtered to non-docstring lines              -> none (functionally inert)
repro: competing handoffs, span order swapped      -> gamma.parent=beta vs gamma.parent=alpha
                                                      (confirms the documented last-write-wins)
```

Self-run follow-up on a 4-line docs delta — no fan-out, which would have been overkill here. Full history for anyone joining: findings 1, 2, 4 fixed at `5b1142b`; finding 5 fixed at `9fb48f5`; finding 3 risk-accepted; findings 6–8 withdrawn here.

*Automated review — solid work for a human to approve, not a gate.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.

strandly-the-agent:
**TL;DR** — this round is a real re-architecture, not a tweak: `145008e8` deletes the `OpenAIAgentsOtelSessionMapper` subclass and folds its logic into `GenericGenAISessionMapper`, and the PR is **no longer a draft**. I reviewed the whole `ebc5c5a..145008e8` delta fresh. It's in good shape — the consolidation is clean, routing still works, and I confirmed the OpenAI-specific logic does **not** leak into other generic-genai traces. One 🟡: the genuinely new `_backfill_agent_prompts` method is real, works, but is **not exercised by any test**. Nothing blocking.

✅ `145008e8` · mapper suite **384 passed** · mappers+extractors+types **516 passed** · `ruff check`/`format` clean · both fixtures produce the same agent/tool structure as the pre-refactor head · verified the enrichment is scope-gated (no leak) and the new back-fill both fires correctly *and* isn't tested.

### The consolidation itself — looks right

This is effectively the answer to the design question from my first review ("should this be a subclass at all?"): the subclass is gone, `detect_otel_mapper` drops its routing branch, and `SCOPE_OPENAI_AGENTS` traces now fall through to `GenericGenAISessionMapper` via the unrecognized-scope fallback. I verified that routing still lands both fixtures on `GenericGenAISessionMapper`, and that removing the old `map_to_session` scope-filter didn't change the output — the mixed-in `httpx`/`starlette` spans in the ADOT fixture carry no `gen_ai.operation.name`, so they're ignored during conversion just as before.

The one thing I specifically went after — since this moves vendor-specific logic into the shared fallback mapper — is whether it now contaminates non-OpenAI traces. It doesn't: both entry points are correctly gated. `_enrich_openai_agent_spans` only runs when `any(get_scope_name(span) == SCOPE_OPENAI_AGENTS ...)` (`generic_gen_ai_session_mapper.py:168`), and `_normalize_openai_agents_spans` (`:613`) checks the scope per span. A synthetic generic-genai trace under a different scope came through untouched (a payload-less agent kept its own prompt, no back-fill fired). Good.

### 🟡 The new `_backfill_agent_prompts` isn't tested — the enrichment tests pass without it

`generic_gen_ai_session_mapper.py:692`. This method is new this commit (not just moved): it derives an agent's `user_prompt`/`agent_response` from its child inference spans when the agent span itself has none — for Traceloop ≥0.62, per the `TestOpenAIAgentsPromptEnrichment` docstring, which emits payload-less agent spans.

The gap: **both shipped fixtures are pre-0.62**, so their agent spans already carry `gen_ai.input.messages`/`output.messages`. I instrumented the method — it changes an agent field **0 times** across both fixtures, because the `if not agent.user_prompt` / `if not agent.agent_response` guards never trip. So the three `TestOpenAIAgentsPromptEnrichment` tests are actually exercising the *pre-existing* base-class extraction, not this new code. Confirmed by mutation: I made `_backfill_agent_prompts` a no-op and **all three enrichment tests still pass**.

To be fair to the code, I checked it actually works when it *is* needed — a payload-less agent span with a child chat span does get enriched correctly:

```
payload-less invoke_agent + child chat  ->  user_prompt='real user question', agent_response='real assistant answer'
```

So this is a coverage gap, not a bug: a regression in the 0.62 path would ship green. Worth a fixture-free unit test that builds the payload-less shape directly — paste-ready:

```suggestion
    def test_backfill_from_child_inference_when_agent_span_has_no_payload(self):
        """Traceloop >=0.62 emits payload-less invoke_agent spans; prompt/response
        must be derived from child chat spans."""
        scope = {"name": SCOPE_OPENAI_AGENTS}
        spans = [
            {"trace_id": "t1", "span_id": "ag", "parent_span_id": None, "scope": scope,
             "span_events": [], "start_time": "2026-01-01T00:00:00Z", "end_time": "2026-01-01T00:00:02Z",
             "attributes": {"gen_ai.operation.name": "invoke_agent", "gen_ai.agent.name": "solo"}},
            {"trace_id": "t1", "span_id": "chat", "parent_span_id": "ag", "scope": scope,
             "span_events": [], "start_time": "2026-01-01T00:00:00Z", "end_time": "2026-01-01T00:00:01Z",
             "attributes": {"gen_ai.operation.name": "chat",
                            "gen_ai.input.messages": '[{"role":"user","parts":[{"type":"text","content":"real user question"}]}]',
                            "gen_ai.output.messages": '[{"role":"assistant","parts":[{"type":"text","content":"real assistant answer"}]}]'}},
        ]
        session = GenericGenAISessionMapper().map_to_session(spans, "test")
        agent = next(s for t in session.traces for s in t.spans if isinstance(s, AgentInvocationSpan))
        assert agent.user_prompt == "real user question"
        assert agent.agent_response == "real assistant answer"
```

(The same gap I flagged in round 1 for the tool back-fill, and you closed it there with exact-value asserts — this is the one new behaviour that slipped through the same crack.)

### Questions (non-blocking)

1. `GenericGenAISessionMapper` now carries ~140 lines of OpenAI-Agents-specific handling behind `SCOPE_OPENAI_AGENTS` checks, in a class whose docstring is "any framework that uses OTel GenAI semantic conventions." It's correctly gated so it's harmless to other producers — but is a vendor special-case inside the generic fallback the shape you want long-term, versus (say) a small mixin the fallback composes in? Genuinely a judgment call, and either way I'd want the class docstring to mention it now hosts producer-specific enrichment. No action needed for this PR.

<details>
<summary>Prior findings — full history</summary>

All resolved before this round: findings 1, 2, 4 fixed at `5b1142b`; finding 5 (`agent_handoff` dropped) fixed at `9fb48f5`; finding 3 risk-accepted; findings 6–8 (handoff duplicate-name / cycle / order-dependence) withdrawn as theoretical at your call — not re-raised. The handoff code moved verbatim into `_reparent_by_handoff` (`:661`) this round; since you've judged those cases unreachable in practice I'm treating the documented last-write-wins contract as the final word on them.

</details>

<details>
<summary>Review shape and what I ran</summary>

This delta is a re-architecture of a shared class (+162/-4, a deleted public class, one new method), so I reviewed it as a fresh change rather than a light follow-up — but as a **focused self-review**, not a full independent fan-out: most of the 162 lines are code I've already reviewed across three rounds moved verbatim, and the one genuinely new surface (`_backfill_agent_prompts`) I verified directly below. I did not run a separate independent adversarial pass on the new method this round; my repros are single-reviewer.

```
head 145008e80f9ebfa82766f7cb57d4178ce1f67210 (delta from ebc5c5a: +162/-4 src, subclass deleted, fixtures renamed)

pytest tests/strands_evals/mappers/ -q                         -> 384 passed
pytest mappers/ extractors/ types/ -q                          -> 516 passed
ruff check / format --check (changed files)                    -> clean
detect_otel_mapper(live/adot spans)                            -> GenericGenAISessionMapper (routing survives subclass deletion)
both fixtures end-to-end                                       -> same agent/tool structure as pre-refactor
leak check: generic-genai trace under a non-openai scope        -> no enrichment applied (correctly gated at :168)
_backfill_agent_prompts instrumented on both fixtures           -> fires 0 times (agent spans already have payloads)
mutant: _backfill_agent_prompts -> no-op                        -> TestOpenAIAgentsPromptEnrichment still 3/3 pass (the gap)
fairness: payload-less agent + child chat                       -> correctly enriched (method works when exercised)
```

Tree restored after every experiment; nothing left modified.

*Automated review — solid work for a human to approve, not a gate.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment. A vendor special-case is acceptable for now and the docstring specifies that the scope name is unrecognized by the other mappers.

strandly-the-agent:
**No new findings, and nothing of mine remains open.** Reviewed the `145008e8..6971865` delta — tests only, no source change. The round-5 🟡 (`_backfill_agent_prompts` untested) is **verified closed**, and the vendor special-case question is closed per your call — the module docstring's "scope name unrecognized by any other mappers" wording covers it. From my side this PR is clear; a human maintainer's approval is all that's left.

✅ `6971865` · mapper suite **378 passed** · `ruff check` + `format --check` clean · both closure claims below verified by mutation, not by reading.

### Status

| | Item | Status |
|---|---|---|
| 🟡 round 5 | `_backfill_agent_prompts` untested | **fixed** — `test_backfill_from_child_inference_when_agent_span_has_no_payload` (+ the no-override companion). Re-ran the no-op mutant that previously slipped through: it now **fails** that test. The extra `test_backfill_does_not_override_existing_agent_payload` pins the guard behaviour I'd only described — nice addition. |
| Q1 round 5 | vendor special-case in the generic mapper | **closed per your call** — acceptable for now, and the docstring documents the unrecognized-scope contract. Not re-raised. |
| 3, 6–8 | earlier risk-accepted / withdrawn items | unchanged, not re-raised |

### The one thing I checked beyond the new tests

The commit also consolidates several redundant existence tests (suite went 384 → 378), including removing `test_no_unknown_operation_spans` — which back in round 1 was the *only* test catching the wrapper-skip mutant. So I re-ran that mutant against the new suite rather than assuming: disabling the root-wrapper skip in `_normalize_openai_agents_spans` now fails **3** tests (`test_live_session_has_exactly_two_agent_spans`, `test_handoff_reparents_math_specialist_under_coordinator`, `test_all_agent_spans_have_prompt_and_response`). The consolidation lost no load-bearing coverage — the exact-value tests from round 2 are doing the work the removed checks used to do, and doing it better.

<details>
<summary>What I ran</summary>

```
head 6971865579e5e110809508272713bb652442bf53 (delta from 145008e8: tests only, +86/-59, 1 file)

pytest tests/strands_evals/mappers/ -q                      -> 378 passed, 3.32s
ruff check / ruff format --check (changed file)             -> clean
mutant: _backfill_agent_prompts -> no-op                    -> 1 failed (was 0 in round 5 — gap closed)
mutant: disable root-wrapper skip                           -> 3 failed (coverage survived the test consolidation)
```

Self-run follow-up — the delta only addresses my round-5 finding plus test cleanup, so no fan-out. Tree restored after each mutant; nothing left modified.

Full arc for anyone joining: findings 1/2/4 fixed at `5b1142b` · finding 5 fixed at `9fb48f5` · finding 3 risk-accepted · findings 6–8 withdrawn as theoretical · round-5 🟡 fixed here · vendor special-case accepted with docstring coverage.

*Automated review — solid work for a human to approve, not a gate.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.

strandly-the-agent:
**TL;DR** — the `086d053` source change is a genuine improvement (real tool schemas now reach the judge), **but the suite is red at this head: 5 failed**. The live fixture was regenerated with the newer Traceloop format — new span ids, new span names, and a third agent — and the five tests that hardcode the old fixture's ids/topology weren't updated with it. Mapper's fine; tests are stale. Corrected assertions below, verified passing.

🔴 `086d053` · `pytest tests/strands_evals/mappers/` → **5 failed, 373 passed** · `ruff` clean · with the five assertion fixes below: **378 passed**.

### 🔴 Five tests still assert the *old* live fixture

Every failure is `StopIteration`/`assert 3 == 2` from hardcoded old span ids (`91578cd4f1232ef0`, `8c1028e91e04dfa2`) and the old two-agent topology. The regenerated fixture has **three** agents with new ids. Verified-passing updates (378 passed with these applied):

<details>
<summary>Paste-ready fixes for the 5 tests</summary>

```python
# TestOpenAIAgentsPerAgentToolAttribution.test_live_session_has_exactly_two_agent_spans
# (worth renaming to ..._three_agent_spans)
assert len(agent_spans) == 3

# test_coordinator_tools
coordinator = next(
    s for s in all_spans if isinstance(s, AgentInvocationSpan) and s.span_info.span_id == "d72488b1a2d28f70"
)
tool_names = sorted(t.name for t in coordinator.available_tools)
assert tool_names == ["ask_math_specialist", "ask_research_specialist"]
assert all(t.description and t.parameters for t in coordinator.available_tools)  # pins the new definitions path

# test_math_specialist_tools
math_specialist = next(
    s for s in all_spans if isinstance(s, AgentInvocationSpan) and s.span_info.span_id == "186e8db9421d49d0"
)
tool_names = sorted(t.name for t in math_specialist.available_tools)
assert tool_names == ["add_numbers", "divide_numbers", "multiply_numbers"]

# TestOpenAIAgentsPromptEnrichment.test_coordinator_prompt_is_top_level_user_request
coordinator = next(
    s for s in all_spans if isinstance(s, AgentInvocationSpan) and s.span_info.span_id == "d72488b1a2d28f70"
)
assert coordinator.user_prompt == (
    "Look up the stock price for AMZN, then calculate what 1000 shares "
    "would be worth by multiplying the price by 1000."
)
assert coordinator.agent_response.startswith("The current stock price")

# test_math_specialist_prompt_is_delegated_query
math_specialist = next(
    s for s in all_spans if isinstance(s, AgentInvocationSpan) and s.span_info.span_id == "186e8db9421d49d0"
)
assert math_specialist.user_prompt == "240.98 multiplied by 1000"
assert math_specialist.agent_response
```

While you're in there, a `research_specialist` (`54044faa1adce44c`) assertion would pin the best part of the new fixture: `sorted(...) == ["get_weather", "lookup_stock_price"]` — see below for why that pair matters.

</details>

### The source change itself — verified working, and it closes a round-1 appendix item

`_backfill_agent_tools` (`generic_gen_ai_session_mapper.py:697-721`) + `_parse_tool_definitions` (`:435-457`) now read `gen_ai.tool.definitions` from child chat spans, with the called-tool-names path demoted to a fallback. Running the regenerated fixture end-to-end:

```
research_specialist -> get_weather (desc+params), lookup_stock_price (desc+params)
math_specialist     -> add_numbers, divide_numbers, multiply_numbers (all desc+params)
coordinator         -> ask_math_specialist, ask_research_specialist (all desc+params)
```

Two things worth calling out as genuinely better for evaluation quality: `research_specialist` now lists **`get_weather`, which was never called** — the first time a judge can see a foregone alternative (my round-1 note that `available_tools` was call-tautological is properly fixed, not worked around) — and tools now carry descriptions + parameter schemas, so the tool-parameter-accuracy rubric finally has a schema to check against. The regenerated fixture also uses payload-less agent spans, so `_backfill_agent_prompts` is now exercised by a shipped fixture too, not just the synthetic unit test. The OpenAI-style `{"type":"function","function":{...}}` unwrap in `_parse_tool_definitions` matches what the fixture actually carries.

Also checked: the ADOT fixture is untouched and its tests all pass; the definitions path only fills an agent whose `available_tools` is empty, and the name-only fallback still appends called-but-undeclared tools — sensible precedence.

<details>
<summary>What I ran · prior items</summary>

```
head 086d0530f3e1e4ef562d75ed753ffd111ef96499 (delta from 6971865: mapper +79 lines net, live fixture regenerated)

pytest tests/strands_evals/mappers/ -q               -> 5 failed, 373 passed   <- the finding
ruff check / format --check                          -> clean
mapper on regenerated fixture                        -> 3 agents, correct per-agent tools+prompts (values above)
raw-topology check                                   -> new format is genuinely flat (all agents parent to the
                                                        unconverted 'Agent Workflow' wrapper; no agent_handoff spans)
5 corrected assertions applied locally               -> 378 passed  (then tree restored)
```

Prior items unchanged and not re-raised: finding 3 risk-accepted · findings 6–8 withdrawn · vendor special-case accepted per your call · rounds 1–5 findings all fixed. Self-run follow-up; single-reviewer.

*Automated review — solid work for a human to approve, not a gate.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.

strandly-the-agent:
**Suite is green again — the round-7 🔴 is fixed, and nothing of mine remains open.** The amended head applies the five assertion updates (delta is tests-only: +27/-14, source and fixtures untouched) and adds the `test_research_specialist_tools` case pinning the best property of the new definitions path — `get_weather` listed as available despite never being called. No new findings.

✅ `916ece7` · mapper suite **379 passed** · mappers+extractors+types **511 passed** · `ruff check` + `format --check` clean · verified the delta touches only the test file.

Full ledger across the PR: findings 1/2/4 fixed at `5b1142b` · finding 5 fixed at `9fb48f5` · finding 3 risk-accepted · findings 6–8 withdrawn as theoretical · vendor special-case accepted · round-5 backfill-coverage 🟡 fixed at `6971865` · round-7 stale-tests 🔴 fixed here. From my side this is merge-ready pending a human maintainer's approval.

<details>
<summary>What I ran</summary>

```
head 916ece7dd848ca5b95e0677dfc3c81520081dd3f (amend of 086d053; delta tests-only, +27/-14)

pytest tests/strands_evals/mappers/ -q                       -> 379 passed (was 5 failed, 373 passed)
pytest mappers/ extractors/ types/ -q                        -> 511 passed
ruff check / ruff format --check (changed file)              -> clean
git diff 086d053..HEAD -- src/ fixtures/                     -> empty (no source or fixture change)
```

Self-run follow-up — the delta only repairs my round-7 finding. *Automated review — solid work for a human to approve, not a gate.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment. I removed the handoff reparenting because handoffs are semantically sibling agents.

strandly-the-agent:
**No blocking findings.** Reviewed the `916ece7..2210c15` delta — the handoff-reparenting removal is clean and complete, the sibling-agents semantic call makes sense to me, and the suite is green. One ⚪: the rewritten test's docstring describes the root-selection rule incorrectly (suggestion below). From my side the PR remains merge-ready pending human approval.

✅ `2210c15` · mapper suite **379 passed** · mappers+extractors+types **511 passed** · `ruff check` + `format --check` clean · no dangling reparenting references left in `src/` (remaining "handoff" hits are the unrelated, pre-existing `swarm_extractor.py`).

### The reversal, for the ledger

Accepting the semantics: OpenAI Agents handoffs transfer control between peers rather than nesting one agent under another, so parent/child was the wrong encoding — siblings it is. Two knock-on effects worth recording:

- **Finding 5's resolution changes shape but stays resolved.** My original ask (round 1) was that dropping `agent_handoff` shouldn't be *silent*. It's now a deliberate, stated design decision with a test pinning the flat shape — that's exactly the "decision rather than an accident" outcome I was after, even though the mechanism (reparenting) went away.
- **Findings 6–8 go from withdrawn to moot** — the code they lived in no longer exists.

### ⚪ The new test's docstring teaches the wrong root-selection rule

`test_handoff_agents_are_flat_with_coordinator_as_root` says *"root is the earliest-start agent"*. The actual rule in `_find_root_agent_span` (`types/trace.py:135-149`) is: **first parentless-with-content agent in span-list order**; `start_time` is only the final fallback when *no* agent is parentless. In this fixture both rules pick the coordinator, so the assertion is correct and stable (span order is deterministic from the fixture) — but I probed it: reversing the agent-list order flips the result to `c2828d53371818f0`, which start-time ordering never would. A maintainer who trusts the docstring will mispredict behaviour on the first trace where list order and start order diverge.

```suggestion
        """Handoff sub-agents are left independent (parentless); the root resolves to the
        coordinator as the first parentless agent with content (span-list order —
        start_time only breaks ties when no agent is parentless)."""
```

<details>
<summary>What I ran</summary>

```
head 2210c15fce1e7c50e713412d07dfcb79b01da84b (delta from 916ece7: -41 src / ~17 test lines)

pytest tests/strands_evals/mappers/ -q                 -> 379 passed
pytest mappers/ extractors/ types/ -q                  -> 511 passed
ruff check / ruff format --check (changed files)       -> clean
grep for dangling reparent/handoff refs in src/        -> only pre-existing swarm_extractor.py
root-selection probe on the ADOT fixture               -> both agents parentless; root = first-in-list
                                                          (reversed list -> different root; start_time not the rule)
```

Self-run follow-up — the delta is a scoped reversal with its test updated in the same commit. Full ledger: findings 1/2/4 fixed · 5 resolved (now via documented sibling semantics) · 3 risk-accepted · 6–8 moot (code deleted) · vendor special-case accepted · round-5 🟡 and round-7 🔴 fixed.

*Automated review — solid work for a human to approve, not a gate.*

</details>

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
arielnabavian, line 41: Could we create openai_agents_session_mapper.py with OpenAIAgentsSessionMapper(GenericGenAISessionMapper) that overrides two hooks: _pre_normalize(spans) and _post_enrich(trace, raw_spans) (both no-ops on the base).

liramon2: This was considered before, but is a one-way door decision. No other mapper uses inheritance, so doing it here would create the precedence for future mappers and existing ones to also use inheritance.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -72,6 +75,9 @@ def map_to_session(self, data: Any, session_id: str) -> Session:
 
         spans = self._normalize_to_flat_spans(data)
 
+        if any(get_scope_name(span) == SCOPE_OPENAI_AGENTS for span in spans):
```
jjbuck, line 78: The current order is effectively
```
  spans = normalize_input(data)
  spans = remove_openai_wrappers(spans)  # may remove session.id
  group_and_filter_by_session(spans)
```
If the wrapper is the only tagged span, line 98 sees no session metadata and includes every trace. I reproduced this with the checked-in ADOT fixture. Retaining `session.id` only on its wrapper makes both the real and an invalid session ID return the same six spans.

Could we filter using the original spans, then suppress wrappers only during conversion Let's also add a two- session test with tagged wrappers and untagged children.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -116,6 +122,13 @@ def _convert_trace(self, trace_id: str, spans: list[dict], session_id: str) -> T
         """Convert a list of dict spans (same trace_id) to a Trace."""
         converted_spans: list[InferenceSpan | ToolExecutionSpan | AgentInvocationSpan] = []
 
+        # Build raw parent map for bridge_parent_gaps (all spans, not just converted)
```
jjbuck, line 125: This is not actually a raw ancestry map after line 78. `bridge_parent_gaps` needs every raw parent edge, e.g.,
```
  A (converted) -> B (unconverted) -> C (converted)
```
It uses `B -> A` to reparent` C` to `A`. But if `B` is an OpenAI wrapper removed at line 78, `raw_parent_map` lacks B and so C becomes parentless instead.

Could we build this map from the original trace and separately decide which spans become canonical spans?

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -150,7 +163,14 @@ def _convert_trace(self, trace_id: str, spans: list[dict], session_id: str) -> T
                     e,
                 )
 
-        return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
+        bridge_parent_gaps(converted_spans, raw_parent_map)
+
+        trace = Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
+
+        if any(get_scope_name(span) == SCOPE_OPENAI_AGENTS for span in spans):
```
jjbuck, line 170: This is essentially
```
  if any(span_is_openai for span in spans):
      self._enrich_openai_agent_spans(trace, spans)
```
But `_enrich_openai_agent_spans` later selects `s for s in trace.spans if isinstance(s, AgentInvocationSpan)`.

There is no scope check in that selection. In a mixed-scope reproduction, adding one unrelated OpenAI chat span caused a custom agent to acquire prompt, response, and tool data that it did not otherwise receive.

Could we restrict enrichment to canonical spans whose raw IDs came from the OpenAI scope, or move this behind a producer-specific mapper?

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -426,6 +433,31 @@ def _convert_agent_invocation_span(self, span: dict, span_info: SpanInfo) -> Age
...
+        configs: list[ToolConfig] = []
+        for item in tool_list:
+            if isinstance(item, str):
+                if item:
+                    configs.append(ToolConfig(name=item))
+            elif isinstance(item, dict):
+                fn = item["function"] if isinstance(item.get("function"), dict) else item
+                name = fn.get("name", "")
+                if name:
+                    configs.append(
```
jjbuck, line 456: The risk here is that one malformed optional tool schema can remove the entire agent.

Suppose `parameters` is a JSON string  The failure path is
```
 ToolConfig validation fails
  -> _convert_trace catches ValueError
  -> AgentInvocationSpan is omitted
```
That effectively makes the agent execution disappear. Could we validate each optional field and fall back to `ToolConfig(name=name)` when `description` or `parameters` has an unsupported representation?

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
```diff
@@ -150,7 +163,14 @@ def _convert_trace(self, trace_id: str, spans: list[dict], session_id: str) -> T
                     e,
                 )
 
-        return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
+        bridge_parent_gaps(converted_spans, raw_parent_map)
```
jjbuck, line 166: This is a devex critique -- this call hides a multi-stage side effect / mutation pipeline.

The call stack is essentially
```
  bridge_parent_gaps: mutates span_info.parent_span_id
  Trace(...): mutates ToolExecutionSpan.agent_span_id
  OpenAI enrichment: mutates prompts, responses, and available_tools
```

The latter stages depend on mutations made by the earlier ones, but that dependency is not visible here. Let's make this explicit by making `bridge_parent_gaps` return rebuilt spans (preferred), or at least be renamed `reparent_converted_spans_in_place` and document that `Trace` derives ownership from the result?

This generic topology change also seems independently reviewable from broader OpenAI support.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
jjbuck: I'm a little confused by the design philosophy here. This class is documented as the scope-agnostic fallback, but the new logic here is actually producer-specific. Boiling it all down, it essentially does the following steps:

1. if OpenAI: remove wrappers
2. generic conversion
3. bridge_parent_gaps: mutate parent_span_id
4 Trace.model_post_init: mutate tool ownership
5. if OpenAI: mutate prompts, responses, and tools

  Correctness now depends on these side effects running in exactly this order. The two if `any(...SCOPE_OPENAI_AGENTS ...)` checks and the ~120-line “OpenAI Agents Handling” section are evidence that this is a separate producer dialect, not generic normalization (and note that as a general pattern, anytime you find yourself writing big comment banners, it's generally a tell-tale sign that the code would benefit from some refactoring).

The existing precedent keeps producer selection at dispatch essentially the following: `SCOPE_OPENINFERENCE_CLAUDE_AGENT_SDK -> OpenInferenceSessionMapper`.

  Could we similarly route OpenAI Agents through a dedicated mapper or producer adapter? Composition could be a good alternative to inheritance in this type of setting. The important invariant is that GenericGenAISessionMapper should not vary behavior by producer scope.

I.e., `bridge_parent_gaps` may be a generally useful fix, but its in-place mutation should be explicit and reviewed separately as a contract change for every generic producer. Wrapper suppression should also occur after session  partitioning because the wrapper may carry `session.id`.

jjbuck (changes requested):

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: We can return the rebuilt spans since that seems cleaner. Will add that change in a separate PR.

Update: PR is in https://github.com/strands-agents/evals/pull/375

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: By composition, if you mean the GenAI mapper has-a OpenAI Agents adapter, then I don't see much benefit over a separate mapper because there is no commonality between traces from the same producers and different instrumentations. To give an example with the existing mappers, the Langchain OTel and OpenInference mappers both handle Langchain produced traces, but they have completely different fields so their conversion logic cannot be extracted into a common adapter. This would leave us with a separate adapter for each instrumentation + producer combination, adapter boilerplate in the mappers, and mappers that must still know about the scope to select the right producer adapter. This is an issue for OpenAI Agents mapping because both Traceloop and OpenInference paths are planned to be supported (another case of same producer, different instrumentations).

If you mean the OpenAI Agents mapper has-a GenAI mapper, then it seems similar to inheritance. We would be trading less coupling for the inability to override methods. In this case, using inheritance and overriding `_convert_trace` from the GenAI mapper would be the simplest change to support OpenAI Agents. Otherwise, we would need a workaround to drop the workflow wrapper mid-conversion or re-think the bridge_parent_gaps + model_post_init logic; they depend on the span structure being finalized before running.

Semantically, inheritance would also make sense if `GenericGenAISessionMapper` were treated as a GenAI convention mapper (similar to the other mappers) rather than a scope-agnostic fallback. i.e. OpenAI Agents + Traceloop mapper is a GenAI convention mapper with small deviations. Either way, we would be adding a producer-specific mapper, breaking the current practice of writing convention-specific mappers (ex. traceloop, genai, openinference, ADK/gcp). I'll need you to confirm whether this is desired. This decision could also affect the `StrandsInMemoryMapper`, since it also mostly follows genai conventions.

Tl;dr: Which is preferred between the following?
1. GenAI has-a OpenAI adapter
2. OpenAI mapper has-a GenAI mapper
3. OpenAI mapper is-a GenAI mapper
4. Defer composition/inheritance decision to the future

I personally lean towards 3 or 4. @poshinchen seemed to prefer 4.

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: Done in 682b2f446f3fa5f6ec3bf5034b67b373f57d41dd

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: Done in 682b2f446f3fa5f6ec3bf5034b67b373f57d41dd

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: Done in 682b2f446f3fa5f6ec3bf5034b67b373f57d41dd

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: Done in 682b2f446f3fa5f6ec3bf5034b67b373f57d41dd

jjbuck (approved):

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
jjbuck: Thanks, this clarifies the distinction. My preference is for option 3, i.e., a private subclass with narrow hooks.

IMO the mapper boundary should follow the emitted telemetry dialect, not the application framework that produced it. OpenAI Agents instrumented through OpenInference can continue through OpenInferenceSessionMapper. It doesn't need to share a mapper with OpenAI Agents instrumented through Traceloop.

My read is that this is the way the current implementation already works, i.e., it uses a the generic conversion for everything except wrapper suppression and post-conversion enrichment.

I would avoid overriding _convert_trace, because that would duplicate or replace the ordering-sensitive pipeline we're trying to keep in one place. Instead, the base mapper can own the algorithm and expose two narrow, producer-neutral hooks:

class GenericGenAISessionMapper(SessionMapper):
    def _prepare_trace_spans(self, spans):
        return spans

    def _enrich_trace(self, trace, raw_spans):
        return trace

    def _convert_trace(self, trace_id, spans, session_id):
        raw_parent_map = ...
        spans_to_convert = self._prepare_trace_spans(spans)
        converted_spans = ...
        converted_spans = bridge_parent_gaps(converted_spans, raw_parent_map)
        trace = Trace(spans=converted_spans, ...)
        return self._enrich_trace(trace, spans_to_convert)


class _OpenAIAgentsGenAISessionMapper(GenericGenAISessionMapper):
    def _prepare_trace_spans(self, spans):
        # Suppress the Traceloop workflow wrapper.
        ...

    def _enrich_trace(self, trace, raw_spans):
        # Backfill OpenAI agent prompts, responses, and tools.
        ...

Then detect_otel_mapper would select the private subclass for SCOPE_OPENAI_AGENTS. GenericGenAISessionMapper would no longer import that scope or branch on producer identity.

Keeping the subclass private avoids creating a public extension contract or committing future mappers to inheritance.

With option https://github.com/strands-agents/evals/pull/4, merging the current implementation establishes the stronger precedent that the generic mapper may accumulate producer-specific scope checks and processing sections. IMO that's actually harder to unwind later.

jjbuck (changes requested):

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.

strandly-the-agent:
**TL;DR** — I independently verified the fixes for `jjbuck`'s review threads at the rebased head, and they hold: his mixed-scope leak repro comes back clean, the wrapper's `session.id` is honored, the parent map is built pre-filtering, and the new private-subclass-with-hooks shape is wired correctly. **One 🟡 residue in the malformed-schema fix**: a tool definition whose `name` isn't a string still drops the *entire agent span* — the same failure mode the fix targeted — verified one-line guard below. Suite green.

✅ `a11f390` · mapper suite **384 passed** · mappers+extractors+types **523 passed** · `ruff check`/`format` clean · `bridge_parent_gaps` return-value change confirmed to come from main (#375, merge-base `925e754`) — not smuggled into this PR.

### `jjbuck`'s threads, re-verified independently at this head

| Thread | Verified |
|---|---|
| Mixed-scope enrichment leak | ✅ re-ran his shape (custom-framework agent + one unrelated OpenAI chat span in the same trace): custom agent keeps `user_prompt=''`, `tools=[]` — enrichment is now scoped by per-span-id `openai_span_ids`, not the old trace-wide `any()` gate |
| `session.id` only on the wrapper span | ✅ wrapper removal now happens in `_prepare_trace_spans` inside `_convert_trace`, *after* session grouping — a trace whose only session tag rides the wrapper is included, wrapper still skipped, gap bridged |
| `raw_parent_map` built after filtering | ✅ built from the original spans at `generic_gen_ai_session_mapper.py:118-120`, before `_prepare_trace_spans` at `:122` |
| Malformed tool schema kills the agent span | ⚠️ mostly — see the 🟡 below |
| Hidden-mutation DevX | ✅ resolved via #375 on main, as `liramon2` said; this branch is rebased onto it |

Also: the coverage survived the re-architecture into the new subclass — I re-ran my historical mutants against the moved code (never-skip-wrapper → **2 failures**, no-op prompt-backfill → **4 failures**), and my old `known_scopes` nit is fixed as a side effect (`utils.py` now includes `SCOPE_OPENAI_AGENTS` in the fallback set).

### 🟡 The malformed-schema fix guards `description`/`parameters` but not `name`

`generic_gen_ai_session_mapper.py:458-486` (`_parse_tool_definitions`). Probing the fix with hostile definitions:

```
params-as-string   agent survives ✓ (tool kept name-only)
desc-as-dict       agent survives ✓
name-as-dict       agent DROPPED ✗   <- pydantic ValidationError is a ValueError,
name-as-int        agent DROPPED ✗      caught by _convert_trace's except → whole span skipped
```

So one malformed `name` still silently removes the whole agent from evaluation — the exact consequence the thread was about, one field over. Verified fix (all probes survive, 384 still pass):

```suggestion
                name = fn.get("name", "")
                if isinstance(name, str) and name:
```

### A correction to my own round-5 claim, for the record

I wrote that the enrichment "does not leak into other generic-genai traces." That was true only for the case I tested — a trace with *no* OpenAI spans at all, where the old `any()` gate stayed closed. `jjbuck`'s mixed-scope case (one OpenAI span alongside a custom-framework agent) tripped the gate and leaked, and my test never covered it. His finding was right where mine was incomplete; the per-span-id scoping in the new subclass fixes both cases, and this time I verified the mixed one.

<details>
<summary>On the new shape · what I ran</summary>

The private `_OpenAIAgentsGenAISessionMapper` + two documented base-class hooks (`_prepare_trace_spans` `:166`, `_enrich_trace` `:180`) is the design `jjbuck` and `liramon2` converged on, so I'm not relitigating it — just noting it also resolves my old appendix nit about undocumented extension points: both hooks now carry real contracts, and the class docstring says explicitly why it's private. The 405-line dedicated test file moved with the code.

```
head a11f390ecf5e17c2a0b479f528b35050af694e61 (rebased; delta since my last review = the 682b2f4 scoping fixes + the subclass split + rebase onto main incl. #375)

pytest tests/strands_evals/mappers/ -q                    -> 384 passed
pytest mappers/ extractors/ types/ -q                     -> 523 passed
ruff check / format --check                               -> clean
jjbuck mixed-scope repro                                  -> no leak (fixed)
session.id-on-wrapper repro                               -> trace included, wrapper skipped, bridged
malformed-defs probes                                     -> 4/6 survive; non-string name drops the agent (the yellow)
one-line name guard                                        -> all probes survive, 384 passed (then tree restored)
mutants: never-skip-wrapper / no-op prompt-backfill        -> 2 / 4 failures (coverage moved with the code)
```

Prior items unchanged and not re-raised: finding 3 risk-accepted · 6–8 moot · handoffs-are-siblings design recorded · all earlier findings fixed. *Automated review — solid work for a human to approve, not a gate.*

</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.

strandly-the-agent:
**The round-10 🟡 is fixed and verified — no new findings, nothing of mine open.** The delta is exactly the `name` guard plus a unit test that pins it (mixed malformed + valid definitions → only `valid_tool` survives, agent span intact). All six of my hostile-schema probes now pass, and I confirmed the new test *bites*: reverting the guard makes it fail. From my side the PR is clear — what remains is the humans' side: `jjbuck`'s CHANGES_REQUESTED review and `arielnabavian`'s one unresolved thread.

✅ `5e76978` · mapper suite **385 passed** · mappers+extractors+types **524 passed** · `ruff check` + `format --check` clean · delta is the one-line guard + one test, nothing else.

<details>
<summary>What I ran · full ledger</summary>

```
head 5e76978fd1d355fb946d33b964c85df1f050dd74 (delta from a11f390: +1 src line, +15 test lines)

pytest tests/strands_evals/mappers/ -q                 -> 385 passed
pytest mappers/ extractors/ types/ -q                  -> 524 passed
ruff check / ruff format --check                       -> clean
hostile-schema probes (params-as-string, desc-as-dict,
  name-as-dict, name-as-int, not-json, list-of-nulls)  -> all 6 agent spans survive (was 4/6)
guard reverted                                          -> new test FAILS (coverage is real; then tree restored)
```

Ledger across the PR: rounds 1–2 findings fixed · finding 3 risk-accepted · 5 resolved via sibling semantics · 6–8 moot · vendor special-case superseded by the private-subclass split · round-5 and round-7 items fixed · `jjbuck`'s threads verified fixed at `a11f390` · round-10 name-guard 🟡 fixed here. *Automated review — solid work for a human to approve, not a gate.*

</details>

src/strands_evals/mappers/generic_gen_ai_session_mapper.py
liramon2: Refactored the OpenAI Agents logic into a private subclass in a11f390ecf5e17c2a0b479f528b35050af694e61.

src/strands_evals/mappers/openai_agents_gen_ai_session_mapper.py
```diff
@@ -0,0 +1,157 @@
...
+                continue
+            agent = agent_spans.get(span.span_info.parent_span_id or "")
+            if agent is None or agent.available_tools:
+                continue
+            configs = self._parse_tool_definitions(raw_attrs_by_id.get(span.span_info.span_id, {}))
+            if configs:
+                agent.available_tools = list(configs)
+
+        # Fallback: get the names of tools the agent called.
+        for span in trace.spans:
```
jjbuck, line 96: If I'm reading this correctly, the scope fix is incomplete in this fallback. `agent_span_id` is not sufficient evidence that this is an OpenAI tool because `Trace.model_post_init` assigns otherwise unowned tools to the root agent.

Let's apply the same raw-scope restriction used above, i.e.,
```
  if (
      isinstance(span, ToolExecutionSpan)
      and span.span_info.span_id in scoped_span_ids
      and span.tool_call.name
  ):
      ...
```
Let's also add a mixed-scope regression containing an OpenAI agent and an unrelated custom-scope tool, asserting that the foreign tool is not added to available_tools.

jjbuck (changes requested):

src/strands_evals/mappers/openai_agents_gen_ai_session_mapper.py
liramon2: Done in b6eb4bb41f046e3101efd3ee371ecc01f37cd9dd

jjbuck (approved):

mehtarac (approved):

Issue #364, opened 2026-08-11
[FEATURE] Add support for OpenAI Agents SDK to OpenInference mapper


liramon2:
### Problem Statement

Teams building AI agents with OpenAI Agents SDK cannot use strands-evals to evaluate their agent traces if they instrument with OpenInference. Without this support, OpenAI users must either manually transform their traces into Session format or forgo trace-based evaluation entirely.

### Proposed Solution

Modify OpenInference mapper to support agents made with the OpenAI Agents SDK.

### Use Case

- As a user of OpenAI Agents SDK and OpenInference instrumentation, I want to evaluate my agent.

poshinchen:
@strandly-the-agent Here are the instrumentation:
1. "https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-openai-agents",                                                                                              
2. "https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-openai",                                                                                                          
3. "https://github.com/Arize-ai/openinference/tree/main/python/instrumentation/openinference-instrumentation-openai-agents",                                                                                    
4. "https://github.com/Arize-ai/openinference/tree/main/python/instrumentation/openinference-instrumentation-openai" 

Here are the current findings:

### For OpenAI agents SDK + traceloop instrumentation:
- It mostly follows GenAI conventions
- The instrumentation does not populate agent spans' available_tools . Tools only show up as descendant ToolExecution spans, so available_tools will be handled by backfilling from descendant tool spans
- The traces have an empty root span that serves as a wrapper. Its operation is invoke_agent instead of the genai convention of invoke_workflow . These should be stripped to avoid having an empty agent span.
- Agents that perform handoffs represent orchestrator and sub-agent spans as siblings. Need to use agent_handoff spans to correct parent ids if having valid parent ids is relevant

### For OpenAI agents SDK + OpenInference instrumentation:

- It mostly follows openinference conventions
- input.value and output.value are not agent span attributes. They are instead on descendant LLM spans. So they are handled with a walk down to those spans and looking into llm.input_messages.N.message.content.0.message_content.text and llm.output_messages.N.message.contents.0.message_content.text
[...]

PR #366, opened 2026-08-12
feat: add OpenAI Agents SDK support to OpenInference mapper

liramon2:
## Description
Add OpenAI Agents SDK support to OpenInference mapper. This handles both Cloudwatch ADOT and in-memory exporter formats. No inheritance or composition is used because that is a one-way door decision that calls for a larger refactor.

This also fixes parent span ids in `OpenInferenceSessionMapper` so that they point to converted spans. This preserves the agent-tool scopes when converting from OpenAI Agent traces (or other OpenInference traces) to evaluator inputs.

## Related Issues

https://github.com/strands-agents/evals/issues/364

## Type of Change

New feature

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -1023,7 +1023,7 @@ class OpenInferenceSessionMapper(SessionMapper):
                             parameters=tool_info.get("input_schema"),
                         )
                     )
-                except (json.JSONDecodeError, AttributeError, ValueError):
+                except (json.JSONDecodeError, AttributeError):
                     pass
 
         return sorted(tools, key=lambda t: t.name or "")
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -126,10 +126,13 @@ class OpenInferenceSessionMapper(SessionMapper):
         # shared conversion logic receives a uniform representation.
         for span in openinference_spans:
             scope = self._get_scope_name(span)
-            if scope == SCOPE_OPENINFERENCE_SMOLAGENTS:
-                self._normalize_smolagents_span(span)
-            elif scope == SCOPE_OPENINFERENCE_OPENAI_AGENTS:
-                self._normalize_openai_agents_span(span)
+            try:
+                if scope == SCOPE_OPENINFERENCE_SMOLAGENTS:
+                    self._normalize_smolagents_span(span)
+                elif scope == SCOPE_OPENINFERENCE_OPENAI_AGENTS:
+                    self._normalize_openai_agents_span(span)
+            except Exception as e:
+                logger.warning(f"Failed to normalize {scope} span {span.get('span_id', 'unknown')}: {e}")
 
         # Group spans by trace_id
         grouped = defaultdict(list)
@@ -140,7 +143,10 @@ class OpenInferenceSessionMapper(SessionMapper):
         # OpenAI Agents SDK normalization requires the full trace group, so it runs after grouping.
         for trace_spans in grouped.values():
             if any(self._get_scope_name(s) == SCOPE_OPENINFERENCE_OPENAI_AGENTS for s in trace_spans):
-                self._normalize_openai_agents_trace(trace_spans)
+                try:
+                    self._normalize_openai_agents_trace(trace_spans)
+                except Exception as e:
+                    logger.warning(f"Failed to normalize openai agents trace: {e}")
 
         # Build traces
         result_traces: list[Trace] = []
@@ -269,19 +275,23 @@ class OpenInferenceSessionMapper(SessionMapper):
 
     @classmethod
     def _normalize_openai_agents_span(cls, span: dict) -> None:
-        """Normalize OpenAI tool schemas in-place.
+        """Normalize OpenAI agent spans in-place.
 
         This normalizes by:
         1. Unwrapping tool schemas in the format {"type": "function", "function": {...}}.
-        2. Aliasing "parameters" to "input_schema".
+        2. Aliasing "parameters" to "input_schema" on tool schemas.
         """
         attrs = span.get("attributes") or {}
         for idx in cls._get_message_indices(attrs, "llm.tools."):
             key = f"llm.tools.{idx}.tool.json_schema"
-            try:
-                schema = json.loads(attrs.get(key))
-            except (TypeError, json.JSONDecodeError):
-                continue
+            raw = attrs.get(key)
+            if isinstance(raw, str):
+                try:
+                    schema = json.loads(raw)
+                except json.JSONDecodeError:
+                    continue
+            else:
+                schema = raw
             if not isinstance(schema, dict):
                 continue
             if "name" not in schema and isinstance(schema.get("function"), dict):
@@ -320,17 +330,19 @@ class OpenInferenceSessionMapper(SessionMapper):
                 attrs["input.value"] = user_prompt
 
             # Copy LLM tool schemas onto the AGENT span so it carries its own tools.
-            idx = 0
-            while first_attrs.get(f"llm.tools.{idx}.tool.json_schema") is not None:
-                attrs[f"llm.tools.{idx}.tool.json_schema"] = first_attrs[f"llm.tools.{idx}.tool.json_schema"]
-                idx += 1
+            for tool_idx in self._get_message_indices(first_attrs, "llm.tools."):
+                key = f"llm.tools.{tool_idx}.tool.json_schema"
+                schema = first_attrs.get(key)
+                if schema is not None:
+                    attrs[key] = schema
 
             last_attrs = llm_spans[-1].get("attributes", {})
             agent_response = self._get_last_message_text(last_attrs, "llm.output_messages.")
 
             # Fallback the response to the last message's tool calls if no LLM output was found
             if not agent_response:
-                if tool_calls := self._get_last_tool_calls(last_attrs):
+                tool_calls = self._get_last_tool_calls(last_attrs)
+                if tool_calls:
                     agent_response = f"[delegated] {tool_calls}"
             if agent_response:
                 attrs["output.value"] = agent_response
@@ -339,8 +351,12 @@ class OpenInferenceSessionMapper(SessionMapper):
         """Return the LLM spans descending from `span_id`, earliest first."""
         llm_spans: list[dict] = []
         stack = [span_id]
+        seen: set[str] = set()
         while stack:
             current = stack.pop()
+            if current in seen:
+                continue
+            seen.add(current)
             for child in spans_by_parent_id.get(current, []):
                 kind = (child.get("attributes") or {}).get("openinference.span.kind", "")
                 if kind == "LLM":
@@ -349,8 +365,6 @@ class OpenInferenceSessionMapper(SessionMapper):
                     cid = child.get("span_id", "")
                     if cid:
                         stack.append(cid)
-        # Sort via parse_timestamp so heterogeneous start_time types (int / ISO str /
-        # None) are normalized to comparable datetimes instead of raising TypeError.
         llm_spans.sort(key=lambda s: self.parse_timestamp(s.get("start_time")))
         return llm_spans
 
@@ -361,7 +375,7 @@ class OpenInferenceSessionMapper(SessionMapper):
         for key in attrs:
             if key.startswith(prefix):
                 seg = key.removeprefix(prefix).split(".", 1)[0]
-                if seg.isdigit():
+                if seg.isascii() and seg.isdigit():
                     indices.add(int(seg))
         return sorted(indices, reverse=True)
 
@@ -372,10 +386,30 @@ class OpenInferenceSessionMapper(SessionMapper):
             base = f"{prefix}{idx}.message"
             role_match = role is None or attrs.get(f"{base}.role") == role
             is_reasoning = attrs.get(f"{base}.contents.0.message_content.type") == "reasoning"
-            text = attrs.get(f"{base}.content") or attrs.get(f"{base}.contents.0.message_content.text")
+            text = attrs.get(f"{base}.content") or cls._join_content_text(attrs, base)
             if role_match and not is_reasoning and text:
                 return text
         return None
+
+    @staticmethod
+    def _join_content_text(attrs: dict, base: str) -> str | None:
+        """Concatenate all `text` parts under `{base}.contents.N`, or None.
+
+        Multimodal messages emit ordered parts (e.g. [image, text]), so scan every
+        part rather than only index 0.
+        """
+        parts: list[str] = []
+        i = 0
+        while True:
+            part_type = attrs.get(f"{base}.contents.{i}.message_content.type")
+            if part_type is None:
+                break
+            if part_type == "text":
+                text = attrs.get(f"{base}.contents.{i}.message_content.text")
+                if text:
+                    parts.append(text)
+            i += 1
+        return "".join(parts) or None
 
     @classmethod
     def _get_last_tool_calls(cls, attrs: dict) -> str | None:
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -38,11 +38,12 @@ from ..types.trace import (
 from .constants import (
     SCOPE_OPENINFERENCE,
     SCOPE_OPENINFERENCE_CLAUDE_AGENT_SDK,
+    SCOPE_OPENINFERENCE_OPENAI_AGENTS,
     SCOPE_OPENINFERENCE_SMOLAGENTS,
     SCOPES_OPENINFERENCE_FAMILY,
 )
 from .session_mapper import SessionMapper
-from .utils import safe_json_parse
+from .utils import bridge_parent_gaps, safe_json_parse
 
 logger = logging.getLogger(__name__)
 
@@ -132,6 +133,11 @@ class OpenInferenceSessionMapper(SessionMapper):
         for span in openinference_spans:
             trace_id = span.get("trace_id", "")
             grouped[trace_id].append(span)
+
+        # OpenAI Agents SDK normalization requires the full trace group, so it runs after grouping.
+        for trace_spans in grouped.values():
+            if any(self._get_scope_name(s) == SCOPE_OPENINFERENCE_OPENAI_AGENTS for s in trace_spans):
+                self._normalize_openai_agents_trace(trace_spans)
 
         # Build traces
         result_traces: list[Trace] = []
@@ -258,6 +264,65 @@ class OpenInferenceSessionMapper(SessionMapper):
             # Fallback: no tool.parameters available to map positional args
             attrs["input.value"] = json.dumps({"args": args})
 
+    def _normalize_openai_agents_trace(self, spans: list[dict]) -> None:
+        """Normalize OpenAI Agents SDK AGENT spans in-place by injecting input/output.
+
+        OpenAI Agents SDK AGENT spans carry no input.value or output.value.
+        The user prompt and final response live on descendant LLM spans
+        (reachable through intermediate CHAIN "turn" spans).
+        """
+        # Collect all spans keyed by parent id for span traversal
+        spans_by_parent_id: dict[str, list[dict]] = defaultdict(list)
+        for s in spans:
+            parent = s.get("parent_span_id")
+            if parent:
+                spans_by_parent_id[parent].append(s)
+
+        for span in spans:
+            attrs = span.get("attributes", {})
+            span_id = span.get("span_id", "")
+            if attrs.get("openinference.span.kind") != "AGENT" or attrs.get("input.value") or not span_id:
+                continue
+
+            # Walk from agent spans to their descendant LLM spans
+            llm_spans: list[dict] = []
+            stack = [span_id]
+            while stack:
+                current = stack.pop()
+                for child in spans_by_parent_id.get(current, []):
+                    kind = child.get("attributes", {}).get("openinference.span.kind", "")
+                    if kind == "LLM":
+                        llm_spans.append(child)
+                    elif kind == "CHAIN":
+                        cid = child.get("span_id", "")
+                        if cid:
+                            stack.append(cid)
+
+            if not llm_spans:
+                continue
+            llm_spans.sort(key=lambda s: s.get("start_time", 0))
+
+            # Get first user message from the earliest LLM span
+            first_attrs = llm_spans[0].get("attributes", {})
+            user_prompt = first_attrs.get("llm.input_messages.1.message.content")
+            if user_prompt:
+                attrs["input.value"] = user_prompt
+
+            # Copy LLM tool schemas onto the AGENT span so _extract_tools_from_attributes
+            # finds this agent's own tools.
+            idx = 0
+            while first_attrs.get(f"llm.tools.{idx}.tool.json_schema") is not None:
+                attrs[f"llm.tools.{idx}.tool.json_schema"] = first_attrs[f"llm.tools.{idx}.tool.json_schema"]
+                idx += 1
+
+            # Get text output from the last LLM span
+            last_attrs = llm_spans[-1].get("attributes", {})
+            agent_response = last_attrs.get(
+                "llm.output_messages.0.message.contents.0.message_content.text"
+            ) or last_attrs.get("llm.output_messages.0.message.content")
+            if agent_response:
+                attrs["output.value"] = agent_response
+
     def _build_trace(self, trace_id: str, spans: list[dict], session_id: str) -> Trace:
         """Build a Trace from spans with the same trace_id."""
         converted_spans: list[InferenceSpan | ToolExecutionSpan | AgentInvocationSpan] = []
@@ -278,6 +343,10 @@ class OpenInferenceSessionMapper(SessionMapper):
                         converted_spans.append(agent_span)
             except Exception as e:
                 logger.warning(f"Failed to convert span {span.get('span_id', 'unknown')}: {e}")
+
+        # Fix parent_span_id on converted spans that point to skipped intermediaries
+        raw_parent_map = {s.get("span_id", ""): s.get("parent_span_id") for s in spans}
+        bridge_parent_gaps(converted_spans, raw_parent_map)
 
         # In multi-agent LangGraph systems, each nested sub-graph produces its own
         # LangGraph CHAIN span. Keep only the last one (root graph finishes last).
@@ -380,7 +449,11 @@ class OpenInferenceSessionMapper(SessionMapper):
         # routing nodes that aren't true agent invocations — reject those by default.
         if span_kind == "AGENT":
             scope_name = self._get_scope_name(span)
-            if scope_name in (SCOPE_OPENINFERENCE_SMOLAGENTS, SCOPE_OPENINFERENCE_CLAUDE_AGENT_SDK):
+            if scope_name in (
+                SCOPE_OPENINFERENCE_SMOLAGENTS,
+                SCOPE_OPENINFERENCE_CLAUDE_AGENT_SDK,
+                SCOPE_OPENINFERENCE_OPENAI_AGENTS,
+            ):
                 input_val = attrs.get("input.value")
                 if input_val:
                     output_val = attrs.get("output.value")
@@ -637,6 +710,9 @@ class OpenInferenceSessionMapper(SessionMapper):
             return None
 
         available_tools = sorted(
+            self._extract_tools_from_attributes(attrs),
+            key=lambda t: t.name,
+        ) or sorted(
             self._trace_tools_map.get(trace_id, {}).values(),
             key=lambda t: t.name,
         )
@@ -826,14 +902,17 @@ class OpenInferenceSessionMapper(SessionMapper):
             if key.startswith("llm.tools.") and key.endswith(".tool.json_schema"):
                 try:
                     tool_info = json.loads(value) if isinstance(value, str) else value
+                    # OpenAI format wraps under {"type": "function", "function": {...}}
+                    if "function" in tool_info and "name" not in tool_info:
+                        tool_info = tool_info["function"]
                     tools.append(
                         ToolConfig(
                             name=tool_info.get("name"),
                             description=tool_info.get("description"),
-                            parameters=tool_info.get("input_schema"),
+                            parameters=tool_info.get("input_schema") or tool_info.get("parameters"),
                         )
                     )
-                except (json.JSONDecodeError, AttributeError):
+                except (json.JSONDecodeError, AttributeError, ValueError):
                     pass
 
         return sorted(tools, key=lambda t: t.name or "")
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -5,6 +5,7 @@ Handles traces from any producer in the OpenInference family:
 - openinference.instrumentation.langchain (LangChain / LangGraph)
 - openinference.instrumentation.smolagents (HuggingFace smolagents)
 - openinference.instrumentation.claude_agent_sdk (Claude Agent SDK)
+- openinference.instrumentation.openai_agents (OpenAI Agents SDK)
 
 Each producer emits spans following the OpenInference semantic conventions but
 with producer-specific encoding differences (e.g. attribute paths for message
@@ -68,12 +69,13 @@ class OpenInferenceSessionMapper(SessionMapper):
     - openinference-instrumentation-langchain (LangChain / LangGraph)
     - openinference-instrumentation-smolagents (HuggingFace smolagents)
     - openinference-instrumentation-claude-agent-sdk (Claude Agent SDK)
+    - openinference-instrumentation-openai-agents (OpenAI Agents SDK)
 
     Span type identification uses the openinference.span.kind attribute:
     - Inference spans: "LLM"
     - Tool execution spans: "TOOL"
-    - Agent invocation spans: "AGENT" (smolagents CodeAgent.run, Claude Agent SDK query) or
-      "CHAIN" with name="LangGraph" (LangGraph root graph)
+    - Agent invocation spans: "AGENT" (smolagents CodeAgent.run, Claude Agent SDK query,
+      OpenAI Agents SDK agent) or "CHAIN" with name="LangGraph" (LangGraph root graph)
 
     Producer-specific encoding differences (e.g. message attribute paths,
     tool argument wrapping) are normalized before shared conversion logic runs.
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -326,7 +326,12 @@ class OpenInferenceSessionMapper(SessionMapper):
         for span in spans:
             attrs = span.get("attributes") or {}
             span_id = span.get("span_id", "")
-            if attrs.get("openinference.span.kind") != "AGENT" or attrs.get("input.value") or not span_id:
+            if (
+                attrs.get("openinference.span.kind") != "AGENT"
+                or attrs.get("input.value")
+                or not span_id
+                or self._get_scope_name(span) != SCOPE_OPENINFERENCE_OPENAI_AGENTS
+            ):
                 continue
 
             llm_spans = self._get_descendant_llm_spans(span_id, spans_by_parent_id)
@@ -369,7 +374,7 @@ class OpenInferenceSessionMapper(SessionMapper):
             seen.add(current)
             for child in spans_by_parent_id.get(current, []):
                 kind = (child.get("attributes") or {}).get("openinference.span.kind", "")
-                if kind == "LLM":
+                if kind == "LLM" and self._get_scope_name(child) == SCOPE_OPENINFERENCE_OPENAI_AGENTS:
                     llm_spans.append(child)
                 elif kind == "CHAIN":
                     cid = child.get("span_id", "")
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -91,6 +91,8 @@ class OpenInferenceSessionMapper(SessionMapper):
         self._trace_tools_map: dict[str, dict[str, ToolConfig]] = defaultdict(dict)
         # Track system prompts per trace
         self._trace_system_prompt_map: dict[str, str] = defaultdict(str)
+        # Track span_id -> parent_span_id for ALL raw spans (including non-OpenInference scopes)
+        self._raw_parent_map: dict[str, str | None] = {}
         # Cache: span_id -> (input_messages, output_messages) from _get_messages_from_span_events
         # Avoids re-parsing span_events body during detection and again during conversion.
         self._span_messages_cache: dict[str, tuple[list[dict], list[dict]]] = {}
@@ -118,6 +120,9 @@ class OpenInferenceSessionMapper(SessionMapper):
 
         # Normalize input to flat spans
         spans = self._normalize_to_flat_spans(data)
+
+        # Build parent map from ALL spans
+        self._raw_parent_map = {s.get("span_id", ""): s.get("parent_span_id") for s in spans}
 
         # Filter to only spans from this scope (including smolagents variant)
         openinference_spans = [s for s in spans if self._get_scope_name(s) in SCOPES_OPENINFERENCE_FAMILY]
@@ -456,10 +461,6 @@ class OpenInferenceSessionMapper(SessionMapper):
             except Exception as e:
                 logger.warning(f"Failed to convert span {span.get('span_id', 'unknown')}: {e}")
 
-        # Fix parent_span_id on converted spans that point to skipped intermediaries
-        raw_parent_map = {s.get("span_id", ""): s.get("parent_span_id") for s in spans}
-        bridge_parent_gaps(converted_spans, raw_parent_map)
-
         # In multi-agent LangGraph systems, each nested sub-graph produces its own
         # LangGraph CHAIN span. Keep only the last one (root graph finishes last).
         agent_spans = [s for s in converted_spans if isinstance(s, AgentInvocationSpan)]
@@ -485,6 +486,9 @@ class OpenInferenceSessionMapper(SessionMapper):
             for converted in converted_spans:
                 if isinstance(converted, AgentInvocationSpan) and not converted.system_prompt:
                     converted.system_prompt = system_prompt
+
+        # Fix parent_span_id on converted spans that point to skipped intermediaries.
+        bridge_parent_gaps(converted_spans, self._raw_parent_map)
 
         return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
 
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -132,7 +132,8 @@ class OpenInferenceSessionMapper(SessionMapper):
                 elif scope == SCOPE_OPENINFERENCE_OPENAI_AGENTS:
                     self._normalize_openai_agents_span(span)
             except Exception as e:
-                logger.warning(f"Failed to normalize {scope} span {span.get('span_id', 'unknown')}: {e}")
+                span_id = span.get("span_id", "unknown")
+                logger.warning("scope=<%s>, span_id=<%s> | failed to normalize span: %s", scope, span_id, e)
 
         # Group spans by trace_id
         grouped = defaultdict(list)
@@ -146,7 +147,7 @@ class OpenInferenceSessionMapper(SessionMapper):
                 try:
                     self._normalize_openai_agents_trace(trace_spans)
                 except Exception as e:
-                    logger.warning(f"Failed to normalize openai agents trace: {e}")
+                    logger.warning("failed to normalize openai agents trace: %s", e)
 
         # Build traces
         result_traces: list[Trace] = []
@@ -275,14 +276,14 @@ class OpenInferenceSessionMapper(SessionMapper):
 
     @classmethod
     def _normalize_openai_agents_span(cls, span: dict) -> None:
-        """Normalize OpenAI agent spans in-place.
+        """Normalize OpenAI Agents spans in-place.
 
         This normalizes by:
         1. Unwrapping tool schemas in the format {"type": "function", "function": {...}}.
         2. Aliasing "parameters" to "input_schema" on tool schemas.
         """
         attrs = span.get("attributes") or {}
-        for idx in cls._get_message_indices(attrs, "llm.tools."):
+        for idx in cls._get_message_indices(attrs, "llm.tools"):
             key = f"llm.tools.{idx}.tool.json_schema"
             raw = attrs.get(key)
             if isinstance(raw, str):
@@ -301,11 +302,12 @@ class OpenInferenceSessionMapper(SessionMapper):
             attrs[key] = json.dumps(schema)
 
     def _normalize_openai_agents_trace(self, spans: list[dict]) -> None:
-        """Normalize OpenAI Agents SDK AGENT spans in-place by injecting input/output.
+        """Normalize OpenAI Agents SDK trace in-place.
 
-        OpenAI Agents SDK AGENT spans carry no input.value or output.value.
-        The user prompt and final response live on descendant LLM spans
-        (reachable through intermediate CHAIN "turn" spans).
+        This normalizes by:
+        1. Copying the user prompt, assistant response, and tool schemas from
+           LLM spans to parent AGENT spans
+        2. Falling back the assistant response to the last tool calls in the message
         """
         # Collect all spans keyed by parent id for span traversal
         spans_by_parent_id: dict[str, list[dict]] = defaultdict(list)
@@ -325,19 +327,19 @@ class OpenInferenceSessionMapper(SessionMapper):
                 continue
 
             first_attrs = llm_spans[0].get("attributes", {})
-            user_prompt = self._get_last_message_text(first_attrs, "llm.input_messages.", role="user")
+            user_prompt = self._get_last_message_text(first_attrs, "llm.input_messages", role="user")
             if user_prompt:
                 attrs["input.value"] = user_prompt
 
-            # Copy LLM tool schemas onto the AGENT span so it carries its own tools.
-            for tool_idx in self._get_message_indices(first_attrs, "llm.tools."):
+            # Copy LLM tool schemas onto the AGENT span for available_tools back-filling
+            for tool_idx in self._get_message_indices(first_attrs, "llm.tools"):
                 key = f"llm.tools.{tool_idx}.tool.json_schema"
                 schema = first_attrs.get(key)
                 if schema is not None:
                     attrs[key] = schema
 
             last_attrs = llm_spans[-1].get("attributes", {})
-            agent_response = self._get_last_message_text(last_attrs, "llm.output_messages.")
+            agent_response = self._get_last_message_text(last_attrs, "llm.output_messages")
 
             # Fallback the response to the last message's tool calls if no LLM output was found
             if not agent_response:
@@ -351,6 +353,7 @@ class OpenInferenceSessionMapper(SessionMapper):
         """Return the LLM spans descending from `span_id`, earliest first."""
         llm_spans: list[dict] = []
         stack = [span_id]
+        # Traverse the spans in top-down order
         seen: set[str] = set()
         while stack:
             current = stack.pop()
@@ -370,11 +373,12 @@ class OpenInferenceSessionMapper(SessionMapper):
 
     @staticmethod
     def _get_message_indices(attrs: dict, prefix: str) -> list[int]:
-        """Return the message indices present under `prefix`, highest first."""
+        """Return the numeric indices N found in keys starting with `{prefix}.N.`, highest first."""
+        prefix_dot = f"{prefix}."
         indices: set[int] = set()
         for key in attrs:
-            if key.startswith(prefix):
-                seg = key.removeprefix(prefix).split(".", 1)[0]
+            if key.startswith(prefix_dot):
+                seg = key.removeprefix(prefix_dot).split(".", 1)[0]
                 if seg.isascii() and seg.isdigit():
                     indices.add(int(seg))
         return sorted(indices, reverse=True)
@@ -383,7 +387,7 @@ class OpenInferenceSessionMapper(SessionMapper):
     def _get_last_message_text(cls, attrs: dict, prefix: str, role: str | None = None) -> str | None:
         """Return the text of the last matching message, or None."""
         for idx in cls._get_message_indices(attrs, prefix):
-            base = f"{prefix}{idx}.message"
+            base = f"{prefix}.{idx}.message"
             role_match = role is None or attrs.get(f"{base}.role") == role
             is_reasoning = attrs.get(f"{base}.contents.0.message_content.type") == "reasoning"
             text = attrs.get(f"{base}.content") or cls._join_content_text(attrs, base)
@@ -414,9 +418,9 @@ class OpenInferenceSessionMapper(SessionMapper):
     @classmethod
     def _get_last_tool_calls(cls, attrs: dict) -> str | None:
         """Return a text rendering of the last assistant message's tool calls, or None."""
-        prefix = "llm.output_messages."
+        prefix = "llm.output_messages"
         for idx in cls._get_message_indices(attrs, prefix):
-            base = f"{prefix}{idx}.message"
+            base = f"{prefix}.{idx}.message"
             calls: list[str] = []
             # Convert the tool calls to a string representation
             i = 0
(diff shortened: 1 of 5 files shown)


liramon2:
@strandly-the-agent Review this PR. Consolidate findings into a single comment.

strandly-the-agent:
Reviewed at `bc6422d` (base `1f752dc`) with a multi-pass agentic review — correctness, adversarial, test-quality, API/design, LLM-context and scope passes, each a fresh-context run, then de-duped under a reachability gate. Consolidated into one comment as you asked.

**Verdict: changes-requested.** The producer works on the two shipped fixtures, but `_normalize_openai_agents_trace` reads three hard-coded attribute paths, and I broke all three against the **real instrumentor's own source** (v2.0.0, the version your ADOT fixture came from): multi-turn chats get a mismatched (prompt, response) pair, reasoning models get the chain-of-thought scored as the answer, and multimodal input drops the agent span. Separately, your own handoff fixture loses the `coordinator` and misattributes its handoff call to the agent that received it.

| ✅/🔴 | Verified |
|---|---|
| ✅ | Mapper suite **385 passed** at `bc6422d` (main: 360) · extractors/types/detectors **235 passed** |
| ✅ | Hunk **D** (`bridge_parent_gaps`) is correct and is a genuine **prerequisite**, not a drive-by — don't split it out |
| ✅ | Hunk **F** is inert for langchain/smolagents/claude (0/9 AGENT spans across all 5 existing fixtures carry `llm.tools.*`; no upstream instrumentor emits them on non-LLM spans) |
| 🔴 | 5 failure modes reproduced in `_normalize_openai_agents_trace`; 3 confirmed against the instrumentor source |
| 🔴 | ADOT fixture: **3 raw AGENT spans → 1** `AgentInvocationSpan`; handoff tool misattributed |
| 🟡 | **10 of 22 mutants survive** all 385 tests; hunk D has **zero** tests |

I patched every finding below locally and re-ran: **385 + 235 still pass**, so these are cheap fixes rather than a redesign.

---

<details>
<summary><b>🔴 1 — <code>_normalize_openai_agents_trace</code>'s hard-coded attribute paths (5 reproduced failure modes)</b></summary>

`openinference_session_mapper.py:307` reads `llm.input_messages.1.message.content`; `:320-322` reads `llm.output_messages.0...`. I downloaded `openinference-instrumentation-openai-agents==2.0.0` and read `_processor.py` rather than guessing:

- **`_get_attributes_from_input(obj, msg_idx: int = 1)`** (line 282) with `enumerate(obj, msg_idx)` (line 284) — the Responses-API input list is numbered **from 1**, so `llm.input_messages.1` is the *first item of the input list*.
- **`_get_attributes_from_chat_completions_message_dicts(..., msg_idx: int = 0)`** (line 513), reached from `GenerationSpanData` (line 186 → 488) — that path numbers **from 0**.
- **`_get_attributes_from_response_output`** (line 684): a `reasoning` item occupies `msg_idx` and then increments it (lines 707-714), pushing the real answer to index 1.

| # | Manifestation | Who hits it | Result |
|---|---|---|---|
| a | **Multi-turn chat** — index 1 is the *oldest* user message | any agent using `result.to_input_list()` or `Session`/`SQLiteSession` (the documented multi-turn pattern) | `user_prompt='What is 15 multiplied by 37?'` paired with `agent_response='555 divided by 5 is 111.'` — a judge scores the answer against the wrong question |
| b | **Reasoning models** — summary sits at output index 0 | gpt-5/o-series with `reasoning.summary="auto"` on instrumentor v2.0.0 | `agent_response='The user wants 15*37. Multiply: 15*37=555.'` — the scratch-pad is scored as the answer. With encrypted reasoning and no summary, the AGENT span is dropped entirely |
| c | **Multimodal / structured content** — instrumentor emits `contents.0.message_content.text` and no scalar `.content` | the documented way to send images/files | `AgentInvocationSpan count = 0` — span silently dropped. Note `_normalize_smolagents_llm_attrs` (`:176`) **already** has this singular→plural fallback |
| d | **Chat-completions path** numbers input from 0 | LiteLLM / Azure / any OpenAI-compatible endpoint via the SDK | index 1 is the *second* message |
| e | **Empty-string or tool-call-only final turn** — guards are `if user_prompt:` / `if agent_response:`, then hunk E requires `output.value` | `tool_use_behavior="stop_on_first_tool"`, a max-turns cut-off, a guardrail trip | `AgentInvocationSpan count = 0` — exactly the runs an eval most wants to see |

All five are **silent** — no warning, and the eval result comes back plausible.

This fix resolves a–d (verified; **e needs a policy call from you**, not a path fix — see the Questions):

```python
_MSG_IDX_RE = re.compile(r"^llm\.(?:in|out)put_messages\.(\d+)\.")

@classmethod
def _message_indices(cls, attrs: dict, prefix: str) -> list[int]:
    """Descending message indices actually present under prefix (no contiguity assumed)."""
    out = set()
    for key in attrs:
        if key.startswith(prefix) and (m := cls._MSG_IDX_RE.match(key)):
            out.add(int(m.group(1)))
    return sorted(out, reverse=True)

@classmethod
def _last_user_message(cls, attrs: dict) -> str | None:
    """Text of the highest-indexed user input message, in either content shape."""
    for idx in cls._message_indices(attrs, "llm.input_messages."):
        base = f"llm.input_messages.{idx}.message"
        if attrs.get(f"{base}.role") != "user":
            continue
        if text := attrs.get(f"{base}.content") or attrs.get(f"{base}.contents.0.message_content.text"):
            return text
    return None

@classmethod
def _last_assistant_text(cls, attrs: dict) -> str | None:
    """Text of the highest-indexed non-reasoning assistant output message."""
    for idx in cls._message_indices(attrs, "llm.output_messages."):
        base = f"llm.output_messages.{idx}.message"
        if attrs.get(f"{base}.contents.0.message_content.type") == "reasoning":
            continue
        if text := attrs.get(f"{base}.contents.0.message_content.text") or attrs.get(f"{base}.content"):
            return text
    return None
```

then `user_prompt = self._last_user_message(first_attrs)` and `agent_response = self._last_assistant_text(last_attrs)`. Deriving indices from the keys also removes the contiguity assumption in the `llm.tools.{idx}` copy loop at `:313-316`.

</details>

<details>
<summary><b>🔴 2 — multi-agent handoff: the coordinator disappears and its handoff call is misattributed to the agent that received it</b></summary>

Run on **your own** `openai_agents_openinference_adot_spans.json`:

```
raw AGENT spans: 3 -> ['coordinator', 'Agent workflow', 'math_specialist']
surviving AgentInvocationSpans: 1
  span_id=2cbdd8e10afa264f
    user_prompt="What is 42 * 7, and what's the weather in Seattle?"
    available_tools=['add_numbers', 'divide_numbers', 'multiply_numbers']
  TOOL 'handoff to math_specialist' parent=None agent_span_id=2cbdd8e10afa264f
  TOOL 'multiply_numbers'           parent=2cbdd8e10afa264f agent_span_id=2cbdd8e10afa264f
```

- `"Agent workflow"` is dropped: the walk (`:289-299`) recurses only into `CHAIN` children and stops at nested `AGENT` children, so it collects no LLM spans and never gets `input.value`.
- `coordinator` is dropped: its LLM span emits only `tool_calls` (the handoff decision) and no text, so `output.value` is never set and hunk E (`:452-458`) rejects it. Its declared `transfer_to_math_specialist` / `transfer_to_research_specialist` schemas are lost with it.
- `math_specialist` survives carrying the **original top-level question**, so the trace reads as one agent that was asked the whole thing and answered it. The delegation is invisible.
- The handoff TOOL span gets `parent=None` from `bridge_parent_gaps` (its real ancestor was dropped), and then `Trace.model_post_init`'s orphan fallback (`types/trace.py:202-206`) assigns `agent_span_id = math_specialist` — **attributing the coordinator's delegation to the agent that received it**.

Concretely, a `ToolSelectionAccuracy` judge is asked to justify `handoff to math_specialist({})` while the "Available tool-calls" list it was shown is `add_numbers/divide_numbers/multiply_numbers` — the tool under evaluation isn't in the candidate set it was given.

For contrast, the Claude Agent SDK producer represents delegation legibly, as `Agent({'description': …, 'prompt': …})` with the subagent's real answer as `tool_result`.

The misattribution is wrong regardless of intent, so it needs fixing either way; whether the coordinator *should* survive is a design call I've put in the Questions. Cheapest correct step: don't let an orphaned tool span inherit `agent_span_id` from an unrelated agent — leave it `None` and let the extractors treat it as unowned.

Worth noting `test_math_specialist_agent_span` currently asserts only "at least one agent span" matching `42` or `weather`, so it **encodes this collapse as expected**. There's no agent-count assertion for this fixture (the live one has `test_one_agent_span_per_trace`).

</details>

<details>
<summary><b>🟡 3 — the normalizer runs outside every <code>try/except</code>, so one bad span can lose the whole session (including other producers' traces)</b></summary>

`map_to_session:137-140` calls the normalizer **before** any span reaches `_build_trace`'s per-span `try/except Exception` + `logger.warning` (`:331`, `:344-345`). Reproduced — a healthy LangChain trace batched with one awkward openai_agents trace:

```
healthy LangChain trace alone -> ['t-langchain']
TypeError ESCAPED map_to_session: '<' not supported between instances of 'int' and 'str'
  -> 0 traces returned: the healthy LangChain trace is destroyed too
attributes=None variant: AttributeError ESCAPED map_to_session: 'NoneType' object has no attribute 'get'
```

Triggers: `llm_spans.sort(key=lambda s: s.get("start_time", 0))` (`:303`) on heterogeneous/`None` `start_time`, and `attributes: None` (`:282`/`:293`).

**Honest reachability, which is why this is 🟡 and not 🔴:** I could *not* reach it through your own `CloudWatchLogsParser` — it drops records lacking `startTimeUnixNano`, so values come out uniform. But `SessionMapper.parse_timestamp` (`session_mapper.py:77`) explicitly documents `start_time` as possibly str/datetime/`None`, so the class's declared input contract permits the crashing input — and the correct idiom already exists next door in `adk_otel_session_mapper.py:194`. Every other per-span quirk here degrades to a warning; this one takes unrelated producers down with it.

```python
llm_spans.sort(key=lambda s: self.parse_timestamp(s.get("start_time")))
```
plus wrapping the call:
```python
try:
    self._normalize_openai_agents_trace(trace_spans)
except Exception as e:
    logger.warning("Failed to normalize OpenAI Agents trace: %s", e)
```
With both applied the batch above returns `['t-langchain']` and logs a warning.

</details>

<details>
<summary><b>🟡 4 — test gaps: 10 of 22 mutants survive all 385 tests, and hunk D is completely unpinned</b></summary>

Deleting the `bridge_parent_gaps` call at `:349` outright leaves **123 passed** in the mapper file and **235 passed** in extractors/types/detectors. The correct behaviour it provides (3 spans re-parented onto `math_specialist`) is real and worth pinning.

Other survivors: input taken from the *last* LLM descendant instead of the first · the entire `llm.tools.*` copy made a no-op · the `.message.content` output fallback deleted · hunk F's `_trace_tools_map` fallback deleted · the `and "name" not in tool_info` guard dropped · the `or tool_info.get("parameters")` fallback dropped · the added `ValueError` removed · the `attrs.get("input.value")` skip guard removed · the `if user_prompt:` guard removed.

Also: hunk F is pinned by exactly one test (`test_openai_agent_span_conversion`), which hand-builds an AGENT span carrying its own `llm.tools.0` — a shape no real producer emits. 5 of the 10 `TestOpenAIAgentsScopeSupport` tests only assert `_is_*_span(...) is True/False` with no mapping exercised. And `test_handoff_tool_span_present` asserts `len(tool_spans) >= 1` without filtering by name, so it passes even if the handoff span specifically is dropped.

Two that pass today and close the D / agent-count gaps:

```python
def test_math_specialist_children_reparented(self, openai_agents_adot_session):
    all_spans = [s for t in openai_agents_adot_session.traces for s in t.spans]
    specialist = next(s for s in all_spans if isinstance(s, AgentInvocationSpan) and "42" in s.user_prompt)
    multiply = next(s for s in all_spans if isinstance(s, ToolExecutionSpan) and s.tool_call.name == "multiply_numbers")
    inferences = [
        s for s in all_spans
        if isinstance(s, InferenceSpan) and s.span_info.parent_span_id == specialist.span_info.span_id
    ]
    assert multiply.span_info.parent_span_id == specialist.span_info.span_id
    assert len(inferences) == 2

def test_adot_agent_span_count_is_one(self, openai_agents_adot_session):
    for trace in openai_agents_adot_session.traces:
        assert len([s for s in trace.spans if isinstance(s, AgentInvocationSpan)]) == 1
```

Plus one that **fails** today and pins the appendix-1 regression: a `make_span` AGENT with `"llm.tools.0.tool.json_schema": "42"` should still yield 1 agent span with `available_tools == []`.

</details>

### Questions

1. **Is the coordinator's disappearance the intended representation of a handoff?** An argument exists for "the agent that produced the final answer *is* the trace" — but then `math_specialist` inheriting the top-level question and owning the handoff call reads as one agent doing everything, and the coordinator's `transfer_to_*` tools vanish. Would you rather a tool-call-only turn counted as a real agent invocation? *(blocking — it decides whether 🔴 2 is a bug or a documented limitation, and whether manifestation (e) above needs the same treatment.)*
2. **#365 vs #366 — two techniques for one problem.** #365 attaches tools directly to the converted `AgentInvocationSpan` (`openai_agents_otel_session_mapper.py:61-68`), whereas here hunk C fabricates `llm.tools.*` onto the *raw* AGENT span (`:313-316`) so hunk F (`:701-706`) can read it back. The `input.value`/`output.value` half of C is load-bearing for the `:452-458` gate and clearly earns its place, but would the tool half be simpler as the #365 approach? Both PRs also make the same parent-span-id fix in parallel — is one meant to land first? *(non-blocking)*
3. Following the #340 → #353 precedent, is there a tracking issue for `tests_integ/test_openai_agents_*`? I couldn't find one. *(non-blocking)*

<details>
<summary><b>Appendix — non-blocking (7)</b></summary>

1. ⚪ **`:889` `if "function" in tool_info and …` assumes a dict.** A `llm.tools.N.tool.json_schema` decoding to a JSON scalar raises `TypeError`, which isn't in the `except (json.JSONDecodeError, AttributeError, ValueError)` at `:898`; it escapes and `_build_trace`'s broad `except Exception` drops the **whole span**, where main skipped just that one tool. Strictly contrived — I checked openai_agents v1.6.1/v2.0.0, vendored langchain (which gates `LLM_TOOLS` to LLM runs) and claude_agent_sdk (emits no `llm.tools.*` at all): **no real instrumentor can emit it**. But it's a regression vs main and the fix is one word: `isinstance(tool_info, dict) and …`.
2. ⚪ **The descendant walk at `:289-299` has no `visited` set**, unlike `bridge_parent_gaps` (`utils.py:298-301`). A duplicate `span_id` closing a cycle hangs the process with unbounded `llm_spans` growth (confirmed, >5s with no termination); self-parented and diamond variants terminate fine. Contrived, one-line fix.
3. ⚪ **Hunk C copies tool schemas from the earliest LLM span only**, so a tool list that changes across turns loses entries (crafted case: 1 tool instead of 3). Reachable via `FunctionTool.is_enabled` per-turn filtering or MCP with `cache_tools_list=False`. Uncommon.
4. ⚪ **The normalizer never checks scope** — `map_to_session:139` gates on "any span in the group is openai_agents" but hands the walk the whole group. Sibling spans are safe (I checked: a smolagents AGENT span sharing a `trace_id` is *not* mutated), so this needs a non-openai AGENT span to be an *ancestor* of openai `CHAIN`/`LLM` spans. Contrived, but a scope check in the loop would make the intent explicit.
5. ⚪ Hunk F is safe for the 3 existing producers by *upstream construction*, not by a guard here — worth a comment saying so, since nothing in the suite would catch a future instrumentor that starts attaching `llm.tools.*` to AGENT spans.
6. ⚪ Fixture provenance skew: the live fixture is instrumentor **v1.6.1**, the ADOT one **v2.0.0**. Worth a word on whether that's deliberate.
7. ⚪ Correcting myself on a couple of things a reviewer might otherwise raise, in case they come up: the in-place mutation of caller span dicts is **pre-existing** in this class since #308 (`main:191`, `:204`, `:256`, `:259`) and is observably idempotent, so it isn't this PR's problem; and the `_adot_` fixture *is* legitimately ADOT-captured (`aws.genai.*`, `PlatformType: AWS::BedrockAgentCore`) — the existing `claude_adot` fixture has the same flat shape, so "it doesn't exercise `_parse_adot_output`" isn't a fault of this PR either.

Three pre-existing items are out of scope here but might deserve their own issues — happy to file them if useful: the orphan-fallback in `types/trace.py:202-206` silently attributing *any* unowned tool span to the sole surviving agent (affects every mapper); the `is_langchain` dedup exemption at `:351-357` having had no test since #340; and no mapper documenting the `AgentInvocationSpan`-per-`Trace` expectation that `trace_extractor` and `root_cause_analyzer` each interpret differently.

</details>

<sub>Automated multi-pass review by `strandly-the-agent`. Every 🔴/🟡 above has a repro I ran at `bc6422d`, and I applied all the suggested fixes locally to confirm they work and keep 385 + 235 green — but I'm an experimental agent, so please treat this as work for a human to approve rather than a verdict. Happy to dig into any of it further.</sub>

poshinchen:
@strandly-the-agent Here are the instrumentation:
1. "https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-openai-agents",                                                                                              
2. "https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-openai",                                                                                                          
3. "https://github.com/Arize-ai/openinference/tree/main/python/instrumentation/openinference-instrumentation-openai-agents",                                                                                    
4. "https://github.com/Arize-ai/openinference/tree/main/python/instrumentation/openinference-instrumentation-openai" 

Here are the current findings:

### For OpenAI agents SDK + traceloop instrumentation:
- It mostly follows GenAI conventions
- The instrumentation does not populate agent spans' available_tools . Tools only show up as descendant ToolExecution spans, so available_tools will be handled by backfilling from descendant tool spans
- The traces have an empty root span that serves as a wrapper. Its operation is invoke_agent instead of the genai convention of invoke_workflow . These should be stripped to avoid having an empty agent span.
- Agents that perform handoffs represent orchestrator and sub-agent spans as siblings. Need to use agent_handoff spans to correct parent ids if having valid parent ids is relevant

### For OpenAI agents SDK + OpenInference instrumentation:

- It mostly follows openinference conventions
- input.value and output.value are not agent span attributes. They are instead on descendant LLM spans. So they are handled with a walk down to those spans and looking into llm.input_messages.N.message.content.0.message_content.text and llm.output_messages.N.message.contents.0.message_content.text
- Available tools from _trace_tools_map are scoped to the trace while multiple agents can exist in the trace. This makes judges compare orchestrator tool calls against all available tools in the trace, causing false positives (bug seen last week). This can be handled by copying LLM tool schemas llm.tools.N.tool.json_schema onto AGENT spans and extracting them into available_tools . Granted, this can be deferred since the other frameworks using openinference also don't handle multi-agent traces correctly.


NOTE: In the current mappers' implementation strategy, I don't think we should use composition, and inheritance. And the mappers are grouped by instrumentation frameworks.

I need you to:
1. verify the above claims, 
2. tell me how to implement the classes, should they be an additional class or just have the logics in the current mapper(s). 
3. I don't think we need inheritance for now. Tell me the suggestions

strandly-the-agent:
Verified all seven claims against the instrumentation sources (`openinference-instrumentation-openai-agents` 2.0.0, `opentelemetry-instrumentation-openai-agents` 0.62.3 from Traceloop) and the captured fixtures in both PRs (#366 at `bc6422d`, #365 at `ebc5c5a`). Six hold; three need precision that changes the remedy.

| # | Claim | Verdict |
|---|---|---|
| T1 | Traceloop mostly follows GenAI conventions | ✅ (deviations noted below) |
| T2 | Traceloop doesn't populate available_tools → backfill from descendant tool spans | ⚠️ true for the fixtures, but **current v0.62.3 emits `gen_ai.tool.definitions`** — prefer it; backfill has a semantic trap |
| T3 | Empty root wrapper with `invoke_agent` → strip it | ✅ (one precision: semconv has no `invoke_workflow`, and the wrapper's attrs change across versions) |
| T4 | Handoff agents are siblings → reparent via `agent_handoff` spans | ✅ (v0.62.3 adds a simpler signal: `gen_ai.agent.handoff_parent` on the sub-agent span) |
| O1 | OpenInference mostly follows its conventions | ✅ |
| O2 | input/output live on descendant LLM spans → walk + read message paths | ⚠️ mechanism right; the key paths as written are off, and index selection is the actual hard part |
| O3 | `_trace_tools_map` is trace-scoped → multi-agent false positives; copy `llm.tools.*` onto AGENT spans; deferrable | ✅ — but the copy alone doesn't close the bug you saw; the orchestrator has to survive too |

**Design answer (Q2/Q3): agree — no inheritance, and no new classes.** `openai_agents` is a *producer inside two convention families*, not a framework of its own: keep its logic inside the two family mappers (`OpenInferenceSessionMapper`, `GenericGenAISessionMapper`) as scope-gated hooks in a fixed three-phase shape — **normalize → shared convert → post-link**. Applied consistently, that means folding #365's subclass back into `GenericGenAISessionMapper`. Details and a sketch below.

<details>
<summary><b>Traceloop verification (T1–T4)</b></summary>

**T1 — GenAI conventions ✅.** Fixture chat spans carry `gen_ai.operation.name` (`chat`/`execute_tool`/`invoke_agent`), the new semconv message format (`gen_ai.input.messages`/`gen_ai.output.messages` as role/parts JSON), `gen_ai.usage.*`, `gen_ai.agent.name/id/description`. Deviations to plan for: per-turn wrapper spans with `gen_ai.operation.name: "unknown"` (8–12 attrs, no content); the custom `agent_handoff` operation (not in semconv); deprecated `gen_ai.system` alongside `gen_ai.provider.name`.

**T2 — ⚠️ version-dependent, and the backfill has a semantic trap.**
- In both #365 fixtures the claim is exactly right: the only tool-ish attributes anywhere are `gen_ai.tool.call.{name,arguments,result,type}` on `execute_tool` spans — zero declared-tool schemas.
- But the **current release emits the real declared list**: v0.62.3's `_end_generation_span` sets `gen_ai.tool.definitions` (full name/description/parameters JSON, via `_extract_tool_definitions`) on generation/response spans when content tracing is on (`_hooks.py:1021-1028`; runs for `GenerationSpanData`/`ResponseSpanData`, `:800-803`). The #365 fixtures simply predate it.
- So the remedy should be: **prefer `gen_ai.tool.definitions` when present, fall back to backfill**. Two reasons the backfill alone is worth improving: (i) descendant `ToolExecutionSpan`s are tools *used*, not tools *available* — a tool-selection judge sees a shrunken candidate set with the un-chosen options missing, which is the mirror image of the false-positive bug you saw on the OpenInference side (candidate set too big there, too small here); (ii) #365's backfill builds `ToolConfig(name=...)` only — no description/parameters for the judge to reason over, while `gen_ai.tool.definitions` carries both.

**T3 — ✅ confirmed in both fixtures.** Root `Agent workflow` span with `gen_ai.operation.name: "invoke_agent"` and no agent name / no messages (live: 5 attrs, all provider/server boilerplate). Stripping is right, and #365's rule (no `gen_ai.agent.name` **and** no `gen_ai.input.messages` → skip) is the robust one — better than keying on the operation name, because the wrapper's attributes are version-dependent: in v0.62.3 the root gets `traceloop.span.kind: workflow` and **no** `gen_ai.operation.name` at all (`on_trace_start`, `_hooks.py:676-688`). One nit on the claim's wording: GenAI semconv defines no `invoke_workflow` operation (`chat`, `execute_tool`, `invoke_agent`, `create_agent`, `embeddings`, …) — workflow-ness is Traceloop's own `traceloop.span.kind`, so there's no "correct" operation the wrapper *should* have had; it just needs stripping.

**T4 — ✅ confirmed.** In the ADOT fixture, `invoke_agent coordinator` and `invoke_agent math_specialist` are **siblings** (children of the same turn wrapper under the root), with the `agent_handoff` span (`gen_ai.handoff.from_agent: coordinator`, `gen_ai.handoff.to_agent: math_specialist`) sitting under the coordinator's turn — the name-based reparenting in #365's `_apply_handoff_reparenting` is the right join, and the `ebc5c5a` tiebreaker doc covers the duplicate-name case. Two additions:
- The **agents-as-tools pattern needs no correction** — in the live fixture, `execute_tool ask_math_specialist` → nested `invoke_agent math_specialist` nests properly already. Sibling-flattening is handoff-specific.
- v0.62.3 also stamps **`gen_ai.agent.handoff_parent`** directly on the handed-off agent's own `invoke_agent` span (`utils.py:14`, applied in `_start_agent_span` via `_reverse_handoffs_dict`). When present it's a simpler, collision-free signal than the handoff-span join — worth preferring, with the span-join as fallback for older captures like the fixture.

</details>

<details>
<summary><b>OpenInference verification (O1–O3)</b></summary>

**O2 — mechanism ✅, paths and indices need precision.** Verified against the v2.0.0 instrumentor source and both #366 fixtures:
- **Input**: plain-text messages carry the scalar `llm.input_messages.N.message.content` — that's the only input shape in both shipped fixtures. The `contents.0.message_content.text` shape appears for structured/multimodal input (`_get_attributes_from_message_param` emits one or the other, not both). The path as written in your comment (`...message.content.0.message_content.text`) mixes the two — a mapper reading only that would drop every plain-text trace. Read scalar first, `contents.0...text` as fallback (that's also exactly review finding 1c: the current PR code reads only the scalar and drops multimodal spans).
- **Output**: `contents.0.message_content.text` on v2.0.0, scalar `.message.content` on v1.6.1 — both fallbacks needed (the PR does this part right).
- **The hard part is N, not the path**: the Responses-API path numbers input from **1** (`_get_attributes_from_input`, `msg_idx=1`), the chat-completions path (`GenerationSpanData` — LiteLLM/Azure/compatible endpoints) from **0**, and reasoning items consume output indices. So: highest-index message with `role == "user"` for input, highest-index non-reasoning text for output — hardcoded indices break multi-turn, reasoning models, and the chat-completions path (review finding 1, with repros).

**O3 — ✅ verified, with one important limit.** `_trace_tools_map[trace_id]` is populated from every LLM span in the trace during conversion, so in a multi-agent trace each agent's `available_tools` is the trace-wide union — your false-positive mechanism is exactly what my LLM-context pass reproduced (a judge asked to justify `handoff to math_specialist({})` against a candidate set that doesn't contain it). Copying `llm.tools.N.tool.json_schema` onto AGENT spans fixes the *scoping*, but as shipped it doesn't close the bug you saw, for two reasons:
- The **orchestrator's AGENT span is currently dropped** (its final LLM turn is tool-calls-only, so `output.value` never gets injected and the `_is_agent_invocation_span` gate rejects it). Its handoff tool span is then misattributed to the surviving specialist by the orphan fallback (`types/trace.py:202-206`) and judged against the *specialist's* tools — same species of false positive, one hop over. The fix needs the orchestrator to survive (or orphaned tool spans to stay unowned), not just the tool copy. That's review blocker 2 / Question 1 — it's the decision this hinges on.
- The copy takes schemas from the **earliest LLM span only**, so per-turn tool filtering (`FunctionTool.is_enabled`, MCP without cache) loses entries. Deriving indices from keys and merging turns is cheap.
- The "deferrable because other openinference producers don't handle multi-agent either" part is accurate: langchain dedups to one agent span (`is_langchain` gate), and smolagents/claude multi-agent traces get the same trace-wide union today.

</details>

<details>
<summary><b>Design recommendation (Q2/Q3): no new classes, no inheritance — three named phases inside the two family mappers</b></summary>

**The rule that matches the repo you already have:** one mapper per *emitting-convention family* (`OpenInferenceSessionMapper` already serves langchain/smolagents/claude via scope-gated tweaks; `GenericGenAISessionMapper` serves GenAI-convention traces), routed by `detect_otel_mapper` on scope. `openai_agents` is a producer *inside* two families — OpenInference scope in #366, GenAI scope in #365 — so it doesn't earn a class in either. A per-producer class would duplicate each family's conversion machinery (span-kind detection, message parsing, ADOT handling) for an 80-line delta.

**Why I'd agree about inheritance specifically:** #365's subclass works by overriding `_convert_trace` / `_convert_agent_invocation_span` — it's coupled to `GenericGenAISessionMapper` calling those exact template methods in that exact order. A refactor of the base silently changes the subclass's behaviour with no test failing in the subclass's own file (fragile-base-class). Scope-gated hooks keep the coupling visible in one place. Composition (a strategy object per producer) would also work but buys nothing over a plain method-dispatch table at this size — it's structure without payoff.

**The shape I'd standardize (both mappers already approximate it):**

1. **Normalize** (per-producer, before shared conversion): rewrite raw span dicts into the family's canonical shape. `_normalize_smolagents_span` and `_normalize_openai_agents_trace` are this today. Registered in a dispatch table rather than a growing if-chain:
   ```python
   # producer scope -> (granularity, hook)
   _PRODUCER_NORMALIZERS = {
       SCOPE_OPENINFERENCE_SMOLAGENTS: ("span", _normalize_smolagents_span),
       SCOPE_OPENINFERENCE_OPENAI_AGENTS: ("trace", _normalize_openai_agents_trace),
   }
   ```
2. **Convert** (shared, producer-blind): one code path per family. Producer conditionals should not appear here — if one is needed, the data should have been normalized instead.
3. **Post-link** (per-producer where needed, after conversion): structural fixes on converted spans — #365's handoff reparenting and tools backfill live here; `bridge_parent_gaps` is the family-agnostic version and stays shared in `utils`.

Three guardrails, all from bugs already found in these two PRs: each hook is **exception-isolated** (one producer's quirk must degrade to a warned-and-skipped trace, not kill the session — the #366 normalizer currently runs outside every try/except); each hook **checks scope on the spans it touches** (the #366 walk doesn't); and each hook gets **producer fixtures that pin it** (mutation testing showed the `bridge_parent_gaps` call can be deleted with all 385 tests green).

**Applied to the two PRs:**
- **#366**: keep everything in `OpenInferenceSessionMapper` — no new class. The changes it needs are the review fixes plus, optionally, the dispatch table above.
- **#365**: the consistent no-inheritance answer is to fold `OpenAIAgentsOtelSessionMapper` into `GenericGenAISessionMapper`: wrapper-strip becomes a normalize/skip rule, handoff-reparent + tools-backfill become a post-link step, both gated on `SCOPE_OPENAI_AGENTS`, and `detect_otel_mapper:140-141` routes to the generic mapper. It's ~118 lines whose three behaviours are already discrete methods, so folding is cheapest now, before it grows. Keeping the subclass is *workable*, but then the codebase carries two different strategies for the same problem — which is the divergence that prompted your question.

One thing I'd *not* do: don't introduce a producer-level abstraction (base class or protocol for "producers") until a third family needs it. Two call sites is a pattern; codifying it early is how the mapper layer gets rigid.

</details>

<sub>Verification by `strandly-the-agent`: claims checked against instrumentor sources (openinference 2.0.0, traceloop 0.62.3) and both PRs' fixtures; OpenInference items re-verified with the repros from my review above. As ever — my analysis to check, your call to make.</sub>

liramon2:
@strandly-the-agent Review this PR. Consolidate findings into a single comment. For your questions:
- The coordinator should exist, but the workflow wrapper should not
- They can land in any order. The parallel fix was a coincidence
- No live integration tests yet

strandly-the-agent:
Re-reviewed at `1c9587f` (`fix: address strandly` + `refactor: consolidate tool schema normalization`, base `1f752dc`). Follow-up shape: I re-ran **every round-1 repro cold** against the new head, mutation-tested the key gap, traced judge payloads through `TraceExtractor`, and ran a battery of fresh adversarial probes against the new helper code (the independent subagent pass hit its harness time limit, but its probe scripts survived and I executed all of them myself — findings folded in below).

**Verdict: no blockers remain — both round-1 🔴s are fixed and verified.** Three 🟡 should-fixes before un-drafting:
1. **Pin `bridge_parent_gaps`** — deleting the call leaves the whole suite green while `multiply_numbers` silently flips ownership to the coordinator (verified test supplied).
2. **Image-first multimodal input still drops the agent span** — `_get_last_message_text` reads only part 0; a `[image, text]` user message has its text at part 1.
3. **A unicode-digit attribute key crashes the whole session** — `seg.isdigit()` admits `'²'`/`'①'` but `int(seg)` raises, and the `ValueError` escapes `map_to_session`. One-word guard; and it means the un-wrapped normalizer call has a live trigger again.

| Round-1 finding | Status at `1c9587f` |
|---|---|
| 🔴 1 — hard-coded attribute paths (5 manifestations) | ✅ **fixed** — multi-turn ✓, reasoning ✓, multimodal (text part 0) ✓, chat-completions indexing ✓, tool-call-only turn ✓ (`[delegated]` fallback) |
| 🔴 2 — handoff collapse + misattribution | ✅ **fixed** — coordinator + specialist both survive; handoff owned by the coordinator; per-agent tools; judge payloads verified end-to-end |
| 🟡 3 — normalizer crash escaping `map_to_session` | ✅ both round-1 triggers fixed — but the probes found a new one (🟡 3 below), so the wrap suggestion is back on the table |
| 🟡 4 — test gaps | ◐ **mostly fixed** — 10 new tests pin the helpers/counts/crash cases; `bridge_parent_gaps` remains unpinned |
| ⚪ hunk-G non-dict-schema regression | ✅ **fixed properly** — unwrap moved into a scope-gated per-span normalizer with an `isinstance` check |

Suites: mapper **389 passed** (was 385) · extractors/types/detectors **235 passed**.

Your three answers, closed out: **coordinator exists / wrapper doesn't** — implemented exactly that way and verified in both fixtures. **Ordering with #365** — noted, no dependency. **No integ tests yet** — matches the #340 → #353 precedent; a tracking issue would close the loop but isn't blocking.

<details>
<summary><b>Fix verification — round-1 repros re-run at <code>1c9587f</code>, verbatim</b></summary>

```
multi-turn:        user_prompt='Now divide that by 5.'  response='555 divided by 5 is 111.'  (was: turn-1's question)
reasoning @ idx 0: response='REAL ANSWER'                                                    (was: the scratch-pad)
multimodal input:  span survives, prompt='MULTIMODAL Q'                                      (was: dropped)
chat-completions:  input numbered from 0 → prompt='CC QUESTION'                              (was: wrong index)
tool-call-only:    response='[delegated] transfer_to_x({})'                                  (was: dropped)
het. start_time:   no crash; healthy LangChain trace in the same batch survives              (was: TypeError killed the session)
attributes=None:   no crash; degrades to a per-span warning                                  (was: AttributeError killed the session)
non-dict schema:   tools=[] and the span survives                                            (was: whole span dropped)
OpenAI envelope:   ('calc', description, parameters) intact via the new per-span normalizer
```

Residual from manifestation (e): an **empty-string** final answer still drops the span (`if agent_response:` + the gate). The reachable case — tool-call-only — is fixed; the empty-string one is rare, ⚪.

The new unit tests (`test_get_last_message_text_uses_current_turn_user_message`, `test_get_last_message_text_skips_reasoning_item`, `test_normalization_survives_heterogeneous_start_times`, `test_normalization_survives_none_attributes_span`) pin the helper behaviours — these were among the 10 surviving mutants last round, so that's the right set to have added.

</details>

<details>
<summary><b>Coordinator survival — verified end-to-end, including what the judge sees</b></summary>

ADOT fixture: 3 raw AGENT spans → **2** `AgentInvocationSpan`s (wrapper dropped, per your intent):

```
coordinator:      response='[delegated] transfer_to_math_specialist(); transfer_to_research_specialist()'
                  tools=['transfer_to_math_specialist', 'transfer_to_research_specialist']
math_specialist:  response='The product of 42×7 is 294. As for the weather…'
                  tools=['add_numbers', 'divide_numbers', 'multiply_numbers']
TOOL 'handoff to math_specialist' → agent_span_id = coordinator      ✓
TOOL 'multiply_numbers'           → agent_span_id = math_specialist  ✓
```

TOOL_LEVEL judge payload — this closes the false-positive `poshinchen` reported (handoff judged against the wrong candidate set):

```
target: handoff to math_specialist  available_tools=['transfer_to_math_specialist', 'transfer_to_research_specialist']
target: multiply_numbers            available_tools=['add_numbers', 'divide_numbers', 'multiply_numbers']
```

The recaptured live fixture (agents-as-tools, 19 spans) also maps cleanly: wrapper dropped, coordinator (with `ask_*` tools) + specialist both survive, and the specialist's prompt is the actual tool-call argument it received.

One inherent oddity, flagged so nobody rediscovers it as a bug: both agents carry the same `user_prompt` (the SDK threads the full conversation into the handed-off agent), so the specialist's TRACE_LEVEL history shows the user question twice. Faithful to what the specialist actually received — ⚪.

</details>

<details>
<summary><b>🟡 1 — pin <code>bridge_parent_gaps</code> (verified test supplied)</b></summary>

With the call at `openinference_session_mapper.py:423` replaced by `pass`, the mapper file's **127 tests all pass** — and `multiply_numbers.agent_span_id` flips from `math_specialist` to `coordinator` via the orphan fallback (`types/trace.py`). So the suite stays green while tool attribution goes wrong in exactly the way this PR fixes. `test_handoff_tool_owned_by_coordinator` doesn't catch it: the handoff lands on the coordinator through the root-fallback whether or not the bridge runs.

This test passes with the call and fails without it (verified both ways):

```python
def test_multiply_numbers_owned_by_math_specialist(self, openai_agents_adot_session):
    """Pins bridge_parent_gaps: without parent-gap bridging, the orphan fallback
    hands multiply_numbers to the coordinator instead of the specialist that ran it."""
    all_spans = [s for t in openai_agents_adot_session.traces for s in t.spans]
    specialist = next(
        s for s in all_spans
        if isinstance(s, AgentInvocationSpan) and "[delegated]" not in (s.agent_response or "")
    )
    multiply = next(
        s for s in all_spans if isinstance(s, ToolExecutionSpan) and s.tool_call.name == "multiply_numbers"
    )
    assert multiply.agent_span_id == specialist.span_info.span_id
    assert multiply.span_info.parent_span_id == specialist.span_info.span_id
```

</details>

<details>
<summary><b>🟡 2 & 🟡 3 — new findings from the adversarial probes (both reproduced end-to-end)</b></summary>

**🟡 2 — image-first multimodal user message drops the agent span.** `_get_last_message_text` reads `contents.0.message_content.text` only. The instrumentor emits multimodal parts in order, so a user message shaped `[image, text]` has `contents.0.type = "image"` and its text at `contents.1` — the helper returns `None`, `input.value` is never injected, and the AGENT span is dropped (reproduced: text-first `[text, image]` works; image-first yields `(no traces)`). Same reachability as round-1's 1c — the documented way to send images — which this fix only half-covered. Fix: scan the message's `contents.N` parts for the first `message_content.type == "text"` instead of hardcoding part 0. (Related ⚪, same loop: a multi-part text output returns only part 0, truncating the response.)

**🟡 3 — `_get_message_indices` crashes the whole session on a unicode-digit key.** `seg.isdigit()` is true for `'²'` and `'①'`, but `int(seg)` raises `ValueError: invalid literal for int()` — and both call paths (the per-span tools normalizer and the trace normalizer) sit outside any `try/except`, so one malformed key kills the entire `map_to_session` call, other producers' traces included (reproduced end-to-end on both paths). No real instrumentor emits such keys — this is corrupt/adversarial data — but the blast radius is total and the fix is one word: `if seg.isascii() and seg.isdigit()`. It also retires my appendix note that the un-wrapped normalizer call had “no known trigger”: it has one now, so the round-1 `try/except logger.warning` wrap around the normalizer calls is worth adding after all.

</details>

<details>
<summary><b>Appendix — non-blocking (9)</b></summary>

1. ⚪ **Stale provenance comment**: the live-fixture header (test file ~:2517-2519) still says “single-agent … v1.6.1”, but the recaptured fixture is multi-agent (19 spans, agents-as-tools) and its outputs use the `contents.N` paths (v2.0.0 shape). Worth updating — the provenance comment is what the next maintainer trusts.
2. ⚪ The descendant walk still has no `visited` set — a duplicate `span_id` closing a cycle still hangs (contrived input; one-line fix if you want it).
3. ⚪ The tools-copy loop keeps the contiguous-index assumption (probe-confirmed: a gap at `llm.tools.1` copies only `tool_0` onto the AGENT span, which then *shadows* the fuller trace-wide fallback) and copies from the earliest LLM span only — real instrumentor output is contiguous, so ⚪.
4. ⚪ Same contiguity in `_get_last_tool_calls`: tool_calls at indices {0, 2} render only index 0; a nameless call at 0 hides a named one at 1. Not real instrumentor output.
5. ⚪ `[delegated]` synthesis embeds raw argument JSON unbounded — a 200KB tool argument becomes a 200KB `agent_response` in the judge prompt. A length cap would be kind to judge budgets.
6. ⚪ An empty-string *last* message now silently returns the **previous** message's text (input and output alike) rather than dropping — arguably worse than a drop when it happens, but the trigger (empty final content alongside earlier text in the same span) is contrived.
7. ⚪ A message carrying reasoning at `contents.0` and its real text at `contents.1` is skipped entirely — v2.0.0 emits reasoning as its *own* message, so not real output today; the part-scan fix for 🟡 2 would cover this too.
8. ⚪ A **dict-valued** (non-string) wrapped tool schema now loses the tool (`json.loads(dict)` → `TypeError` → skip, and the extractor no longer unwraps) — OTEL attribute values are strings, so only hand-built input hits it.
9. ⚪ The walk still doesn't scope-check spans in mixed-scope trace groups — needs a non-openai AGENT *ancestor* of openai spans to matter; contrived.

</details>

<sub>Follow-up review by `strandly-the-agent`: self-run re-verification of all prior findings per follow-up protocol (every repro above executed at `1c9587f`), plus adversarial probes of the new helper code — the independent subagent pass timed out at its harness limit, so I recovered its probe scripts and executed the full battery myself; a full mutation sweep of the new helpers didn't complete, though the new unit tests pin the behaviours that survived mutation last round. Solid work on the fixes — the remaining 🟡s are a test and two small guards, not redesign.</sub>

liramon2:
@strandly-the-agent Review this PR. Consolidate findings into a single comment.

strandly-the-agent:
Re-reviewed at `c304ade` (fixes-only delta, +79/−38; self-run follow-up, every check below executed by me at this head).

**All three 🟡s from my last review are fixed and verified — nothing left that I'd block or should-fix.** From my side this is ready to come out of draft.

| Finding | Status at `c304ade` |
|---|---|
| 🟡 1 — pin `bridge_parent_gaps` | ✅ test adopted (`test_multiply_numbers_owned_by_math_specialist`) — re-ran the mutation: deleting the call now **fails 1 test** (was 0) |
| 🟡 2 — image-first multimodal drop | ✅ `_join_content_text` scans all parts — `[image, text]` now yields the prompt; bonus: multi-part text outputs are joined (`'Part A. Part B.'`), retiring an appendix item |
| 🟡 3 — unicode-digit key crash | ✅ `isascii() and isdigit()` guard **and** both normalizer paths wrapped in `try/except logger.warning` — the round-1 wrap suggestion, now in; a poisoned key degrades to a warning and every trace survives |

Also cleaned up from the appendix, verified: cycle guard (`seen` set — the duplicate-`span_id` cycle now terminates), gap-tolerant tools copy (`{0, 2}` → both tools), dict-valued wrapped schemas handled, stale fixture header comments replaced.

Regression spot-checks all green: multi-turn + reasoning still resolve correctly, ADOT fixture still yields 2 agents with per-agent tools, mapper suite **390 passed** · downstream **235 passed**.

Still-open ⚪ (all contrived inputs, previously disclosed, fine to leave): empty-string final answer drops the span; an empty-string *last* message falls back to the previous message's text; a same-message `[reasoning, text]` shape is skipped (`is_reasoning` checks part 0 only — not real v2.0.0 output); `[delegated]` embeds unbounded argument JSON; `_get_last_tool_calls` index contiguity; no scope check in mixed-scope trace groups.

<sub>Follow-up review by `strandly-the-agent` — fixes-only delta, so per protocol this was a self-run verification pass (mutation re-run, repro battery re-run, fixtures end-to-end), not a full pipeline. Nice, thorough round of fixes.</sub>

ybdarrenwang (approved):
LGTM

liramon2:
@strandly-the-agent Review this PR. The new commit fixes the `bridge_parent_gaps` handling for mixed scope traces.

strandly-the-agent:
Re-reviewed at `dbbbd53` (delta since my last review: `656f223` comment polish + `dbbbd53` the mixed-scope `bridge_parent_gaps` fix; self-run follow-up, everything below executed at this head).

**The fix is correct and verified — no blockers.** One 🟡: it ships with **no test**, and reverting it leaves all 390 tests green. Verified pin test below.

What I checked:

- ✅ **Fix confirmed real.** Synthetic mixed-scope chain `AGENT → httpx POST (foreign scope) → TOOL`: at `c304ade` the TOOL came out `parent_span_id=None` (the scope-filtered map couldn't walk through the foreign span); at `dbbbd53` it bridges to `parent='agent-1'`, `agent_span_id='agent-1'`. This is the same class of gap I flagged on #375's `GenericGenAISessionMapper` — good to see it closed here too.
- ✅ **Behaviour-neutral for the 3 existing producers.** Full span-tuple diff (`span_id`, `parent_span_id`, `agent_span_id`) across all 5 pre-existing fixtures (langchain live/adot, smolagents, claude live/adot), old vs new: **0 changed spans**. The all-scopes map and the call-site move change nothing on shipped data.
- ✅ **The call-site move (bridge now runs *after* the LangChain dedup) is sound** — neutral on the fixtures, and strictly better in theory: children of a deduped-away agent span now bridge through it to the survivor instead of dangling at a removed span id.
- ✅ Suites: mapper **390 passed** · extractors/types/detectors **235 passed**. Quick regression spot-check on both openai_agents fixtures: agent counts and tool ownership unchanged.
- ✅ `_raw_parent_map` instance state is rebuilt at the top of every `map_to_session` call — no staleness across calls.

<details>
<summary><b>🟡 pin the fix — reverting to the scope-filtered map keeps all 390 tests green (test verified both ways)</b></summary>

This is the third bridge-related behaviour in this PR's lineage to arrive unpinned — same story as the `bridge_parent_gaps` call itself two rounds ago. This test passes at `dbbbd53` and fails with the fix reverted:

```python
def test_bridge_crosses_foreign_scope_spans(self):
    """A converted span whose raw ancestry passes through a non-OpenInference span
    (e.g. an httpx POST) still bridges to its converted ancestor — pins the
    all-scopes raw parent map built in map_to_session."""
    agent = make_span(
        trace_id="t1", span_id="agent-1", name="math_agent",
        scope_name=OPENAI_AGENTS_SCOPE_NAME,
        attributes={"openinference.span.kind": "AGENT", "input.value": "q", "output.value": "a"},
    )
    http = make_span(
        trace_id="t1", span_id="http-1", parent_span_id="agent-1", name="POST",
        scope_name="opentelemetry.instrumentation.httpx", attributes={},
    )
    tool = make_span(
        trace_id="t1", span_id="tool-1", parent_span_id="http-1", name="calculator",
        scope_name=OPENAI_AGENTS_SCOPE_NAME,
        attributes={"openinference.span.kind": "TOOL", "tool.name": "calculator",
                    "input.value": "{}", "output.value": "42"},
    )
    session = self.mapper.map_to_session([agent, http, tool], "sess-1")
    tool_spans = [s for t in session.traces for s in t.spans if isinstance(s, ToolExecutionSpan)]
    assert len(tool_spans) == 1
    assert tool_spans[0].span_info.parent_span_id == "agent-1"
    assert tool_spans[0].agent_span_id == "agent-1"
```

</details>

<sub>Follow-up review by `strandly-the-agent` — fixes-only delta, self-run per protocol: A/B'd the fix against the previous head, span-tuple-diffed all five existing fixtures, mutation-tested the new map, re-ran the suites. Note `ybdarrenwang`'s approval predates this commit, so this comment covers the delta since.</sub>

liramon2:
@strandly-the-agent Review this PR (resolved merge conflicts + added test since last review). Consolidate findings into 1 comment.

strandly-the-agent:
Re-reviewed at `7de7eb9` (rebase onto `dce1b8a` + `ccc3086` “capture result from bridge_parent_gaps” + `7de7eb9` the bridge test). Self-run follow-up; everything below executed at this head.

**All clear — the round-4 🟡 is closed, the rebase is drift-free, and I have no findings.**

- ✅ **Round-4 🟡 (unpinned bridge fix) — closed.** `test_bridge_crosses_foreign_scope_spans` adopted; I re-ran the mutation and it now fails 1 test (was 0).
- ✅ **Rebase fidelity verified.** Compared the accumulated source diff before (`1f752dc..dbbbd53`) and after (`dce1b8a..7de7eb9`): the **only** content difference is the `ccc3086` reassignment line — zero conflict-resolution drift.
- ✅ **`ccc3086` is correct and necessary.** Main's #375 changed `bridge_parent_gaps` to return patched copies instead of mutating in place, so after the rebase the old bare call would have made bridging a **silent no-op** — nastier than a conflict, since it compiles and only shows up as wrong tool attribution. Verified by mutation: dropping the reassignment fails **2** tests (the new bridge test + `test_multiply_numbers_owned_by_math_specialist`), so this exact hazard is now double-pinned. Also checked the copy semantics: `available_tools`/`system_prompt` are attached before the bridge and `model_copy` carries them, so nothing is lost on the patched spans.
- ✅ Suites at the new baseline (which now includes #375, #371, #330): mapper **398 passed** · extractors/types/detectors **368 passed**.

<sub>Follow-up review by `strandly-the-agent` — rebase + fixes-only delta, self-run per protocol: accumulated-diff comparison for rebase drift, both mutations re-run, suites green at the new base. Nothing outstanding from my side across all five rounds.</sub>

src/strands_evals/mappers/openinference_session_mapper.py
```diff
@@ -67,12 +69,13 @@ class OpenInferenceSessionMapper(SessionMapper):
     - openinference-instrumentation-langchain (LangChain / LangGraph)
     - openinference-instrumentation-smolagents (HuggingFace smolagents)
     - openinference-instrumentation-claude-agent-sdk (Claude Agent SDK)
+    - openinference-instrumentation-openai-agents (OpenAI Agents SDK)
```
JackYPCOnline, line 72: how do we pick up this naming? Wondering `claude-agent-sdk` but for openai agent sdk we only do `openai-agents`

liramon2: These are the official package names from the OpenInference repository: https://github.com/Arize-ai/openinference/tree/main

JackYPCOnline: Thanks!

src/strands_evals/mappers/openinference_session_mapper.py
```diff
@@ -258,6 +281,167 @@ def _normalize_smolagents_tool_attrs(self, attrs: dict) -> None:
...
+                    llm_spans.append(child)
+                elif kind == "CHAIN":
+                    cid = child.get("span_id", "")
+                    if cid:
+                        stack.append(cid)
+        llm_spans.sort(key=lambda s: self.parse_timestamp(s.get("start_time")))
+        return llm_spans
+
+    @staticmethod
+    def _get_message_indices(attrs: dict, prefix: str) -> list[int]:
```
JackYPCOnline: Nit: overall, can we make function naming more accurate

src/strands_evals/mappers/openinference_session_mapper.py
```diff
@@ -258,6 +281,167 @@ def _normalize_smolagents_tool_attrs(self, attrs: dict) -> None:
...
+        prefix_dot = f"{prefix}."
+        indices: set[int] = set()
+        for key in attrs:
+            if key.startswith(prefix_dot):
+                seg = key.removeprefix(prefix_dot).split(".", 1)[0]
+                if seg.isascii() and seg.isdigit():
+                    indices.add(int(seg))
+        return sorted(indices, reverse=True)
+
+    @classmethod
```
JackYPCOnline: Do these helpers need class-level invocation or subclass override behavior? Origin code is just some private helpers. 
if they are a must, can we make code more consistent?

liramon2: They don't need access to `self`, but then again other private helpers in the file don't access `self`. So to answer your question, they can just be private helpers. Done in 8bf879e525444a6c001e437abd30eafa06b307ac

src/strands_evals/mappers/openinference_session_mapper.py
liramon2: Done in 8bf879e525444a6c001e437abd30eafa06b307ac. Standardized the verbs from `get` to `extract`/`collect`.

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into 1 comment.

strandly-the-agent:
Re-reviewed at `02a76ed` (delta since `7de7eb9`: formatting · `cb15d94` foreign-scope checks · `8bf879e` helper rename/reorg · `02a76ed` “remove stale error check”). Self-run follow-up; everything below executed at this head.

**One 🟡 to flag: the “stale” error check wasn't stale.** `02a76ed` removes `ValueError` from the `except` tuple in `_extract_tools_from_attributes` — but it was catching pydantic's `ValidationError` (a `ValueError` subclass), and removing it resurrects a failure mode this PR deliberately fixed in round 1. Everything else in the delta checks out clean.

| Commit | Verdict |
|---|---|
| `cb15d94` foreign-scope checks | ✅ verified — closes the long-standing mixed-scope appendix item, both directions |
| `8bf879e` rename/reorg | ✅ behaviour-neutral — span-tuple diff across all **7** fixtures vs `7de7eb9`: **0 changes**; helper bodies moved verbatim |
| `02a76ed` remove `ValueError` | 🟡 regression — A/B repro below; one-word restore |
| Suites | ✅ mapper **399 passed** · extractors/types/detectors **368 passed** |

<details>
<summary><b>🟡 the removed <code>ValueError</code> was catching pydantic <code>ValidationError</code> — a bad tool schema now drops the whole span again (A/B verified)</b></summary>

`ToolConfig.name` is a required `str` (`types/trace.py:52`), so a tool schema without a usable `name` — e.g. an OpenAI envelope whose inner `function` dict lacks one, or `"name": 123` — makes `ToolConfig(name=None, …)` raise `pydantic.ValidationError`, which **is a `ValueError` subclass**. That's what the removed clause was catching (the `TypeError` scalar-schema case it once also shielded was properly fixed in the per-span normalizer two rounds ago — that half really was stale; this half wasn't).

Same input, previous head vs this head — an AGENT span carrying one nameless schema and one good one:

```
at 7de7eb9:  AgentInvocationSpan count = 1   tools=['good_tool']        # bad tool skipped
at 02a76ed:  AgentInvocationSpan count = 0   (WHOLE SPAN DROPPED)
             "Failed to convert span ag: 1 validation error for ToolConfig … name … input_value=None"
```

The exception escapes to `_build_trace`'s broad per-span `except Exception`, so one malformed tool schema silently discards the entire agent invocation — including the valid tools next to it. This is byte-for-byte the round-1 finding that motivated adding `ValueError` (my round-1 comment noted it was *not* redundant for exactly this case; the round-2 probes confirmed “that's the `ValueError` catch working”).

Reachability is contrived-data territory (real instrumentors emit named tools), so 🟡 not 🔴 — but it's a silent regression of this PR's own earlier behaviour. Fix: restore `ValueError` to the tuple, or skip schemas where `not isinstance(tool_info.get("name"), str)` before constructing `ToolConfig`. A pin test is cheap: the repro above as a unit test asserting `tools == ['good_tool']` and span survival.

</details>

<details>
<summary><b>✅ cb15d94 — foreign-scope checks verified in both directions</b></summary>

- A smolagents-scoped AGENT span that is an *ancestor* of openai_agents CHAIN/LLM spans in a shared trace is no longer normalized (attrs untouched after `map_to_session` — previously it got openai content injected). 
- A foreign-scope LLM *descendant* (e.g. langchain) is no longer harvested for an openai agent's prompt/response — an openai AGENT with only foreign LLM descendants is now dropped rather than borrowing foreign content.
- The new `test_foreign_scope_llm_not_used_for_agent_normalization` pins the second direction. Both openai fixtures: span-tuple diff vs `7de7eb9` = 0 changes.

</details>

<sub>Follow-up review by `strandly-the-agent` — self-run per protocol: A/B'd the delta against the previous head, span-tuple-diffed all 7 fixtures, re-ran the suites. The 🟡 is a one-word restore plus a small pin test.</sub>

JackYPCOnline:
@strandly-the-agent can you check this PR is ready for approval?

strandly-the-agent:
@JackYPCOnline **Yes — ready for approval from my side.** Checked at head `8756547` (the revert of `02a76ed`).

| Readiness check | Status |
|---|---|
| Round-6 🟡 (removed `ValueError` catch → whole-span drop) | ✅ **fixed** — `8756547` reverts it; head tree is byte-identical to the pre-removal tree (`git diff 8756547 02a76ed^` = empty), and my repro now yields span-survives + `tools=['good_tool']` instead of a dropped span |
| All earlier findings (rounds 1–5) | ✅ closed — both round-1 blockers, the handoff/coordinator design, the crash paths, and every bridge behaviour are fixed and test-pinned; remaining ⚪ appendix items are contrived-input hardening, none blocking |
| Local suites at head | ✅ mapper **399 passed** · extractors/types/detectors **368 passed** |
| CI at head | ✅ full matrix green (unit 3.10–3.14 × linux/windows/macOS, lint, CI Gate, **integration tests**) |
| Review threads | ✅ all 3 resolved; branch is MERGEABLE against `dce1b8a` |

Two notes for the approver:

1. **Process**: `reviewDecision` is `REVIEW_REQUIRED` — `ybdarrenwang`'s approval (Aug 18) predates the last five commits, so a fresh approval (yours) is what's needed to unblock.
2. ⚪ **Worth a fast-follow, not blocking**: the integration tests **failed at `02a76ed`** (the `ValueError` removal) and pass at the revert — consistent with that guard being reachable on live traces, not just crafted data (though I haven't ruled out integ flakiness). The restored guard still has no unit pin test; the one from my round-6 comment (nameless schema → `tools == ['good_tool']`, span survives) would keep the next “remove stale check” cleanup from repeating this.

<sub>Readiness check by `strandly-the-agent`: revert fidelity verified by tree diff, round-6 repro re-run at head, both suites re-run, CI rollup and thread state read from the API. Six review rounds on this PR are now fully closed out.</sub>

JackYPCOnline (approved):

PR #371, opened 2026-08-15
fix: select root agent span by earliest start_time in multi-agent traces

### LINKED ISSUE #370: _find_root_agent_span selects an arbitrary sub-agent as root in flat multi-agent traces
### Summary

`_find_root_agent_span` picks the wrong agent as the trace root in flat multi-agent traces. When every `AgentInvocationSpan` is parentless, it returns the first parentless-with-content span *in list order*, which is an arbitrary sub-agent rather than the outer coordinator. `TraceExtractor._extract_tool_level` seeds its conversation history from `root.user_prompt`, so choosing a sub-agent as root silently narrows the "user request" every tool call is judged against. `ToolSelectionAccuracyEvaluator` then returns false negatives with no warning.

### Steps to Reproduce

Run the following script:
```
from datetime import datetime, timezone

from strands_evals.evaluators.tool_selection_accuracy_evaluator import ToolSelectionAccuracyEvaluator
from strands_evals.types.evaluation import EvaluationData
from strands_evals.types.trace import (
    AgentInvocationSpan,
    Session,
    SpanInfo,
    ToolCall,
    ToolConfig,
    ToolExecutionSpan,
    ToolResult,
    Trace,
)
from strands_evals.extractors.trace_extractor import TraceExtractor
from strands_evals.types.trace import EvaluationLevel


T0 = datetime(2025, 1, 1, tzinfo=timezone.utc)
info = lambda span_id, start_sec, end_sec: SpanInfo(
    session_id="s", trace_id="t", span_id=span_id, parent_span_id=None,
    start_time=T0.replace(second=start_sec), end_time=T0.replace(second=end_sec),
)

sub_agent = AgentInvocationSpan(
    span_info=info("sub", 5, 10),
    user_prompt="Weather in London",
    agent_response="50°F",
    available_tools=[ToolConfig(name="get_weather")],
)
coordinator = AgentInvocationSpan(
    span_info=info("coord", 0, 20),
    user_prompt="Weather in New York and London, then calculate the difference.",
    agent_response="18°F difference.",
    available_tools=[ToolConfig(name="get_weather"), ToolConfig(name="calculate")],
)
tool_ny = ToolExecutionSpan(
    span_info=info("tool-ny", 1, 2),
    tool_call=ToolCall(name="get_weather", arguments={"city": "New York"}),
    tool_result=ToolResult(content="68°F"),
)
tool_london = ToolExecutionSpan(
    span_info=info("tool-london", 6, 7),
    tool_call=ToolCall(name="get_weather", arguments={"city": "London"}),
    tool_result=ToolResult(content="50°F"),
)
tool_calc = ToolExecutionSpan(
    span_info=info("tool-calc", 11, 12),
    tool_call=ToolCall(name="calculate", arguments={"expr": "68 - 50"}),
    tool_result=ToolResult(content="18"),
)

session = Session(
    session_id="s",
    traces=[Trace(
        trace_id="t", session_id="s",
        spans=[sub_agent, coordinator, tool_ny, tool_london, tool_calc],
    )],
)

tool_inputs = TraceExtractor(EvaluationLevel.TOOL_LEVEL).extract(session)
for ti in tool_inputs:
    user_msg = ti.session_history[0].content[0].text
    tool_name = ti.tool_execution_details.tool_call.name
    print(f"{tool_name}({ti.tool_execution_details.tool_call.arguments}) → user_prompt: {user_msg!r}")

results = ToolSelectionAccuracyEvaluator().evaluate(EvaluationData(
    input="Weather in New York and London, then calculate the difference.",
    actual_output="18°F difference.",
    actual_trajectory=session,
))

failing = [r for r in results if not r.test_pass]
print(f"\n{len(failing)} failed out of {len(results)}")
```

### Expected Behavior

The output is:
```
get_weather({'city': 'New York'}) → user_prompt: 'Weather in New York and London, then calculate the difference.'
get_weather({'city': 'London'}) → user_prompt: 'Weather in New York and London, then calculate the difference.'
calculate({'expr': '68 - 50'}) → user_prompt: 'Weather in New York and London, then calculate the difference.'

0 failed out of 3
```
The root agent should be the outermost coordinator (earliest start_time among parentless spans), so tool-level evaluation is anchored to the actual user request. All three tool calls (`get_weather(New York)`, `get_weather(London)`, `calculate(68-50)`) are
justified against the full prompt "Weather in New York and London, then calculate the difference" and should score 1.0.

### Actual Behavior

The output is:
```
get_weather({'city': 'New York'}) → user_prompt: 'Weather in London'
get_weather({'city': 'London'}) → user_prompt: 'Weather in London'
calculate({'expr': '68 - 50'}) → user_prompt: 'Weather in London'

2 failed out of 3
```

When all `AgentInvocationSpans` are parentless (flat multi-agent trace), `_find_root_agent_span` returns the first span with content in list order, which is an arbitrary sub-agent. `_extract_tool_level` then seeds every tool call's session_history with that sub-agent's narrow delegated prompt (e.g. "Weather in London"), so the LLM judge in `ToolSelectionAccuracyEvaluator` scores tool calls that serve the real top-level request (get_weather for New York, calculate the difference) as unjustified. This causes false negatives.

### Possible Solution

Break ties by earliest `start_time` in every selection tier so the outermost orchestrator (invoked first, enclosing its sub-agents) is chosen:

```python
def _find_root_agent_span(agent_spans):
    if not agent_spans:
        return None
    parentless = [s for s in agent_spans if s.span_info.parent_span_id is None]
    with_content = [s for s in parentless if s.user_prompt or s.agent_response]
    candidates = with_content or parentless or list(agent_spans)
    return min(candidates, key=lambda s: _to_aware_utc(s.span_info.start_time))
```

liramon2:
## Description

Fixes root-agent selection in multi-agent traces so `_find_root_agent_span` breaks ties by earliest `start_time`. In a flat multi-agent trace where multiple agent spans are parentless, the previous logic returned the first parentless-with-content span *in list order*, which may be an arbitrary sub-agent rather than the outer coordinator. Since the coordinator is invoked first and encloses its sub-agents, selecting the earliest-start agent picks the root agent correctly.

## Related Issues

Fixes #370

Additionally, this is a prerequisite for https://github.com/strands-agents/evals/pull/365 and https://github.com/strands-agents/evals/pull/366. These PRs convert OpenAI Agents SDK traces that may contain multiple parentless span. When an OpenAI agent hands off or delegates to another agent, the original agent's span is a sibling rather than parent of the subsequent agent's span; therefore, there are multiple parentless agent spans. That leads the scenario described in this PR's issue.

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

## Checklist

### DIFF
--- a/src/strands_evals/types/trace.py
+++ b/src/strands_evals/types/trace.py
@@ -135,18 +135,16 @@ def _to_aware_utc(dt: datetime) -> datetime:
 def _find_root_agent_span(agent_spans: Sequence[AgentInvocationSpan]) -> AgentInvocationSpan | None:
     """Search for a trace's root agent span.
 
-    Prefers a parentless agent that has content, then any parentless agent,
-    then the earliest by `start_time`. Returns `None` for an empty sequence.
+    Prefers a parentless agent with content, then any parentless agent, then all
+    agents; within the chosen tier the earliest `start_time` wins.
     """
     if not agent_spans:
         return None
-    for span in agent_spans:
-        if span.span_info.parent_span_id is None and (span.user_prompt or span.agent_response):
-            return span
-    for span in agent_spans:
-        if span.span_info.parent_span_id is None:
-            return span
-    return min(agent_spans, key=lambda s: _to_aware_utc(s.span_info.start_time))
+
+    parentless = [s for s in agent_spans if s.span_info.parent_span_id is None]
+    with_content = [s for s in parentless if s.user_prompt or s.agent_response]
+    candidates = with_content or parentless or list(agent_spans)
+    return min(candidates, key=lambda s: _to_aware_utc(s.span_info.start_time))
 
 
 class Trace(BaseModel):
--- a/tests/strands_evals/extractors/test_trace_extractor.py
+++ b/tests/strands_evals/extractors/test_trace_extractor.py
@@ -453,3 +453,56 @@ def test_span_id_none_does_not_collide():
     result_b = next(r for r in result if r.tool_execution_details.tool_call.name == "tool_b")
     assert [t.name for t in result_a.available_tools] == ["tool_a"]
     assert [t.name for t in result_b.available_tools] == ["tool_b"]
+
+
+def test_tool_level_flat_multi_agent_anchors_to_earliest_coordinator():
+    """Flat trace, sub-agent first in list order: history and orphan tool anchor to the earliest coordinator."""
+    from datetime import timedelta
+
+    base = datetime(2026, 1, 1, tzinfo=timezone.utc)
+    # Sub-agent listed first, but starts 5s after the coordinator.
+    sub = AgentInvocationSpan(
+        span_info=SpanInfo(
+            session_id="test",
+            span_id="sub",
+            parent_span_id=None,
+            start_time=base + timedelta(seconds=5),
+            end_time=base + timedelta(seconds=6),
+        ),
+        user_prompt="current weather in London",
+        agent_response="rainy",
+        available_tools=[ToolConfig(name="get_weather")],
+    )
+    coordinator = AgentInvocationSpan(
+        span_info=SpanInfo(
+            session_id="test",
+            span_id="coordinator",
+            parent_span_id=None,
+            start_time=base,
+            end_time=base + timedelta(seconds=10),
+        ),
+        user_prompt="weather in NY and London, then the difference",
+        agent_response="here is the difference",
+        available_tools=[ToolConfig(name="ask_research"), ToolConfig(name="ask_math")],
+    )
+    # Orphan tool span: no parent and no agent_span_id, so it falls back to the root.
+    orphan_tool = ToolExecutionSpan(
+        span_info=SpanInfo(
+            session_id="test",
+            span_id="tool",
+            parent_span_id=None,
+            start_time=base + timedelta(seconds=5),
+            end_time=base + timedelta(seconds=6),
+        ),
+        tool_call=ToolCall(name="get_weather", arguments={"city": "London"}),
+        tool_result=ToolResult(content="rainy"),
+    )
+    trace = Trace(spans=[sub, coordinator, orphan_tool], trace_id="t1", session_id="test")
+    session = Session(traces=[trace], session_id="test")
+
+    result = TraceExtractor(EvaluationLevel.TOOL_LEVEL).extract(session)
+
+    assert len(result) == 1
+    assert result[0].session_history[0].content[0].text == "weather in NY and London, then the difference"
+    assert result[0].tool_execution_details.agent_span_id == "coordinator"
+    assert [t.name for t in result[0].available_tools] == ["ask_research", "ask_math"]
--- a/tests/strands_evals/types/test_trace.py
+++ b/tests/strands_evals/types/test_trace.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 from strands_evals.types.trace import (
     AgentInvocationSpan,
@@ -19,6 +19,7 @@ from strands_evals.types.trace import (
     Trace,
     TraceLevelInput,
     UserMessage,
+    _find_root_agent_span,
 )
 
 
@@ -348,3 +349,81 @@ def test_tools_without_span_ids_each_owned_by_their_own_agent():
 
     assert forecast.agent_span_id == "weather-agent"
     assert alerts.agent_span_id == "weather-agent"
+
+
+_ROOT_BASE = datetime(2026, 1, 1, tzinfo=timezone.utc)
+
+
+def _root_agent(
+    span_id: str, *, parent: str | None = None, offset: int = 0, prompt: str = "", response: str = ""
+) -> AgentInvocationSpan:
+    """Build an AgentInvocationSpan starting `offset` seconds after a base time."""
+    start = _ROOT_BASE + timedelta(seconds=offset)
+    return AgentInvocationSpan(
+        span_info=SpanInfo(span_id=span_id, session_id="s", parent_span_id=parent, start_time=start, end_time=start),
+        user_prompt=prompt,
+        agent_response=response,
+        available_tools=[],
+    )
+
+
+class TestFindRootAgentSpan:
+    def test_empty_returns_none(self):
+        """An empty sequence yields None."""
+        assert _find_root_agent_span([]) is None
+
+    def test_earliest_parentless_with_content_wins_regardless_of_order(self):
+        """Among parentless-with-content agents the earliest start wins, not list order.
+
+        The coordinator is listed last but starts first — the old list-order logic
+        would have returned the first sub-agent instead.
+        """
+        spans = [
+            _root_agent("sub_a", offset=2, prompt="current weather in London"),
+            _root_agent("sub_b", offset=2, prompt="current weather in New York"),
+            _root_agent("coordinator", offset=0, prompt="weather in NY and London, then the difference"),
+        ]
+        assert _find_root_agent_span(spans).span_info.span_id == "coordinator"
+
+    def test_prefers_content_over_earlier_empty_parentless(self):
+        """A content span beats an earlier empty one; both prompt and response count as content."""
+        by_prompt = [_root_agent("empty", offset=0), _root_agent("with_prompt", offset=1, prompt="do the thing")]
+        assert _find_root_agent_span(by_prompt).span_info.span_id == "with_prompt"
+
+        by_response = [_root_agent("empty", offset=0), _root_agent("with_response", offset=1, response="done")]
+        assert _find_root_agent_span(by_response).span_info.span_id == "with_response"
+
+    def test_falls_back_to_parentless_when_none_have_content(self):
+        """When no parentless agent has content, the earliest parentless wins over any parented span."""
+        spans = [
+            _root_agent("parented", parent="p", offset=0),
+            _root_agent("late", offset=5),
+            _root_agent("early", offset=1),
+        ]
+        assert _find_root_agent_span(spans).span_info.span_id == "early"
+
+    def test_mixed_naive_and_aware_start_times(self):
+        """Naive and aware start times are compared as UTC without raising TypeError."""
+        aware = _root_agent("aware", offset=0, prompt="a")
+        naive_start = datetime(2026, 1, 1, 0, 0, 5)
+        naive = AgentInvocationSpan(
+            span_info=SpanInfo(
+                span_id="naive",
+                session_id="s",
+                parent_span_id=None,
+                start_time=naive_start,
+                end_time=naive_start,
+            ),
+            user_prompt="b",
+            agent_response="",
+            available_tools=[],
+        )
+        assert _find_root_agent_span([aware, naive]).span_info.span_id == "aware"
+
+    def test_falls_back_to_earliest_when_all_parented(self):
+        """When every agent has a parent, the earliest-start agent overall is chosen."""
+        spans = [
+            _root_agent("late", parent="p", offset=5, prompt="x"),
+            _root_agent("early", parent="p", offset=1, prompt="y"),
+        ]
+        assert _find_root_agent_span(spans).span_info.span_id == "early"


liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.

strandly-the-agent:
**TL;DR — approve.** The fix does what #370 asks, verified end-to-end, and on equal `start_time`s it is behaviourally identical to the old code. Nothing blocks the merge. Five should-fix items below; the first two are the ones worth your read, because both change what the judge model actually sees.

**Verified** on `72efc62`, four independent passes (correctness, issue-alignment, test-quality, adversarial):
✅ `pytest tests/strands_evals/{types,extractors,mappers}` → **497 passed** · ✅ #370 repro fixed end-to-end (`session_history[0]` flips from `"current weather in London"` to the full coordinator request) · ✅ 2 of the 5 new tests fail on `origin/main`, so they do catch the regression · ✅ 4 differential repros re-run on head *and* `main`, plus 3 mutants confirmed by hand
⚠️ `ruff`/`mypy` are not installed in my sandbox, so lint and type-check were **not** run · ⚠️ the `langfuse`/`cli` failures are pre-existing and env-only (identical counts on `main`)

<details>
<summary><b>🟡 1. Root selection now depends on timestamp quality — two regressions vs <code>main</code>, reproduced</b> · <code>types/trace.py:146</code></summary>

Tiers 1–2 previously did **no** timestamp comparison — `start_time` was only consulted when nothing was parentless. Now `min()` runs on every path (`trace.py:146`), so root selection inherits whatever the ingestion path did to timestamps. `SessionMapper.parse_timestamp` (`mappers/session_mapper.py:77-111`) can hand it two kinds of untrustworthy value, and each one flips the root away from the coordinator:

| shape | `main` picks | this PR picks |
|---|---|---|
| coordinator `start_time` key absent, or unparseable (`"08/14/2026 12:00:00"`) → stamped `datetime.now(UTC)` (`session_mapper.py:93-94`, `:104-105`) | `coord` ✅ | `sub_r` ❌ |
| coordinator aware UTC, sub-agent naive local (`"2026-08-15T08:00:05"`, US/Eastern) → passed through unchanged (`:95-96`), then read as UTC by `_to_aware_utc` (`trace.py:128-132`) | `coord` ✅ | `sub_r` ❌ |

Both reproduced end-to-end through `GenericGenAISessionMapper` with the coordinator listed **first** in the export (invocation order) — precisely the shape the old list-order code got right. In the naive case the sub-agent genuinely starts 5s *after* the coordinator and still wins, because its wall-clock reads 4h earlier. Verified output, same script against both trees:

```
# PR head (72efc62)                     # origin/main (2127840)
root pick         = 'sub_r'             root pick         = 'coord'
orphan tool owner = 'sub_r'             orphan tool owner = 'coord'
history[0] = 'current weather in        history[0] = 'weather in NY and London,
              London'                                 then the difference'
```

The consequence is the same failure class as #370 itself: silently mis-anchored judge prompt, no error raised. Two caveats keeping this at 🟡 rather than blocking — it needs degraded timestamps *and* a flat multi-agent trace, and I could not find an in-repo fixture or a real captured export that produces the combination. Reach is the public, documented generic mapper (and `providers/langfuse_provider.py:252`, which passes `obs.start_time` straight through).

The helper can't defend itself here: a fabricated `now()` is indistinguishable from a real timestamp. The durable fix is at the boundary — make `parse_timestamp` honour its own docstring ("Returns: Timezone-aware datetime in UTC") by normalising naive values instead of returning them unchanged. That behaviour predates this PR, so it belongs in its own issue rather than piled onto this diff.

**What I'd ask for in *this* PR:** state the new precondition in the docstring (see #4) and pin the tz case with a test — precedent at `test_trace_extractor.py:302`. Deleting `_to_aware_utc` from `trace.py:146` leaves the entire suite green (137 passed), so the guard that stops mixed naive/aware from raising `TypeError` is itself untested on a path that only became reachable in this commit.
</details>

<details>
<summary><b>🟡 2. <code>available_tools</code> follows the root, so a flat trace still gets a wrong judge prompt</b> · <code>extractors/trace_extractor.py:138-140</code></summary>

`Trace.model_post_init` assigns tool spans it cannot attribute to the selected root (`trace.py:199-203`). `trace_extractor.py:138-140` then scopes that tool's `available_tools` to its owning agent — now the coordinator, whose tool list is delegate-only. So `get_weather` gets judged against a list that does not contain `get_weather`, rendered as `## Available tool-calls` at `evaluators/evaluator.py:199-200`.

Reproduced through `GenericGenAISessionMapper` — flat trace, sub-agent first in list order, coordinator earliest `start_time`, one orphan tool span:

```
PR head:  history[0]='weather in NY and London, then the difference'
          available_tools=['ask_research_specialist', 'ask_math_specialist']   # no get_weather
main:     history[0]='current weather in London'
          available_tools=['get_weather']
```

This is **not** a regression — `main` is wrong here too, just in the other field. The point is that in a fully-flat trace the fix *moves* the error rather than removing it: it trades a wrong prompt for a wrong tool list, and a tool call scored against a tool list that omits the tool is the same false-negative class #370 is about.

One line-number correction worth having, since it changes where a fix would go: the wrong list comes from the **per-agent scoping** branch at `:138-140`, *not* the `available_tools` fallback at `:112-113`/`:142`. With `agent_span_id` set and `len(agent_spans) > 1` the fallback is never reached, so a union built at `:112-113` alone would not fix this. Decoupling the tool list from root selection means touching `:138-140` — e.g. union the agent spans' tool lists when the owner came from the root fallback (a guess) rather than from the parent-chain walk — or not assigning a guessed owner to genuinely unattributable tool spans at all.

Entirely reasonable to declare this out of scope for a fix this size. If so, a test pinning the current shape would at least stop it being rediscovered as a bug later.
</details>

<details>
<summary><b>🟡 3. No test drives #370's scenario through the path it was reported on</b> · <code>tests/.../types/test_trace.py:368-405</code></summary>

All five new tests call the private `_find_root_agent_span` directly. The user-visible defect lives at `trace_extractor.py:110-115` and `trace.py:199-203`, and nothing pins it there: every multi-agent fixture in `test_trace_extractor.py` is **nested** (`:86` `parent_span_id="coordinator"`), never flat, and no pre-existing fixture uses differing `start_time`s — so a refactor of the call site could regress #370 with all five new tests green. #370's own repro is evaluator-level and names `_extract_tool_level` as the failure channel.

One fixture closes both call sites. Flat trace, sub-agent first in list order (as `strands_in_memory` emits, span-end order), coordinator earliest `start_time`, plus one orphan tool span:

```python
# tests/strands_evals/extractors/test_trace_extractor.py
result = TraceExtractor(EvaluationLevel.TOOL_LEVEL).extract(session)
assert result[0].session_history[0].content[0].text == "<top-level coordinator request>"
assert result[0].tool_execution_details.agent_span_id == "coordinator"
assert [t.name for t in result[0].available_tools] == ["<coordinator's tools>"]
```

I ran that shape through `_extract_tool_level` on both trees and all three values differ: on head they are the coordinator's, on `main` they are the sub-agent's (`'current weather in London'`, the sub-agent's span id, the sub-agent's tool list). So a test built this way genuinely fails on `main` — it isn't a tautology against head.
</details>

<details>
<summary><b>🟡 4. The docstring now under-describes what the function does</b> · <code>types/trace.py:136-139</code></summary>

The commit that made selection subtler also deleted the line that documented it:

```diff
-    Prefers a parentless agent that has content, then any parentless agent,
-    then the earliest by `start_time`. Returns `None` for an empty sequence.
+    Breaks ties by earliest `start_time`. Returns `None` for an empty sequence.
```

The tier order is still the dominant criterion: `with_content or parentless or list(agent_spans)` at `:145` picks the tier *before* `start_time` is consulted at `:146`. "Breaks ties" reads as though `start_time` were a last resort, when within the chosen tier it is the primary ordering — a reader would conclude the function returns the earliest agent span, which is wrong in exactly the case that matters here. The *why* (the outermost orchestrator is invoked first, so it starts earliest) isn't captured anywhere in the code, and it's the whole premise of the fix.

```python
    """Search for a trace's root agent span.

    Prefers a parentless agent with content, then any parentless agent, then all agents;
    within the chosen tier the earliest `start_time` wins, since the outermost
    orchestrator is invoked before the sub-agents it delegates to. Assumes comparable
    `start_time` values; naive ones are read as UTC. Returns `None` for an empty sequence.
    """
```
</details>

<details>
<summary><b>🟡 5. Three one-line mutations of the rewritten helper survive the whole suite</b> · <code>types/trace.py:144-146</code></summary>

I applied each mutation and re-ran `pytest tests/strands_evals/{types,extractors}` — **137 passed** every time:

| mutation | why no test sees it | fix |
|---|---|---|
| `:145` drop `or parentless` | `test_falls_back_to_parentless_when_none_have_content` (`:394`) has no parented span in its fixture | add `_root_agent("parented", parent="p", offset=0)` |
| `:144` drop `or s.agent_response` | no fixture has a span whose only content is a response | one extra span in `test_prefers_content_over_earlier_empty_parentless` (`:386`) |
| `:146` drop `_to_aware_utc` | nothing mixes naive and aware times here; mixed input then raises `TypeError` | covered by the test asked for in #1 |

6 of 9 mutants were killed by the suite (test-quality pass's count), so the new tests are doing real work — these three are the gaps.
</details>

### Questions

❓ **(blocking — a one-line answer closes it)** A coordinator whose content didn't extract still loses to a later sub-agent: the content tier is evaluated before `start_time`, so **#370's exact symptom still reproduces on the fixed code** when the orchestrator span's `user_prompt` *and* `agent_response` are both empty (coordinator at offset 0 loses to a sub-agent at offset 5). That shape is reachable — `strands_in_memory_session_mapper.py:413-414` defaults both to `""` and only fills them on a successful parse — and it's deliberately locked in by `test_trace.py:386-392`. Your own proposal in #370 chose this ordering, so I'd rather ask than override: is a content-less orchestrator span considered impossible in practice, or is that shape simply out of scope here? Either answer is fine — noting it in the docstring or PR body would settle it.

❓ (non-blocking) `langchain_otel_session_mapper.py:151` and `openinference_session_mapper.py:287` pick the root as `agent_spans[-1]`, documented as "the root graph finishes last" — the opposite heuristic to `trace.py:146`. Different provenance (LangGraph sub-graph nesting), but two contradictory root rules in one repo is a trap for the next reader. Worth a follow-up issue to reconcile or at least cross-reference them?

<details>
<summary><b>Appendix — non-blocking (5)</b></summary>

- ⚪ **Clock skew across hosts (contrived).** A sub-agent on a host 2s behind records `11:59:58.8Z` while the coordinator really started first at `12:00:00Z` → new code picks the skewed sub-agent, old code picked the coordinator. Needs a distributed flat trace with skew exceeding the delegation delay; noted only because it shares a root cause with #1.
- ⚪ **Equal-`start_time` behaviour is unpinned.** The "no change on ties" safety argument rests entirely on `min()` returning the first minimal element. A ~20k-case differential fuzz found ties byte-identical old vs new, so the argument holds — but nothing in the suite would catch it if this became `min(reversed(candidates), ...)`. One assertion, if first-in-list-on-tie is meant to be a contract.
- ⚪ `test_trace.py:357` — `_root_agent` leaves `span_id` and its four keyword params unannotated, while siblings `_info` (`:185`) and `_span_info` (`test_trace_extractor.py:22-26`) fully annotate. CI cannot catch it (`mypy -p src` only, `pyproject.toml:140`; ruff selects no `ANN` rules, `:146-153`), so this is consistency, not breakage.
- ⚪ Commit subject `fix: use start_time tiebraker for all root agent cases` — "tiebraker" → "tiebreaker". Moot if you squash-merge on the PR title.
- ⚪ Not verified by me: `ruff`/`mypy` aren't installed in my sandbox, so `hatch fmt --linter` and the type check are unrun on my side. CI covers both.
</details>

<details>
<summary><b>How this review was produced</b></summary>

Four passes ran independently against `72efc62`, then findings were de-duplicated and re-verified against the code before posting:

| pass | verdict | contributed |
|---|---|---|
| correctness | approve, no blockers | #1, #2, #4, and the blocking question |
| issue-alignment | in-contract but under-tested | #3, both questions; confirmed the diff at `trace.py:135-146` is byte-identical to #370's proposed solution |
| test-quality | gaps found | #3, #5, tie-behaviour and annotation nits |
| adversarial | 2 regressions + 1 contrived | #1's two repros, the clock-skew appendix item, ~20k-case differential fuzz |

Every 🔴 candidate was down-ranked to 🟡 on the reachability gate: the two regressions in #1 are real and reproduced end-to-end, but they require degraded timestamps in a flat multi-agent trace and no in-repo producer or captured export was found that emits that combination. Nine raw findings were dropped as duplicates, and I corrected one finding's `file:line` (the tool-list path in #2 is `:138-140`, not `:112-113`) — worth flagging since these came out of an agentic pass and line citations are exactly what drifts. Please treat the whole thing as work to review, not gospel.
</details>

liramon2:
@strandly-the-agent Review this PR. Consolidate any findings into a single comment.
> The durable fix is at the boundary — make parse_timestamp honour its own docstring ("Returns: Timezone-aware datetime in UTC") by normalising naive values instead of returning them unchanged.

This is effectively a no-op for the `_find_root_agent_span` because the timestamp is normalized in `_to_aware_utc`.  The conversion may happen earlier, but the operation is the same.

> One line-number correction worth having, since it changes where a fix would go: the wrong list comes from the per-agent scoping branch at :138-140, not the available_tools fallback at :112-113/:142. With agent_span_id set and len(agent_spans) > 1 the fallback is never reached, so a union built at :112-113 alone would not fix this. Decoupling the tool list from root selection means touching :138-140 — e.g. union the agent spans' tool lists when the owner came from the root fallback (a guess) rather than from the parent-chain walk — or not assigning a guessed owner to genuinely unattributable tool spans at all.

This is outside the scope of the PR, which is targeting the root agent search.

Question 1: a coordinator without content would most likely error on evaluations, so it is out of scope here.
Question 2: deferring to minimize changes to this PR.

strandly-the-agent:
**TL;DR — approve.** All five findings from my previous review are resolved on `9fa6675`, and I accept both of your rulings (one with a footnote below). No new findings — the `src/` delta is docstring-only, everything else is tests, so this was a fixes-only follow-up review rather than a full fresh-pass fan-out.

**Previous findings — status** (each verified against the diff, not the commit message):

| finding | status |
|---|---|
| 🟡 1 root selection depends on timestamp quality | ✅ in-PR asks delivered: naive/aware-mix test added (kills the `_to_aware_utc` mutant), docstring documents the tiers. Your no-op ruling: **agreed** for root selection — footnote below. |
| 🟡 2 `available_tools` follows the root | ✅ closed as **out of scope per your ruling** — and the new extractor test's third assertion (`['ask_research', 'ask_math']` for the orphan `get_weather`) pins the current shape, which was exactly the fallback I asked for. |
| 🟡 3 no test through the reported path | ✅ `test_tool_level_flat_multi_agent_anchors_to_earliest_coordinator` added — I ran it against `origin/main`'s code and it fails there, so it genuinely pins the regression. |
| 🟡 4 docstring under-describes | ✅ tier ordering restored. |
| 🟡 5 three surviving mutants | ✅ all killed — I re-applied each mutation (`drop or parentless`, `drop or s.agent_response`, `drop _to_aware_utc`) and each now fails exactly one test. |
| ❓ blocking question (contentless coordinator) | settled — the restored docstring documents the ordering, which was one of the two ways I said would close it. The shape remains by design; noted, not re-raised. |

**Verified on `9fa6675`:** ✅ `pytest tests/strands_evals/{types,extractors,mappers}` → **499 passed** locally · ✅ mutation re-runs as above · ✅ differential vs `main` for the new tests · ✅ CI **Lint** green (closes my earlier "ruff/mypy unverified" caveat) and the full unit-test matrix green · ⚠️ one CI note below.

⚠️ **`Run integration tests` failed on this head but passed on `72efc62`.** The delta between the two is a docstring in `src/` plus unit tests, which can't plausibly change live-model behaviour, and I can't read the job logs (no access from my side) — so it looks like judge/infra flakiness rather than this PR. Worth a re-run before merge to confirm: https://github.com/strands-agents/evals/actions/runs/31890316639/job/95025516391

<details>
<summary>Footnote on ruling 1 (parse_timestamp normalisation)</summary>

Agreed on the naive-datetime half: normalising naive → UTC in `parse_timestamp` is the same `replace(tzinfo=UTC)` operation `_to_aware_utc` already performs at the comparison site, so for `_find_root_agent_span` it's a no-op — the true offset of a naive value is unrecoverable either way. Conceded.

The half that argument doesn't cover is the **fabricated `datetime.now(UTC)`** for `None`/unparseable input (`session_mapper.py:93-94`, `:104-105`): no downstream guard can undo a fabricated timestamp, and that's what drove the missing/malformed-timestamp repro in my first review. That behaviour is pre-existing and out of this PR either way — mentioning it only as a candidate for a separate issue; happy to file it with the repro if you want, or drop it if you consider degraded exports out of contract.
</details>

mehtarac (approved):

strandly-the-agent:
Both answers received — nothing further from me. **Q1** (a contentless coordinator would error on evaluations anyway → out of scope): accepted. **Q2** (defer reconciling the `agent_spans[-1]` root heuristic to keep this PR small): accepted — happy to file the cross-reference follow-up issue later if it'd help.

The head is unchanged from `9fa6675`, which I already reviewed and approved, so there's nothing new to review. 🎉 on the merge.

PR #374, opened 2026-08-15
fix: reduce flakiness for integration tests targeting new session mappers

liramon2:
## Description

This PR reduces flakiness from the recent ADK and Claude Agents integration tests.

For Claude Agents:
- This adds an `expected_assertion` to the multi-agent case so the evaluator runs in assertion mode, which gives the judge an explicit success criterion and reduces sensitivity to phrasing differences in the agent's final response.
- This replaces the math specialist with a weather specialist to avoid the judge thinking the calculation task is trivial and does not warrant delegation.
- Soften the score needed to pass the multi-agent test to compensate for harsh judges

For ADK:
- Soften the score needed to pass the multi-tool selection test to compensate for harsh judges

## Type of Change

Bug fix

## Testing

- [x] The CI integration tests were ran 3 times in a row and each run succeeded.

## Checklist

----

### DIFF
--- a/tests_integ/test_adk_eval.py
+++ b/tests_integ/test_adk_eval.py
@@ -263,7 +263,8 @@ def test_adk_single_agent_multi_tool(telemetry, create_multi_tool_runner):
 
     # Exact count: 2 cases × 2 evaluators = 4
     assert len(report.scores) == 4
-    assert report.overall_score >= 0.75, f"Overall score too low ({report.overall_score}): {report.reasons}"
+    low_scores = sum(s < 0.5 for s in report.scores)
+    assert low_scores <= 1, f"Too many low scores ({low_scores}): {report.reasons}"
 
     # Verify mapped sessions have tool spans with tool_call_id populated
     for session in sessions:
--- a/tests_integ/test_adk_eval.py
+++ b/tests_integ/test_adk_eval.py
@@ -263,7 +263,7 @@ def test_adk_single_agent_multi_tool(telemetry, create_multi_tool_runner):
 
     # Exact count: 2 cases × 2 evaluators = 4
     assert len(report.scores) == 4
-    assert all(report.test_passes), f"Some evaluations failed: {report.reasons}"
+    assert report.overall_score >= 0.75, f"Overall score too low ({report.overall_score}): {report.reasons}"
 
     # Verify mapped sessions have tool spans with tool_call_id populated
     for session in sessions:
--- a/tests_integ/test_claude_openinference_eval.py
+++ b/tests_integ/test_claude_openinference_eval.py
@@ -200,6 +200,10 @@ def test_claude_multi_agent_evaluation(telemetry):
             name="multi-agent-math",
             input="Calculate the square root of 1764, then multiply that result by 3.",
             expected_output="126",
+            expected_assertion=(
+                "The agent computed the square root of 1764 (42), multiplied it by 3, "
+                "and responded with the final answer 126."
+            ),
         ),
     ]
 
--- a/tests_integ/test_claude_openinference_eval.py
+++ b/tests_integ/test_claude_openinference_eval.py
@@ -181,15 +181,19 @@ def test_claude_single_agent_evaluation(telemetry):
 def test_claude_multi_agent_evaluation(telemetry):
     """Multi-agent delegation + tool call spans evaluate correctly."""
     agents = {
-        "math-specialist": AgentDefinition(
+        "weather-specialist": AgentDefinition(
             description=(
-                "Math specialist agent. Delegate any mathematical calculations "
-                "including arithmetic, algebra, square roots, and numeric computations."
+                "Weather specialist agent. Delegate any weather lookups to this agent. "
+                "It has access to a get_weather Bash tool that returns real-time weather data "
+                "for a given city. Only this agent can retrieve weather information."
             ),
             prompt=(
-                "You are a math specialist. Use Bash ONLY to run `python3 -c 'import math; print(...)'` "
-                "for calculations. Never use Bash for anything else. "
-                "Be precise and return just the numeric result."
+                "You are a weather specialist. To look up weather, run EXACTLY this Bash command:\n"
+                '  bash -c \'python3 -c "data={\\"seattle\\": \\"Rainy, 55F\\", '
+                '\\"new york\\": \\"Sunny, 72F\\", \\"london\\": \\"Cloudy, 60F\\", '
+                '\\"tokyo\\": \\"Clear, 68F\\"}; city=\\"CITY_HERE\\".lower(); '
+                'print(data.get(city, \\"Mild, 65F\\"))"\'\n'
+                "Replace CITY_HERE with the requested city. Return the weather result verbatim."
             ),
             tools=["Bash"],
         ),
@@ -197,12 +201,12 @@ def test_claude_multi_agent_evaluation(telemetry):
 
     test_cases = [
         Case[str, str](
-            name="multi-agent-math",
-            input="Calculate the square root of 1764, then multiply that result by 3.",
-            expected_output="126",
+            name="multi-agent-weather",
+            input="What is the current weather in Seattle?",
+            expected_output="Rainy, 55F",
             expected_assertion=(
-                "The agent computed the square root of 1764 (42), multiplied it by 3, "
-                "and responded with the final answer 126."
+                "The agent delegated to the weather-specialist which looked up "
+                "the weather for Seattle and responded with Rainy, 55F."
             ),
         ),
     ]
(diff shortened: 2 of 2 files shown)


liramon2:
@strandly-the-agent Is this safe to merge?

strandly-the-agent:
**TL;DR — yes, safe to merge.** Tests-only change (`tests_integ/`, no production code), and on head `747945e` every check is green — including the live *Run integration tests* job. The only hard blocker left is human approval (`reviewDecision: REVIEW_REQUIRED`). Two 🟡 should-fixes below — neither breaks anything, but both affect whether the de-flaking actually holds, so fix-before-merge or fast-follow at your discretion.

🟡 **`tests_integ/test_adk_eval.py:266` — `overall_score >= 0.75` can green-light a real regression.** 4 binary rows (2 cases × 2 evaluators) mean `[1,1,1,0]` averages exactly 0.75 → pass. A mapper bug that garbles arguments only for the 3-param `convert_units` tool zeroes one row and sails through — and that's the regression class this test exists for. Tighter but still judge-tolerant: `assert sum(s < 0.5 for s in report.scores) <= 1` (also survives someone adding a 5th row later, which silently changes what 0.75 means).

🟡 **`tests_integ/test_claude_openinference_eval.py:205` — the weather swap parks tool-selection at exactly 0.5.** The judge never sees the specialist's system prompt, so it rates the hardcoded-dict Bash call "No" (fabricated data ≠ current weather) while `Agent(weather-specialist)` gets "Yes" → `scores[2]` = 0.5 with zero headroom; one extra shell call in a run (a retry, an `ls`) makes it 0.33 → flake. Cheap fix: word the case input as e.g. *"…according to the local mock weather dataset?"* — in our judge probes that flips the Bash verdict to Yes and restores the margin.

**Question (1 min to check):** in your 4 green runs, was `report.scores[2]` exactly `0.5`, with `reasons[2]` rejecting the Bash call as hardcoded data? That confirms or kills the second finding.

<details>
<summary>Verification ledger, per-pass breakdown, appendix (5 nits)</summary>

### Verified
- ✅ Reviewed head `747945e8519fe06a3fbec2d94134ec9f723d2bec`, up to date with `main`.
- ✅ CI on that head: lint + 11 unit jobs + CI Gate + **Run integration tests** all SUCCESS; only *Trigger Strands Review* still waiting (that's this review).
- ✅ Ran the embedded weather command locally with `CITY_HERE→Seattle`: quoting is correct through all three layers, prints exactly `Rainy, 55F`, matching `expected_output` and `expected_assertion`.
- ✅ Live model probes (22/22): the specialist emits the command verbatim and never shortcuts past Bash. *Caveat: probed with Sonnet 4.5, not the pinned Sonnet 4 (Bedrock now flags it Legacy — worth knowing for `DEFAULT_MODEL_ID`); judge prompts were reconstructed, not captured, so directions are solid but the exact numbers aren't gospel.*

### Test-quality pass
- The Claude "softening" is actually a **no-op**: with `expected_assertion` set, `GoalSuccessRateEvaluator` (assertion mode) and `CorrectnessEvaluator` (reference mode) are both binary 0/1, so `scores[i] >= 0.5` ⟺ `test_passes[i]`. Nothing got looser — worth knowing so nobody "restores strictness" or adds slack later. The trap: the equivalence holds only while `expected_assertion` is non-empty; drop it and Correctness falls back to 3-level mode where `PARTIALLY_CORRECT = 0.5` passes. Keeping `assert report.test_passes[1]` would be identical today and mode-proof tomorrow.
- The weather redesign is a genuine **strengthening**, not a bar-lowering: the weather dict lives only in the subagent's prompt, so the task is unsatisfiable without delegation — the old `sqrt(1764)*3` task was solvable by the orchestrator alone. Right shape of flakiness fix. The ADK half of the PR uses the opposite strategy (lower the bar), hence finding 1.
- `scores[0..2]` ordering is guaranteed today (results pre-allocated by case index; sub-reports concatenated in `evaluators=[...]` order) but only because there's exactly one case — add a second and the indices shift while the assertion messages keep their old names. `report.cases[i]["evaluator"]` is available for robust indexing.

### LLM-context pass
- Prompt text itself is sound; the risk is judge-side (finding 2 above). Old math case scored ~1.0 on tool selection in the same probe setup; `Agent(math-specialist)` was judged Yes 6/6, suggesting the original "delegation unwarranted" complaints came from the goal/correctness judges — which the `expected_assertion` addition already fixes on its own.

### Appendix — non-blocking (5)
- ⚪ Drop the `bash -c` nesting (`test_claude_openinference_eval.py:192-195`): the Bash tool already runs a shell; a plain single-quoted `python3 -c '…'` needs zero escaping, and one judge probe rejected the call partly because of the escape soup.
- ⚪ "run EXACTLY this Bash command" + "Replace CITY_HERE" is mildly self-contradictory; harmless in 22/22 trials.
- ⚪ `data.get(city, "Mild, 65F")` turns a key miss into a plausible wrong answer (rc=0), surfacing as "Correctness too low" instead of an obvious lookup failure.
- ⚪ `expected_output="Rainy, 55F"` (`:206`) is never read — no evaluator in this test consumes `expected_output`; same for the decorative fields in `test_adk_eval.py:237,243`.
- ⚪ Commit/PR wording says "soften thresholds" for the Claude test, but per the no-op note above nothing softened there — a one-line body edit would prevent a misguided future revert.

*Process note: full pipeline (triage → test-quality + LLM-context passes, both fresh-context; api/docs/adversarial passes routed out as irrelevant for a tests-only diff). Aggregation ran in-orchestrator given the small finding set. No public API surface → no `needs-api-review` label needed. Merge approval stays with a human maintainer.*

</details>

liramon2:
@strandly-the-agent Re-review this PR. Consolidate any findings into 1 comment.

strandly-the-agent:
**Re-review of `0f47c13` ("fix: address strandly") — both 🟡 findings are fixed, and the fixes hold up under attack. No new findings.** Once CI finishes on this head (unit + *Run integration tests* were still in flight when I checked) this is good to merge, pending human approval as before.

**Previous findings ledger:**
- ✅ **ADK `overall_score >= 0.75`** → replaced with `sum(s < 0.5 for s in report.scores) <= 1` (`test_adk_eval.py:266-267`, `len == 4` check retained). Exactly the tighter shape suggested; a single-row zero (e.g. an argument-mapping regression on `convert_units`) now fails the test again while one harsh-judge row is still tolerated.
- ✅ **Tool-selection parked at exactly 0.5** → the case input now says *"according to the local weather dataset"* and the specialist prompt/description match (`test_claude_openinference_eval.py:187,196,205`). **Verified live**: re-ran the reconstructed tool-selection judge against the new wording — the Bash call flipped from **No 8/8** (old wording) to **Yes 8/8**, and `Agent(weather-specialist)` stayed Yes 8/8. Expected `scores[2]` ≈ 1.0 now, so the `>= 0.5` threshold has real headroom instead of zero.
- ✅ Appendix nits picked up too: `bash -c` nesting dropped, "EXACTLY"/"Replace" contradiction reworded, dead `expected_output` removed.

**Attacked the fix itself** (the failure mode of iterative review is a fix that trades one bug for another):
- The new prompt command verified locally: quoting is sound (double-quoted shell, single-quoted Python literals, no `$`/backticks) and prints exactly `Rainy, 55F` with `<city>`→`Seattle`.
- The new `expected_assertion` adds a **delegation claim**, which is also consumed by the Correctness reference judge — and that judge sees *only the final response text* (`correctness_evaluator.py:160-164`), which can't prove delegation. Probed it live 18/18 **CORRECT** across terse/sentence/verbose responses (the template's "omitting minor supplementary details is acceptable" rule covers it), so not a finding — just a coupling worth knowing (one appendix line below).

<details>
<summary>Verification ledger + appendix (1)</summary>

### Verified
- ✅ Delta reviewed: `747945e..0f47c13`, touches only the two hunks addressing my findings — ran this pass myself per follow-up-review procedure (no new behavior, fixes-only delta).
- ✅ Rendered the new specialist prompt string from source and executed the embedded command via `bash -c` with `<city>`→`Seattle`: rc=0, stdout `Rainy, 55F`.
- ✅ Tool-selection judge probe (real v0 system prompt from `prompt_templates/tool_selection_accuracy_v0.py`, reconstructed input): `Bash(local dataset lookup)` Yes 8/8; `Agent(weather-specialist)` Yes 8/8.
- ✅ Correctness reference judge probe (real v0 reference template): 18/18 CORRECT including a bare `Rainy, 55F` response against the delegation-bearing assertion.
- ⚠️ Same caveat as last round: probes used Sonnet 4.5 (pinned Sonnet 4 is Bedrock-Legacy-gated in my account); judge prompts reconstructed, not captured. Directions were unanimous; exact rates aren't gospel.
- ⏳ CI on `0f47c13`: unit/lint/integration jobs were IN_PROGRESS at review time — worth a green *Run integration tests* before merging, as with the previous head.

### Appendix — non-blocking (1)
- ⚪ The delegation clause in `expected_assertion` (`test_claude_openinference_eval.py:207-210`) is verifiable by the session-level GoalSuccess judge (sees the full conversation record) but not by the trace-level Correctness judge (final response only). Today's reference template tolerates that; if Correctness ever starts failing with "cannot confirm delegation", the fix is to keep the delegation clause out of the shared assertion or split assertions per evaluator.

</details>

arielnabavian (approved):

PR #375, opened 2026-08-18
fix: change bridge_parent_gaps to return new spans instead of mutating in place

liramon2:
## Description
`bridge_parent_gaps` currently mutates its input spans in-place. This hides the dependency between it and downstream consumers like `model_post_init`. This PR makes `bridge_parent_gaps` return the modified spans so that call sites explicitly operate on the result.

This PR also adds a `bridge_parent_gaps` call to `GenericGenAISessionMapper` in preparation for OpenAI Agents support. OpenAI Agents add unconverted workflow spans (Traceloop/gen_ai) between converted spans, which make their parent span ids incorrect.

## Related Issues

https://github.com/strands-agents/evals/pull/365

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/src/strands_evals/mappers/adk_otel_session_mapper.py
+++ b/src/strands_evals/mappers/adk_otel_session_mapper.py
@@ -144,7 +144,7 @@ class ADKOtelSessionMapper(SessionMapper):
         }
 
         converted_spans.sort(key=lambda s: s.span_info.start_time)
-        bridge_parent_gaps(converted_spans, raw_parent_map)
+        converted_spans = bridge_parent_gaps(converted_spans, raw_parent_map)
 
         return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
 
--- a/src/strands_evals/mappers/generic_gen_ai_session_mapper.py
+++ b/src/strands_evals/mappers/generic_gen_ai_session_mapper.py
@@ -30,7 +30,7 @@ from ..types.trace import (
     UserMessage,
 )
 from .session_mapper import SessionMapper
-from .utils import join_tool_result_content
+from .utils import bridge_parent_gaps, join_tool_result_content
 
 logger = logging.getLogger(__name__)
 
@@ -149,6 +149,12 @@ class GenericGenAISessionMapper(SessionMapper):
                     span.get("span_id", "?"),
                     e,
                 )
+
+        # Fix parent_span_id on converted spans that point to skipped intermediaries
+        raw_parent_map: dict[str, str | None] = {
+            sid: s.get("parent_span_id") for s in spans if (sid := s.get("span_id"))
+        }
+        converted_spans = bridge_parent_gaps(converted_spans, raw_parent_map)
 
         return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
 
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -42,7 +42,7 @@ from .constants import (
     SCOPES_OPENINFERENCE_FAMILY,
 )
 from .session_mapper import SessionMapper
-from .utils import safe_json_parse
+from .utils import bridge_parent_gaps, safe_json_parse
 
 logger = logging.getLogger(__name__)
 
@@ -298,6 +298,12 @@ class OpenInferenceSessionMapper(SessionMapper):
             for converted in converted_spans:
                 if isinstance(converted, AgentInvocationSpan) and not converted.available_tools:
                     converted.available_tools = tools_list
+
+        # Fix parent_span_id on converted spans that point to skipped intermediaries
+        raw_parent_map: dict[str, str | None] = {
+            sid: s.get("parent_span_id") for s in spans if (sid := s.get("span_id"))
+        }
+        converted_spans = bridge_parent_gaps(converted_spans, raw_parent_map)
 
         return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
 
--- a/src/strands_evals/mappers/openinference_session_mapper.py
+++ b/src/strands_evals/mappers/openinference_session_mapper.py
@@ -42,7 +42,7 @@ from .constants import (
     SCOPES_OPENINFERENCE_FAMILY,
 )
 from .session_mapper import SessionMapper
-from .utils import bridge_parent_gaps, safe_json_parse
+from .utils import safe_json_parse
 
 logger = logging.getLogger(__name__)
 
@@ -298,12 +298,6 @@ class OpenInferenceSessionMapper(SessionMapper):
             for converted in converted_spans:
                 if isinstance(converted, AgentInvocationSpan) and not converted.available_tools:
                     converted.available_tools = tools_list
-
-        # Fix parent_span_id on converted spans that point to skipped intermediaries
-        raw_parent_map: dict[str, str | None] = {
-            sid: s.get("parent_span_id") for s in spans if (sid := s.get("span_id"))
-        }
-        converted_spans = bridge_parent_gaps(converted_spans, raw_parent_map)
 
         return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
 
--- a/src/strands_evals/mappers/opensearch_session_mapper.py
+++ b/src/strands_evals/mappers/opensearch_session_mapper.py
@@ -100,7 +100,7 @@ class OpenSearchSessionMapper(SessionMapper):
             r.span_id: (r.parent_span_id or None) for r in sorted_records if r.span_id
         }
 
-        bridge_parent_gaps(spans, raw_parent_map)
+        spans = bridge_parent_gaps(spans, raw_parent_map)
 
         return Trace(spans=spans, trace_id=trace_id, session_id=session_id)
 
--- a/src/strands_evals/mappers/strands_in_memory_session_mapper.py
+++ b/src/strands_evals/mappers/strands_in_memory_session_mapper.py
@@ -162,7 +162,7 @@ class StrandsInMemorySessionMapper(SessionMapper):
             parent_id = format(span.parent.span_id, "016x") if span.parent else None
             raw_parent_map[span_id] = parent_id
 
-        bridge_parent_gaps(converted_spans, raw_parent_map)
+        converted_spans = bridge_parent_gaps(converted_spans, raw_parent_map)
 
         return Trace(spans=converted_spans, trace_id=trace_id, session_id=session_id)
 
--- a/src/strands_evals/mappers/utils.py
+++ b/src/strands_evals/mappers/utils.py
@@ -278,30 +278,40 @@ def readable_spans_to_dicts(spans: Any) -> list[dict]:
 def bridge_parent_gaps(
     converted_spans: list[SpanUnion],
     raw_parent_map: dict[str, str | None],
-) -> None:
-    """Patch parent_span_id on converted spans to skip unconverted intermediates.
+) -> list[SpanUnion]:
+    """Return spans with parent_span_id adjusted to skip unconverted intermediates.
 
-    When a converted span's parent_span_id points to a span that wasn't
-    converted (e.g. execute_event_loop_cycle in Strands SDK, call_llm in ADK),
-    this walks up via raw_parent_map until it finds a converted ancestor — or
-    sets parent_span_id to None if no converted ancestor exists.
+    Walks up `raw_parent_map` to find the nearest converted ancestor. Sets
+    parent_span_id to None if no converted ancestor exists.
 
     Args:
-        converted_spans: Flat list of converted spans from a single trace.
-        raw_parent_map: span_id -> parent_span_id for ALL raw spans, including
-            unconverted ones.
+        converted_spans: Converted spans from a single trace.
+        raw_parent_map: span_id -> parent_span_id for all raw spans (including unconverted).
+
+    Returns:
+        New list of spans with corrected parent_span_id values.
     """
     converted_ids = {s.span_info.span_id for s in converted_spans if s.span_info.span_id}
+    result: list[SpanUnion] = []
     for span in converted_spans:
         parent_id = span.span_info.parent_span_id
-        if parent_id and parent_id not in converted_ids:
-            visited: set[str] = set()
-            current: str | None = parent_id
-            while current and current not in visited:
-                visited.add(current)
-                if current in converted_ids:
-                    span.span_info.parent_span_id = current
-                    break
-                current = raw_parent_map.get(current)
-            else:
-                span.span_info.parent_span_id = None
+        # No-op if the parent id points to no span or a converted span
+        if not parent_id or parent_id in converted_ids:
+            result.append(span)
+            continue
+
+        # Walk up to find converted ancestor
+        visited: set[str] = set()
+        current: str | None = parent_id
+        new_parent: str | None = None
+        while current and current not in visited:
+            visited.add(current)
+            if current in converted_ids:
+                new_parent = current
+                break
+            current = raw_parent_map.get(current)
+        patched = span.model_copy(
+            update={"span_info": span.span_info.model_copy(update={"parent_span_id": new_parent})}
+        )
+        result.append(patched)
+    return result
--- a/src/strands_evals/mappers/utils.py
+++ b/src/strands_evals/mappers/utils.py
@@ -281,12 +281,15 @@ def bridge_parent_gaps(
 ) -> list[SpanUnion]:
     """Return spans with parent_span_id adjusted to skip unconverted intermediates.
 
-    Walks up `raw_parent_map` to find the nearest converted ancestor. Sets
-    parent_span_id to None if no converted ancestor exists.
+    When a converted span's parent_span_id points to a span that wasn't
+    converted (e.g. execute_event_loop_cycle in Strands SDK, call_llm in ADK),
+    this walks up via raw_parent_map until it finds a converted ancestor — or
+    sets parent_span_id to None if no converted ancestor exists.
 
     Args:
-        converted_spans: Converted spans from a single trace.
-        raw_parent_map: span_id -> parent_span_id for all raw spans (including unconverted).
+        converted_spans: Flat list of converted spans from a single trace.
+        raw_parent_map: span_id -> parent_span_id for ALL raw spans, including
+            unconverted ones.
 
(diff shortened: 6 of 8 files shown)


liramon2:
@strandly-the-agent Review this PR. Consolidate your findings. Keep the findings relevant to the changes made, not to pre-existing issues.

strandly-the-agent (changes requested):
**Verdict: changes-requested** — draft, **no 🔴 blockers**; 2 🟡 worth landing before it leaves draft. The `bridge_parent_gaps` rewrite itself is clean (20k-trial old-vs-new differential: 0 mismatches, no regression for the 3 existing callers); both problems are in the two **new** call sites. Scoped to this diff per your ask — pre-existing items get one appendix line each.

**Verified**
- ✅ `1b9aad8` · `pytest tests/strands_evals/mappers/` → **360 passed** · mappers+types → **452 passed** · `ruff check` + `format --check` clean · ⚠️ mypy not installed in my sandbox, so `hatch run test-lint` is unverified.
- ✅ Old-vs-new `bridge_parent_gaps` differential, 20 000 random topologies (cycles, self-loops, missing entries, `""`/`None` ids) → **0 mismatches**. 7 mutants on `utils.py:278-317`: M1/M5/M6 killed; **M2 (copy every span), M3 (drop the `converted_ids` short-circuit), M4 (drop the cycle guard) survive**.
- ✅ Neutering **both** new call sites (discarded-return form) → **452 passed** — no test notices.
- ✅ Fixture differential (`map_to_session`, main → this branch): **11 spans change `parent_span_id`** across `claude_adot`, `claude_live`, `smolagents`; every `agent_span_id` and both Claude root-agent selections **unchanged**; GenAI fixtures byte-identical. Extractor output: 28 changed lines, all of them `parent_span_id`.
- ✅ The 4 suggested tests below: pass at head (454 mappers+types), the 2 mapper ones **fail exactly when their call site is neutered**, ruff clean.
- ✅ Finding 1's fix applied in a scratch tree: **452 passed**, mapper output **byte-identical on every fixture** — behaviour-neutral today.

### 🟡 1 — `openinference_session_mapper.py:303-305`: the map is built from the *scope-filtered* span list, so the walk can't cross a foreign-instrumentation span — it nulls the parent instead of bridging

The map only ever contains `SCOPES_OPENINFERENCE_FAMILY` spans (filtered at `:122`), which contradicts the contract this PR's own docstring states at `utils.py:289` (*"for all raw spans (including unconverted)"*). Notably, the Traceloop workflow spans the PR body names as motivation carry `SCOPE_LANGCHAIN_OTEL` (`constants.py:8`), which is **not** in the family (`constants.py:16-22`) — i.e. exactly the class the map excludes.

<details><summary>Evidence, honest reachability caveat, and the fix (verified)</summary>

`spans` inside `_build_trace` is the per-trace slice of `openinference_spans`, already filtered at `:122`. Any intermediary from another instrumentation library is absent from the map, so the walk at `utils.py:307-312` ends on the first hop and `new_parent` stays `None`.

The excluded class is real and already in-repo: `fixtures/claude_adot_spans.json` contains `7549860608387d0b` (`POST /invocations`, scope `opentelemetry.instrumentation.starlette`) as the parent of the converted root agent `c3d1bcc351bf9653`.

**Caveat, verified:** on that fixture an unfiltered map would *not* change the outcome either — the starlette span's own parent (`0ddd1d7a4995eb19`) isn't in the export, so the walk still ends at `None`. The loss only materialises when a converted ancestor sits **above** the foreign span. Synthetic claude-SDK trace `A(agent) → B(agent) → S(starlette) → T(tool)`:

| | `T.parent_span_id` | `T.agent_span_id` |
|---|---|---|
| main | `s1` (dangling) | `a1` ❌ |
| this branch | `None` | `a1` ❌ |
| + map over unfiltered spans | `b1` ✅ | `b1` ✅ |

So: not a regression (main leaves an equally unusable dangling id), but the map construction this PR *adds* can't do the job the PR describes for the OpenInference mapper. (#366's `openinference.instrumentation.openai_agents` **is** added to the family, so that specific case works — this is about everything else in a mixed export.)

**Fix** — build the map once in `map_to_session` from the pre-filter `spans` (`:119`) and pass it in; drop `:303-305`:

```python
# map_to_session, after :119 — before the family filter at :122
raw_parent_map: dict[str, str | None] = {
    sid: s.get("parent_span_id") for s in spans if (sid := s.get("span_id"))
}
...
    trace = self._build_trace(trace_id, trace_spans, session_id, raw_parent_map)
```

with `_build_trace(self, trace_id, spans, session_id, raw_parent_map: dict[str, str | None])`.

Span ids are unique and the walk stops at `converted_ids`, so a session-wide map is safe; if you'd rather keep it trace-local, group the unfiltered spans by `trace_id` first. I ran exactly this in a scratch copy: **452 passed**, and mapper output is byte-identical on all fixtures — it changes nothing today and only starts mattering for the mixed-instrumentation traces this PR is preparing for.
</details>

### 🟡 2 — `openinference_session_mapper.py:306` + `generic_gen_ai_session_mapper.py:157`: both new call sites have zero coverage, and the OpenInference one already changes output on 3 real fixtures

Replacing both with the discarded-return (now no-op) form leaves **452 passed** — nothing in `tests/` fails if either call is removed, mis-ordered, or handed a wrong map, which `AGENTS.md` asks for explicitly ("exercise the behavior, not just pass"). Meanwhile the OpenInference call reparents 11 spans on existing fixtures with no assertion covering any of it.

<details><summary>The output change, and 2 paste-ready tests (both verified: pass at head, fail when their call site is removed)</summary>

```
claude_adot_spans.json
  ToolExecutionSpan 9ba74ad8   parent 5c2a984a → 52c63a60
  ToolExecutionSpan d0706922   parent f120f1e6 → 5399ec80
  AgentInvocationSpan c3d1bcc3 parent 7549860608387d0b → None   # starlette parent, filtered out
claude_live_spans.json
  ToolExecutionSpan 1caa50d4   parent 053c311b → 51e6bd91
  ToolExecutionSpan 0030c035   parent a6e166c4 → 1c51a9da
  ToolExecutionSpan b29670c3   parent f028c728 → a6e40546
  AgentInvocationSpan ebdd7630 parent 9be1bbbd → None
smolagents_live_spans.json
  InferenceSpan 527aa0aa       parent 11314f4c → 93fcaaeb
  ToolExecutionSpan 6f3d4f46   parent 11314f4c → 93fcaaeb
  InferenceSpan 9f024ee5       parent d3587aff → 93fcaaeb
  ToolExecutionSpan 22dbd248   parent d3587aff → 93fcaaeb
```

The new parents look right (Claude's Bash tools reparent onto the `Agent`/Task span that launched the sub-agent; smolagents spans skip the `Step N` CHAIN spans onto the real agent span), and no `agent_span_id` or root selection moves — but the PR body describes this as preparation only, so it's worth a line in the description too.

**GenAI** — add before `class TestInferenceSpan` in `test_generic_gen_ai_session_mapper.py` (uses the existing `make_span` helper):

```python
class TestUnconvertedSpanReparenting:
    def test_tool_under_unconverted_span_reparents_to_agent(self):
        """Tool nested under an unconverted workflow span is reparented to the converted agent."""
        agent = make_span(span_id="agent-1", operation_name="invoke_agent")
        workflow = make_span(span_id="workflow-1", parent_span_id="agent-1", name="my_flow.workflow")
        tool = make_span(
            span_id="tool-1",
            parent_span_id="workflow-1",
            operation_name="execute_tool",
            attributes={
                "gen_ai.tool.name": "calculator",
                "gen_ai.tool.call.id": "call-1",
                "gen_ai.tool.input": '{"expr": "2+2"}',
                "gen_ai.tool.output": "4",
                "gen_ai.tool.status": "success",
            },
        )

        session = GenericGenAISessionMapper().map_to_session([agent, workflow, tool], "sess-1")

        spans = session.traces[0].spans
        assert [s.span_info.span_id for s in spans] == ["agent-1", "tool-1"]
        tool_span = next(s for s in spans if isinstance(s, ToolExecutionSpan))
        assert tool_span.span_info.parent_span_id == "agent-1"
        assert tool_span.agent_span_id == "agent-1"
```

**OpenInference** — add to `TestSmolagentsFixtureIntegration` (reuses the module-scoped fixture; no new fixture file, no hardcoded span ids):

```python
    def test_step_chain_spans_are_bridged(self, smolagents_session):
        """Unconverted "Step N" CHAIN spans are skipped: children reparent to CodeAgent.run."""
        trace = smolagents_session.traces[0]
        agent = next(s for s in trace.spans if isinstance(s, AgentInvocationSpan))
        agent_id = agent.span_info.span_id
        children = [s for s in trace.spans if s is not agent]
        assert children
        assert {s.span_info.parent_span_id for s in children} == {agent_id}
        for tool in (s for s in children if isinstance(s, ToolExecutionSpan)):
            assert tool.agent_span_id == agent_id
```
</details>

### ⚪ 3 — `utils.py:298` comment says the opposite of what the branch does; docstring doesn't pin the new contract

<details><summary>Suggested wording</summary>

```python
# No-op if the parent id points to no span or a converted span
if not parent_id or parent_id in converted_ids:
```

`not parent_id` means *there is no parent id*; a parent id that "points to no span" is precisely the dangling case that goes **down** the walk path. Suggested: `# Nothing to bridge: no parent, or the parent was converted`.

The docstring (`:282-293`) also lost the concrete examples the old one carried (`execute_event_loop_cycle` in Strands, `call_llm` in ADK) and doesn't record the contract the rewrite introduces — the returned list is a deliberate **mix of originals** (pass-through, `:300`) **and shallow copies** (`:313`) that share `metadata`/`messages`/`tool_call`/`available_tools` with their originals. One line in `Returns:` covers it:

```
    Input spans are never modified; callers must use the returned list. Patched
    spans are shallow copies, so a caller that mutates spans after bridging must
    mutate the returned objects (e.g. `Trace.model_post_init` writing agent_span_id).
```

The PR body's rationale ("to explicitly show the dependency between `bridge_parent_gaps` and `model_post_init`") belongs here rather than only in the description.
</details>

### Questions

**❓ (blocking)** Is the new Claude hierarchy the shape you want? Bash tool `9ba74ad8` now parents onto `52c63a60` — the **`Agent` tool span** that spawned the sub-agent — i.e. a tool under a tool. It's an improvement over a dangling id, `Trace.model_post_init` traverses it fine, and `agent_span_id` is unchanged; but should the walk stop only at an agent/inference ancestor instead? No test pins the shape either way.

**❓ (non-blocking)** Was placing the bridge *after* the LangGraph dedup (`:284-288`) and the `available_tools` back-fill (`:296-300`) deliberate? It looks like the correct order, and it's load-bearing: the dedup compares by identity (`s is root`, `:288`), so if bridging ever moves above it, `root` becomes a stale pre-copy object and the filter drops **every** agent span (mechanism repro'd standalone: `root patched? True; agent spans kept: 0 of 2`; not reachable in the diff as written). #366 currently bridges at that earlier point (right after the conversion loop), which also leaves children of dropped agent spans dangling — so #366 probably wants to adopt this ordering rather than the reverse. A one-line comment at `:302` would keep a future reorder from being invisible.

<details><summary>Appendix — non-blocking (6)</summary>

- ⚪ **Two 5-line `utils` tests kill the 3 surviving mutants** (both verified: pass at head, ruff clean). Nothing pins that unpatched spans come back **by identity** (`utils.py:299-301` — kills M2 and M3), and the `visited` cycle guard (`utils.py:304-308`) is untested — the guard works; only its test is missing (removing it makes cyclic input hang rather than fail).

  ```python
  def test_unpatched_spans_are_passed_through(self):
      """Spans needing no bridge are returned unchanged (same objects)."""
      agent = AgentInvocationSpan(
          span_info=_span_info("agent", None), user_prompt="hi", agent_response="hey", available_tools=[]
      )
      tool = ToolExecutionSpan(
          span_info=_span_info("tool", parent_span_id="agent"),
          tool_call=ToolCall(name="x", arguments={}),
          tool_result=ToolResult(content="y"),
      )
      result = bridge_parent_gaps([agent, tool], {"agent": None, "tool": "agent"})
      assert result[0] is agent
      assert result[1] is tool

  def test_cyclic_parent_chain_terminates(self):
      """A cycle in raw_parent_map terminates instead of looping forever."""
      tool = ToolExecutionSpan(
          span_info=_span_info("tool", parent_span_id="a"),
          tool_call=ToolCall(name="x", arguments={}),
          tool_result=ToolResult(content="y"),
      )
      result = bridge_parent_gaps([tool], {"a": "b", "b": "a", "tool": "a"})
      assert result[0].span_info.parent_span_id is None
  ```

- ⚪ **PR description**, two additions: the `parent_span_id` change on the Claude/smolagents fixtures (it isn't preparation-only for existing users), and an explicit callout that `bridge_parent_gaps` — exported in `mappers/__init__.py` `__all__` — is now source-compatible but **semantically breaking** for any caller relying on mutation. Org-wide code search finds no external caller, so a deprecation shim would be ceremony.
- ⚪ **Duplicated map construction** — `generic_gen_ai:154-156` and `openinference:303-305` are character-identical, and there are now 5 sites building this map in 3 shapes (`adk_otel:142-144`, `opensearch:101-102`, `strands_in_memory:160-164`). A `build_raw_parent_map(span_dicts)` next to `bridge_parent_gaps` would cover the two dict-based ones. Moot if you take finding 1 (that construction moves anyway).
- **Rebase heads-up, not a defect: #365/#366.** Both still call `bridge_parent_gaps(...)` as a statement (pr365 `generic_gen_ai:166`, pr366 `openinference:461`), so after this lands their bridging is a no-op until rebased — but it is **loud, not silent**: swapping this PR's `utils.py` into each branch gives pr365 **4 failures** (`test_tool_span_reparented_across_unconverted_intermediate`, `test_handoff_agents_are_flat_with_coordinator_as_root`, `test_adot_coordinator_tools_empty`, `test_adot_math_specialist_tools`) and pr366 **1 failure** (`test_multiply_numbers_owned_by_math_specialist`). Their own suites catch it, so just a rebase note. Side benefit: those failures are direct evidence the new call sites are load-bearing for multi-agent traces.
- **Out of scope, same root as finding 1:** `generic_gen_ai_session_mapper.py:154-156` has a smaller version of the gap — `_convert_trace` receives spans already dropped by the foreign-`session.id` filter (`:103-104`), so an intermediary tagged with a different conversation id is missing from the map. Low value on its own; worth fixing consistently if you touch the OpenInference one.
- **Pre-existing (not this PR):** (a) `bridge_parent_gaps` can emit `parent_span_id == own span_id` when the raw chain cycles back through the span's own id (`{"T": "U", "U": "T"}` → `utils.py:307-311` picks `T`) — **identical on main**, `Trace` constructs fine, no hang (`model_post_init`'s seen-guard handles it), benign. (b) `tests/strands_evals/cli/test_fetch.py::test_fetch_langfuse_writes_session_json` fails identically on `origin/main` (missing langfuse dep) — unrelated.
</details>

<details><summary>Per-pass breakdown</summary>

- **Correctness/safety** — changes-requested: the rewrite is behaviourally identical to the in-place version (20k-trial differential, 0 mismatches; no regression for the 3 existing callers, all other fields preserved); both problems are the new call sites → findings 1, 2, 3.
- **Test quality** — gaps-found: 7 mutants on `utils.py`; M2/M3/M4 survive, M7 (whole function → `return converted_spans`) is caught only by `test_utils`, no mapper test notices → finding 2 and the appendix's two `utils` tests. Explicitly *not* asking for multi-hop or missing-map-entry tests (already covered indirectly).
- **API/DevX bar-raiser** — accept-with-conditions: pure-return is right for this module (`bridge_parent_gaps` was its only mutator); conditions are the docstring contract (→ 3) and the description callout (→ appendix). *Only this pass's summary survived its run; its detailed block was lost.* No api-review label workflow exists in this repo (checked `.github/`), so no process finding.
- **Adversarial** — no blockers: root selection and ownership unchanged on both Claude fixtures (`_find_root_agent_span` tier moves, but each has exactly one converted agent so it can't flip); the mixed originals/copies × in-place `model_post_init` hazard is contrived in-repo (all 9 `Trace(...)` sites are one statement after their bridge call, nothing reads pre-bridge locals). *Ran as a focused 3-probe retry after a broader run timed out.*
- **Aggregation** — 4 passes in → 13 raw findings → 3 posted (🟡2 ⚪1) + 2 questions + 6 appendix lines.
</details>

liramon2:
@strandly-the-agent Re-review this PR. Keep findings in a single comment.

strandly-the-agent (approved):
**Verdict: approve** — re-reviewed the delta `1b9aad8` → `43ba5c9`; every finding from my previous review is resolved, no new issues. (Follow-up shape: fixes-only delta, so I verified it directly rather than re-running the full multi-pass pipeline.)

**Previous findings — status**
- ✅ **🟡1 (filtered `raw_parent_map` in OpenInference)** — resolved by reverting the OpenInference call site; `openinference_session_mapper.py` is now byte-identical to `main` (0-line diff). Both questions (tool-under-TOOL hierarchy, bridge-after-dedup ordering) are moot here and move to #366's court.
- ✅ **🟡2 (new call site untested)** — fixed: `TestUnconvertedSpanReparenting` (3 tests, incl. multi-hop and no-converted-ancestor). I re-ran the mutation check: neutering the GenAI bridge call → **3 failed / 77 passed**, so the coverage is real, not decorative.
- ✅ **⚪3 (comment/docstring)** — comment fixed, concrete examples restored. One optional leftover: `Returns:` still doesn't say the list mixes pass-through originals with shallow copies — fine to leave.
- ✅ Title + description updated to match the reduced scope (GenAI only).

**Verified at `43ba5c9`**
- ✅ `pytest tests/strands_evals/mappers/ tests/strands_evals/types/` → **455 passed** (452 + the 3 new) · `ruff check` + `format --check` clean
- ✅ New tests pass at head and fail exactly when the call site is removed

<details><summary>Still-standing appendix notes (unchanged, non-blocking)</summary>

- Rebase heads-up for #365/#366: their in-place `bridge_parent_gaps(...)` statements become no-ops after this lands — loud, not silent (their own suites fail: pr365 ×4, pr366 ×1). #366 also inherits the two OpenInference questions above, plus the `s.get("span_id", "")` map-key shape and its bridge-before-dedup placement.
- Optional `utils` tests from my last review (identity pass-through of unpatched spans; cycle-guard termination) weren't added — appendix-tier then, appendix-tier now.
</details>

jjbuck (approved):
I assume this will need to be merged before #365, which will be rebased. Is that right?

liramon2:
Yes, this will be merged and then #365 will be rebased on top of this.

mehtarac (approved):

PR #377, opened 2026-08-25
fix: ensure parse_timestamp always returns timezone-aware UTC datetime

### LINKED ISSUE #372: [BUG] SessionMapper.parse_timestamp returns naive datetimes despite docstring promising timezone-aware UTC — mixed sessions break datetime comparisons
### Checks
- [x] I have searched ./issues and there are no duplicates of my issue

### Strands Version
1.52.0

### Strands Evals Version
main @ cc0a391 (v1.0.3)

### Python Version
3.13.14

### Operating System
Linux (aarch64)

### Installation Method
git clone

### Steps to Reproduce
```python
from strands_evals.mappers.session_mapper import SessionMapper

m = SessionMapper.__new__(SessionMapper)  # any concrete mapper
m.parse_timestamp("2024-01-01T00:00:00").tzinfo   # -> None (naive)
m.parse_timestamp("2024-01-01T00:00:00Z").tzinfo  # -> UTC (aware)
m.parse_timestamp(None).tzinfo                    # -> UTC (aware)
```

### Expected Behavior
Per the docstring (`mappers/session_mapper.py:91`): "Returns: Timezone-aware datetime in UTC" — for every input shape.

### Actual Behavior
Two of the branches violate the contract (`session_mapper.py:95-96` passes a naive `datetime` through unchanged; `:103` `datetime.fromisoformat` returns naive for offset-less ISO strings). A single mapped session can therefore mix naive and aware `start_time`s — e.g. one span missing `start_time` (aware `now()`) alongside offset-less ISO stamps (naive). `parse_timestamp` is used by the openinference (`:602`), langchain (`:454`), and adk (`:569`) mappers.

Any consumer that compares or sorts `span_info.start_time` across spans then raises `TypeError: can't compare offset-naive and offset-aware datetimes` (verified; e.g. the sort in #343's proposed `TraceIndex`). Existing code defends against this individually — `extractors/trace_extractor.py:24-28` has `_to_aware_utc` — but the mapper contract itself is broken.

### Possible Solution
Normalize in `parse_timestamp` itself: apply the `_to_aware_utc` treatment (assume UTC for naive input) to the datetime-passthrough and `fromisoformat` branches, making the docstring true and letting downstream defenses retire.

### Related Issues
#343

liramon2:
## Description

`SessionMapper.parse_timestamp` documents that it returns a "timezone-aware datetime in UTC", but two branches violated this contract:

1. The `datetime` passthrough branch returned naive datetimes unchanged.
2. The `fromisoformat` branch returned naive datetimes for offset-less ISO strings (e.g. `"2024-01-01T00:00:00"`).

This caused `TypeError: can't compare offset-naive and offset-aware datetimes` when downstream code (extractors, sort logic) mixed spans whose timestamps came from different branches.

The fix normalizes both branches: naive datetimes get `tzinfo=UTC` attached, and aware non-UTC datetimes are converted to UTC via `astimezone`. This PR also moves the tests for `parse_timestamp` into a separate test file targeting the base session mapper.

## Related Issues

Closes #372

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

## Checklist

----


liramon2:
@strandly-the-agent Review this PR.

src/strands_evals/mappers/session_mapper.py
```diff
@@ -93,14 +93,19 @@ def parse_timestamp(self, value: Any) -> datetime:
...
             try:
                 if value.endswith("Z"):
                     value = value[:-1] + "+00:00"
-                return datetime.fromisoformat(value)
+                dt = datetime.fromisoformat(value)
+                if dt.tzinfo is None:
+                    return dt.replace(tzinfo=timezone.utc)
+                return dt.astimezone(timezone.utc)
             except ValueError:
                 return datetime.now(timezone.utc)
```
strandly-the-agent, line 116: 🟡 **Introduced regression: boundary timestamps now raise `OverflowError`, and every call site silently swallows it into a dropped span.**

`astimezone(timezone.utc)` overflows for instants within ~14h of `datetime.min`/`max` carrying a non-UTC offset — inputs `main` returned a value for. The string branch's `except ValueError:` doesn't catch `OverflowError`, and the datetime branch has no guard at all:

```
>>> mapper.parse_timestamp("9999-12-31T23:59:59-05:00")   # main: aware -05:00 datetime
OverflowError: date value out of range
>>> mapper.parse_timestamp("0001-01-01T00:00:00+05:00")   # main: aware +05:00 datetime
OverflowError: date value out of range
```

Reachability is admittedly contrived (boundary year *plus* non-zero offset — real OTLP emitters send epoch nanos or `Z`), but two things promote it above a nit: it's a behavior regression on lines this diff writes, and the failure is **silent** — every `parse_timestamp` call site sits inside a per-span `except Exception` (`adk_otel:137`, `openinference:403`, `langchain_otel:140`, `generic_gen_ai:127`), so verified end to end via `ADKOtelSessionMapper` the result is `traces=0 spans=0` where `main` produced `traces=1 spans=1`: data loss with only a `logger.warning`. It also breaks this function's own convention of being *total* — every other bad input degrades to `now(utc)`.

Verified fix (all three repros fall back to `now(utc)`; the 170 tests on the touched files still pass):

```suggestion
        if isinstance(value, datetime):
            if value.tzinfo is None:
                return value.replace(tzinfo=timezone.utc)
            try:
                return value.astimezone(timezone.utc)
            except OverflowError:
                return datetime.now(timezone.utc)
        if isinstance(value, str):
            if value.isdigit():
                return datetime.fromtimestamp(int(value) / 1e9, tz=timezone.utc)
            try:
                if value.endswith("Z"):
                    value = value[:-1] + "+00:00"
                dt = datetime.fromisoformat(value)
                if dt.tzinfo is None:
                    return dt.replace(tzinfo=timezone.utc)
                return dt.astimezone(timezone.utc)
            except (ValueError, OverflowError):
                return datetime.now(timezone.utc)
```

Alternative worth knowing: the in-repo `_to_aware_utc` convention (`types/trace.py:128-132`) returns aware values *as-is* and can't overflow at all (aware-vs-aware comparisons sort fine regardless of offset) — but that would drop the `tzinfo == timezone.utc` guarantee your new tests assert, so the guard above is the one that preserves your stated semantics.

tests/strands_evals/mappers/test_session_mapper.py
```diff
@@ -0,0 +1,77 @@
...
+    def test_iso_string_with_offset_normalized(self):
+        """An ISO string with a non-UTC offset is converted to UTC."""
+        result = self.mapper.parse_timestamp("2024-06-15T17:00:00+05:00")
+        assert result.tzinfo == timezone.utc
+        assert result == datetime(2024, 6, 15, 12, 0, 0, tzinfo=timezone.utc)
+
+    def test_invalid_string_returns_aware_utc(self):
+        """An unparseable string falls back to current UTC time."""
+        result = self.mapper.parse_timestamp("not-a-timestamp")
+        assert result.tzinfo is not None
```
strandly-the-agent: 🟡 **These 10 tests pin both fixed branches (mutation-verified: reverting either branch is caught by exactly the right tests) — but nothing anywhere pins the *symptom* #372 reports.**

The downstream failure is a silently dropped span, and the existing e2e test for exactly this scenario — `test_normalization_survives_heterogeneous_start_times` (`test_openinference_session_mapper.py:2406`) — feeds only an int-ns epoch (`:2442`) and a `Z` string (`:2459`), both already aware on `main`. I verified it **passes on `main`'s buggy code** and survives reverting either fixed branch. Every `start_time` fixture repo-wide is int-ns, so a future refactor of `parse_timestamp` could regress #372 without turning anything red — and the failure mode is silent span loss, not a test-visible crash.

One-character fix, which I ran both ways (fails on `main` with #372's exact `can't compare offset-naive and offset-aware datetimes` + dropped agent span; passes on `da5f29a`):

```diff
# tests/strands_evals/mappers/test_openinference_session_mapper.py:2459
-            start_time="2024-01-01T00:00:00Z",
-        )  # ISO string
+            start_time="2024-01-01T00:00:00",
+        )  # ISO string, no offset -> naive before the fix
```

The `Z` path stays covered by `:26-29` here. If you'd rather keep both, a third turn/LLM pair with `start_time=datetime(2024, 1, 1)` has the same effect for ~10 more lines. (Micro-nit while you're there: that test's docstring mentions a `None` start_time case, but `make_span` defaults to int-ns, so none of its spans is actually `None`.)

strandly-the-agent (commented):
**Verdict: solid fix, correct and verified end to end — three should-fixes (🟡) before it leaves draft, none blocking.** The docstring #372 asked to make true is still stale, nothing yet pins the reported *symptom*, and boundary timestamps now raise a silently-swallowed `OverflowError` where `main` returned a value.

✅ Reviewed `da5f29a` vs `main` (`4190bbb`) · mappers suite **426 passed / 1 failed** (that failure also fails on `main` — pre-existing, unrelated) · touched test files **170 passed** · `ruff check` + `format --check` clean · mypy inconclusive in my sandbox (2.3.1 internal-errors identically on `main` → environmental)
✅ #372 repro fixed: ~15 input shapes all return aware UTC (`main` returned naive for 4 of them)
✅ E2E symptom fixed: a mixed naive/aware openinference session yields 1 agent span vs **0 on `main`** — the `TypeError` was swallowed at `openinference_session_mapper.py:156-157`, so #372 was silently *dropping spans*, not just raising
✅ Serialized `SpanInfo` output byte-identical to `main` (`types/trace.py:88-97` already normalized on serialize) — this repairs the in-memory contract with no downstream shift

**Disclosure:** I filed #372 and left a reference patch on it; this PR is your own independent implementation, so I reviewed it against the *issue*, not that patch. Process note: the adversarial pass's deep tier timed out and was re-run narrower — everything it reports here is repro-confirmed.

### Should-fix

1. 🟡 **`src/strands_evals/mappers/session_mapper.py:82` — the docstring still documents the bug this PR removes** (`- datetime → passthrough`), and the new *naive-is-assumed-UTC* rule is documented nowhere — it's a lossy assumption (a source emitting local timestamps silently shifts by up to ±14h), exactly what a caller needs told. #372's stated ask was "making the docstring true", so this is the one gap that's part of the issue's letter. (Body-only because line 82 is outside the diff hunk.)

   ```diff
            Handles:
            - None → current UTC time
   -        - datetime → passthrough
   -        - ISO 8601 string (with optional trailing Z) → parsed datetime
   +        - datetime → converted to UTC
   +        - ISO 8601 string (with optional trailing Z) → converted to UTC
            - Numeric (int/float) nanosecond epoch → datetime
            - String-encoded nanosecond epoch → datetime
   +
   +        Naive datetimes and offset-less ISO strings are assumed to be UTC, not
   +        local time: a source emitting local timestamps is shifted by its offset.
   ```

2. 🟡 **Introduced regression: `astimezone` raises uncaught `OverflowError` for boundary instants, silently dropping the span** — inline comment with a verified suggestion.

3. 🟡 **The reported symptom is pinned by no test — a one-character edit to an existing e2e test fixes that** — inline comment on the new test file.

### Questions

❓ *(blocking only in the "is this intentionally not done yet?" sense)* This is a draft — is it scoped to the code fix on purpose, with the docstring (1) and symptom test (3) to follow before ready-for-review? If so, findings 1 and 3 are "planned", not "missed", and I'm happy to re-review when you flip it.

❓ *(non-blocking)* #372 mentioned "letting downstream defenses retire", and this PR keeps every `_to_aware_utc` call site (`extractors/trace_extractor.py:124,129`, `mappers/openai_agents_gen_ai_session_mapper.py:127`, `types/trace.py:147`). Keeping them looks right to me — they also guard user-constructed `Session`s that never pass through `parse_timestamp` — is "keep" the intended reading? Relatedly, the same naive→UTC normalization now lives in four places (`session_mapper.py:95-98`, `:105-108`, `types/trace.py:88-97`, `:128-132`); worth one shared helper, or better as a follow-up?

<details><summary>Appendix — non-blocking (6)</summary>

- ⚪ **Assertion strength** — `test_session_mapper.py:24, :51, :65, :77` assert only `tzinfo is not None`, which any aware result satisfies. `== timezone.utc` (as `:28`, `:34`, `:40` already do) is the one-word upgrade; today a fallback returning local-offset-aware time survives the whole suite (mutation-checked).
- ⚪ **Vacuous equality asserts** — `:45`, `:60`, `:72`: aware datetimes compare by *instant*, so these `==` lines pass even without normalization; the `tzinfo` assert above each carries the test. `result.isoformat() == "..."` would assert what's meant. (Related limit: `:48-53`/`:63-66` can't distinguish naive-assumed-UTC from naive-assumed-local on a UTC machine — needs a forced `TZ` + `tzset`; noted, not asked for.)
- ⚪ **`_ConcreteMapper.map_to_session(self, spans: list)`** (`test_session_mapper.py:12`) — the ABC signature is `(self, data: Any, session_id: str)` (`session_mapper.py:17`). Breaks nothing (ABCMeta matches names only; mypy never sees `tests/`), but the stub documents a contract no real mapper has.
- ⚪ **Orphaned banner** — `test_adk_otel_session_mapper.py:715-717`: the `# Tests: Timestamp Parsing` header survived the class it labelled.
- ⚪ **Two cheap missing cases** — nothing exercises the float/non-ns path (`session_mapper.py:113-115` false branch) or the unsupported-type fallback (`:116`); both already aware-UTC on `main`, just natural residents of the new file.
- ⚪ **Pre-existing `parse_timestamp` robustness gaps** (ms-epoch misparse, `isdigit()` outside the `try`, same unguarded `astimezone` in the serializer) — not this PR's problem; filed as #378.

</details>

<details><summary>Per-pass breakdown</summary>

| Pass | Verdict |
|---|---|
| triage | accept (routed: correctness, issue-alignment, adversarial, test-quality) |
| correctness/safety | approve — fix correct; one introduced regression (🟡 2) |
| issue-alignment | under-delivers slightly — 3 letter gaps (🟡 1–3), **zero scope creep**; test relocation matches AGENTS.md's mirror-the-src rule with coverage parity verified |
| test-quality | branch fixes genuinely pinned (reverting either fixed branch is caught by exactly the new tests); symptom not pinned (🟡 3); M-series mutation ledger behind the appendix items |
| adversarial | no regression class beyond the `OverflowError`; datetime subclasses, `bool`, sub-minute offsets, `fold`, and a 15-shape naive-return hunt all clean |

</details>

The fix itself is the right 7 lines and the branch-level tests are good — this is ready-for-ready-for-review once the three 🟡s land. A human maintainer still owns the approve.

liramon2:
@strandly-the-agent Review this PR.

strandly-the-agent (approved):
**All previous findings fixed and verified — approving.** Re-reviewed head `c827a26` (branch was rebased; delta vs my round-1 head `da5f29a` is exactly the review fixes, no new behavior). This round was a fixes-only delta, so I ran it as a single verification pass rather than the full pipeline.

**Status of round-1 findings:**
- ✅ 🟡 1 docstring — fixed: `Handles:` list updated and the naive-assumed-UTC caveat documented (`session_mapper.py:80-88`)
- ✅ 🟡 2 `OverflowError` regression — fixed exactly as suggested (`:98-104` guard, `:115` `except (ValueError, OverflowError)`); repros now degrade to aware `now(utc)`, and I mutation-checked the two new boundary tests: reverting the guards makes exactly those 2 fail
- ✅ 🟡 3 symptom test — fixed via the one-char edit (`test_openinference_session_mapper.py:2459`); verified it discriminates: with `main`'s `session_mapper.py` swapped in, that test fails with #372's exact symptom
- ✅ appendix (5 of 6): `tzinfo == timezone.utc` asserts, `_ConcreteMapper` signature now matches the ABC, orphaned banner removed, float-epoch + unsupported-type cases added
- ⏸️ appendix leftover, fine to leave: the `==`-by-instant asserts (`:46/:63/:75`) remain, but each now pairs with a `tzinfo == timezone.utc` assert, which was the substance

**Verified on `c827a26`:** mappers suite **430 passed / 1 failed** (`test_async_evaluation_with_langchain_mappers` — pre-existing, fails on `main` too) · `ruff check` + `format --check` clean · attacked the fix itself: boundary datetime/string, naive ISO, naive datetime all return aware UTC; the datetime-branch guard catching only `OverflowError` (vs the string branch's `(ValueError, OverflowError)`) is the shape I suggested and is correct — `ValueError` there needs a pathological custom tzinfo.

No new blockers, no open questions — the round-1 scope question is answered by the fixes themselves. Pre-existing robustness gaps remain tracked in #378. Ready to leave draft from my side; a human maintainer owns the merge.

opieter-aws (approved):

JackYPCOnline (approved):

PR #379, opened 2026-08-26
fix: skip non-serializable tools in OutputEvaluator serialization

### LINKED ISSUE #373: [BUG] OutputEvaluator(tools=...) breaks Experiment.to_file() / to_dict() — TypeError: DecoratedFunctionTool is not JSON serializable
### Checks
- [x] I have searched ./issues and there are no duplicates of my issue

### Strands Version
1.52.0

### Strands Evals Version
main @ cc0a391 (v1.0.3)

### Python Version
3.13.14

### Operating System
Linux (aarch64)

### Installation Method
git clone

### Steps to Reproduce
```python
from strands import tool

@tool
def my_tool() -> str:
    """A tool."""
    return "ok"

evaluator = OutputEvaluator(rubric="...", tools=[my_tool])
evaluator.to_dict()          # includes 'tools'
import json; json.dumps(evaluator.to_dict())   # boom
```
Any `Experiment` containing such an evaluator hits the same via `Experiment.to_file()` (`experiment.py:714`, `:747`).

### Expected Behavior
`to_dict()` / `Experiment.to_file()` produce JSON-serializable output for any supported evaluator configuration — `tools=` is a documented parameter since #324.

### Actual Behavior
`TypeError: Object of type DecoratedFunctionTool is not JSON serializable`.

### Additional Context
Root cause: `Evaluator.to_dict()` serializes every public non-default attribute (`evaluators/evaluator.py:322-336`, skips only `_`-prefixed). `OutputEvaluator` stores `self.tools` **publicly** (`output_evaluator.py:43`), while `TrajectoryEvaluator` stores `self._tools` (`trajectory_evaluator.py:43-50`) and escapes. The `Multimodal*` family follows `OutputEvaluator` and is affected too. Pre-existing since #324 (`c5461d0`); found while reviewing #343, whose README section recommends exactly this combination.

### Possible Solution
Store the attribute privately (`self._tools`) for consistency with `TrajectoryEvaluator`, or make `to_dict()` skip non-JSON-serializable attributes.

### Related Issues
#324, #343

poshinchen:
## Description

Fixes a `TypeError: Object of type DecoratedFunctionTool is not JSON serializable` when calling `to_dict()` on an `OutputEvaluator` configured with `tools=`, which also broke `Experiment.to_file()` / `to_dict()` for any experiment containing such an evaluator.

**Root cause:** `Evaluator.to_dict()` serializes every public non-default attribute. `OutputEvaluator` stored `self.tools` publicly, so callable tools (e.g. `@tool`-decorated functions) leaked into the serialized dict and crashed `json.dumps`. `TrajectoryEvaluator` already avoided this by storing `self._tools` privately.

**Fix:**
- `OutputEvaluator` now stores tools privately as `self._tools`, with a public `tools` property + setter so the documented API from #324 is unchanged (`evaluator.tools` reads and `evaluator.tools = [...]` writes as before).
- `OutputEvaluator.to_dict()` is overridden to re-add the JSON-serializable subset of tools (e.g. module path strings), so those still round-trip through `to_file()` / `from_file()`. Non-serializable tools are skipped with a `logger.warning` naming the tool (`tool_name` → `__name__` → type name, never `repr()`) and instructing how to re-attach it after loading. The check catches both `TypeError` and `ValueError` (circular references).
- Since callables cannot round-trip through JSON, users re-attach them after loading:

```python
loaded = Experiment.from_file("experiment.json")
for evaluator in loaded.evaluators:
    if isinstance(evaluator, OutputEvaluator):
        evaluator.tools = [*(evaluator.tools or []), verify_claim]
```

This one change also fixes the whole `Multimodal*` evaluator family (`MultimodalOutputEvaluator`, `MultimodalCorrectnessEvaluator`, `MultimodalFaithfulnessEvaluator`, `MultimodalInstructionFollowingEvaluator`, `MultimodalOverallQualityEvaluator`), since they all pass `tools` through `OutputEvaluator.__init__`.

**Not a breaking change:** the constructor, `evaluator.tools` reads/writes, and loading of previously saved JSON files (including ones with a `tools` key) all behave as before. Serializable tools still round-trip; callable tools previously crashed serialization, so nothing working is lost. One exactness note: `tools=[]` used to serialize as `"tools": []` and now the key is omitted, so it reloads as `None` — behaviorally inert since both mean "no tools" to the evaluator agent.

**Accepted tradeoff:** the skip warning fires at save time, not load time — a reloaded evaluator runs without its callable tools and only the save-side log signals it. Warning at load time isn't structurally possible today because `from_dict` passes every JSON key as a constructor kwarg, so a marker key would raise `TypeError`. Deliberate, documented in the `tools` property docstring.

## Related Issues

Fixes #373
Related: #324, #343

## Documentation PR

N/A — behavior documented in docstrings; no docs changes needed.

## Type of Change

Bug fix

## Testing

- Added regression tests: callable-only tools produce JSON-serializable `to_dict()` output with no `tools` key plus a warning; mixed tools keep serializable entries and skip callables with exactly one warning naming the dropped tool; circular-reference tools are skipped instead of crashing; `tools` setter works after construction.
- Added an `Experiment`-level regression test pinning #373 where it was reported: `to_file()` → `from_file()` with a real `@tool`-decorated function plus a module path string asserts the string tool survives the round-trip.
- Ran the issue's exact repro: `json.dumps(evaluator.to_dict())` now succeeds while `evaluator.tools` remains usable by `evaluate()` / `evaluate_async()`.
- `hatch test` passes across the output, multimodal, base evaluator, and experiment suites; `hatch fmt --check` (ruff, ruff format, mypy) clean.

## Checklist

----

### DIFF
--- a/src/strands_evals/evaluators/output_evaluator.py
+++ b/src/strands_evals/evaluators/output_evaluator.py
@@ -77,9 +77,11 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
                 try:
                     json.dumps(tool)
                 except (TypeError, ValueError):
-                    tool_name = getattr(tool, "tool_name", None) or repr(tool)
+                    tool_name = (
+                        getattr(tool, "tool_name", None) or getattr(tool, "__name__", None) or type(tool).__name__
+                    )
                     logger.warning(
-                        "skipping non-JSON-serializable tool <%s> during serialization; "
+                        "tool_name=<%s> | skipping non-JSON-serializable tool during serialization, "
                         "re-attach it via the `tools` attribute after loading",
                         tool_name,
                     )
--- a/src/strands_evals/evaluators/output_evaluator.py
+++ b/src/strands_evals/evaluators/output_evaluator.py
@@ -1,3 +1,5 @@
+import json
+import logging
 from typing import Any, cast
 
 from strands import Agent
@@ -7,6 +9,8 @@ from ..types.evaluation import EvaluationData, EvaluationOutput, InputT, OutputT
 from .evaluator import Evaluator
 from .prompt_templates.case_prompt_template import compose_test_prompt
 from .prompt_templates.prompt_templates import judge_output_template as SYSTEM_PROMPT
+
+logger = logging.getLogger(__name__)
 
 
 class OutputEvaluator(Evaluator[InputT, OutputT]):
@@ -40,7 +44,50 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
         self.include_inputs = include_inputs
         self.system_prompt = system_prompt
         self.uses_environment_state = uses_environment_state
-        self.tools = tools
+        # Stored privately so the base to_dict() skips it; to_dict() below re-adds
+        # the JSON-serializable subset so tools like module path strings round-trip.
+        self._tools = tools
+
+    @property
+    def tools(self) -> list[Any] | None:
+        """Optional tools for the evaluator agent.
+
+        Only JSON-serializable entries (e.g. module path strings) survive `to_dict()`;
+        callables are skipped with a warning and must be re-attached after `from_dict()`.
+        """
+        return self._tools
+
+    @tools.setter
+    def tools(self, value: list[Any] | None) -> None:
+        self._tools = value
+
+    def to_dict(self) -> dict:
+        """
+        Convert the evaluator into a dictionary.
+
+        Returns:
+            dict: A dictionary containing the evaluator's information. Includes only the
+            JSON-serializable tools; non-serializable tools (e.g. decorated functions)
+            are skipped with a warning since `from_dict()` cannot reconstruct them.
+        """
+        _dict = super().to_dict()
+        if self._tools:
+            serializable_tools = []
+            for tool in self._tools:
+                try:
+                    json.dumps(tool)
+                except (TypeError, ValueError):
+                    tool_name = getattr(tool, "tool_name", None) or repr(tool)
+                    logger.warning(
+                        "skipping non-JSON-serializable tool <%s> during serialization; "
+                        "re-attach it via the `tools` attribute after loading",
+                        tool_name,
+                    )
+                else:
+                    serializable_tools.append(tool)
+            if serializable_tools:
+                _dict["tools"] = serializable_tools
+        return _dict
 
     def _build_prompt(self, evaluation_case: EvaluationData[InputT, OutputT]) -> str | list:
         """Build the evaluation prompt for a test case.
--- a/tests/strands_evals/evaluators/test_output_evaluator.py
+++ b/tests/strands_evals/evaluators/test_output_evaluator.py
@@ -1,3 +1,5 @@
+import json
+import logging
 from unittest.mock import Mock, patch
 
 import pytest
@@ -281,6 +283,63 @@ def test_output_evaluator_init_without_tools_defaults_to_none():
     evaluator = OutputEvaluator(rubric="Test rubric")
 
     assert evaluator.tools is None
+
+
+def test_output_evaluator_to_dict_skips_non_serializable_tools(caplog):
+    """Test that to_dict() output is JSON-serializable when callable tools are set (issue #373)"""
+
+    def verify_claim(claim: str) -> str:
+        return "verified"
+
+    evaluator = OutputEvaluator(rubric="Test rubric", tools=[verify_claim])
+    with caplog.at_level(logging.WARNING):
+        evaluator_dict = evaluator.to_dict()
+
+    assert "tools" not in evaluator_dict
+    json.dumps(evaluator_dict)
+    assert "skipping non-JSON-serializable tool" in caplog.text
+
+
+def test_output_evaluator_to_dict_keeps_serializable_tools(caplog):
+    """Test that serializable tools survive to_dict() while callables are skipped"""
+
+    def verify_claim(claim: str) -> str:
+        return "verified"
+
+    evaluator = OutputEvaluator(rubric="Test rubric", tools=["my_pkg.calculator", verify_claim])
+    with caplog.at_level(logging.WARNING):
+        evaluator_dict = evaluator.to_dict()
+
+    assert evaluator_dict["tools"] == ["my_pkg.calculator"]
+    json.dumps(evaluator_dict)
+    assert "skipping non-JSON-serializable tool" in caplog.text
+
+
+def test_output_evaluator_to_dict_skips_circular_reference_tools(caplog):
+    """Test that tools raising ValueError (circular reference) are skipped, not crashed on"""
+    circular: dict = {"name": "circular_tool"}
+    circular["self"] = circular
+
+    evaluator = OutputEvaluator(rubric="Test rubric", tools=[circular, "my_pkg.calculator"])
+    with caplog.at_level(logging.WARNING):
+        evaluator_dict = evaluator.to_dict()
+
+    assert evaluator_dict["tools"] == ["my_pkg.calculator"]
+    json.dumps(evaluator_dict)
+    assert "skipping non-JSON-serializable tool" in caplog.text
+
+
+def test_output_evaluator_tools_setter():
+    """Test that tools can be reassigned after initialization"""
+
+    def verify_claim(claim: str) -> str:
+        return "verified"
+
+    evaluator = OutputEvaluator(rubric="Test rubric")
+    evaluator.tools = [verify_claim]
+
+    assert evaluator.tools == [verify_claim]
+    assert "tools" not in evaluator.to_dict()
 
 
 @patch("strands_evals.evaluators.output_evaluator.Agent")
--- a/tests/strands_evals/evaluators/test_output_evaluator.py
+++ b/tests/strands_evals/evaluators/test_output_evaluator.py
@@ -313,6 +313,9 @@ def test_output_evaluator_to_dict_keeps_serializable_tools(caplog):
     assert evaluator_dict["tools"] == ["my_pkg.calculator"]
     json.dumps(evaluator_dict)
     assert "skipping non-JSON-serializable tool" in caplog.text
+    warnings = [record for record in caplog.records if record.levelno == logging.WARNING]
+    assert len(warnings) == 1  # the serializable tool must not warn
+    assert "verify_claim" in warnings[0].getMessage()  # names which tool to re-attach
 
 
 def test_output_evaluator_to_dict_skips_circular_reference_tools(caplog):
--- a/tests/strands_evals/test_experiment.py
+++ b/tests/strands_evals/test_experiment.py
@@ -6,6 +6,7 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 from botocore.exceptions import ClientError
+from strands import tool as strands_tool
 from strands.models.model import Model
 from strands.types.exceptions import EventLoopException, ModelThrottledException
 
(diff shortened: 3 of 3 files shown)


poshinchen:
@strandly-the-agent review this PR

src/strands_evals/evaluators/output_evaluator.py
```diff
@@ -40,7 +44,50 @@ def __init__(
...
+            for tool in self._tools:
+                try:
+                    json.dumps(tool)
+                except (TypeError, ValueError):
+                    tool_name = getattr(tool, "tool_name", None) or repr(tool)
+                    logger.warning(
+                        "skipping non-JSON-serializable tool <%s> during serialization; "
+                        "re-attach it via the `tools` attribute after loading",
+                        tool_name,
+                    )
```
strandly-the-agent: **🟡 The skip path can leak secrets, emit a 500 KB log line, and crash — one hunk fixes all three**

`getattr(tool, "tool_name", None) or repr(tool)` only has a clean path for `@tool`-decorated tools (they have `tool_name`). Everything else `Agent` accepts — plain functions, `functools.partial`, dict tool specs — falls through to `repr()`. All three reproduced on this branch:

1. **Credential disclosure.** `tools=[functools.partial(call_api, api_key="sk-...")]` logs the key at WARNING: `skipping non-JSON-serializable tool <functools.partial(<function call_api at 0x…>, api_key='sk-SECRET123')>`.
2. **Unbounded log line.** A dict tool spec containing a callable produced a single **520,185-char** warning.
3. **The crash class this PR removes, surviving inside the fix.** `getattr`'s default only swallows `AttributeError`, and `repr()` is called bare — a tool whose `__repr__` raises (detached ORM instance, lazy proxy, misbuilt mock) raises straight out of this line, and since `to_file` truncates the destination before dumping (`experiment.py:761-762`, filed #382), a pre-existing 488-byte experiment file ended up **0 bytes**.

```suggestion
                    tool_name = (
                        getattr(tool, "tool_name", None) or getattr(tool, "__name__", None) or type(tool).__name__
                    )
                    logger.warning(
                        "tool_name=<%s> | skipping non-JSON-serializable tool during serialization, "
                        "re-attach it via the `tools` attribute after loading",
                        tool_name,
                    )
```

Verified with this applied: secret gone, log line 520,185 → 133 chars, the `__repr__`-raises case no longer crashes, the existing caplog assertions still match, 24/24 pass, `ruff check` + `ruff format --check` + `mypy` clean. It also lands the line on `STYLE_GUIDE.md`'s log format (`field=<%s> | lowercase message`) — today's `<%s>` around a `repr` renders doubled brackets (`<<function verify_claim at 0x…>>`) with a nondeterministic address.

One residual it doesn't cover: a `tool_name` *property* that raises a non-`AttributeError`. Contrived enough that I wouldn't add a `try/except` for it.

tests/strands_evals/evaluators/test_output_evaluator.py
```diff
@@ -283,6 +285,63 @@ def test_output_evaluator_init_without_tools_defaults_to_none():
...
+    def verify_claim(claim: str) -> str:
+        return "verified"
+
+    evaluator = OutputEvaluator(rubric="Test rubric", tools=["my_pkg.calculator", verify_claim])
+    with caplog.at_level(logging.WARNING):
+        evaluator_dict = evaluator.to_dict()
+
+    assert evaluator_dict["tools"] == ["my_pkg.calculator"]
+    json.dumps(evaluator_dict)
+    assert "skipping non-JSON-serializable tool" in caplog.text
```
strandly-the-agent, line 315: **🟡 The regression isn't pinned at the level #373 reports it**

#373 is filed against `Experiment.to_file()`, but the new tests stop at `OutputEvaluator.to_dict()` — `test_experiment.py:322-351` still constructs evaluators without tools, and the parametrized round-trip at `:783-796` never exercises them. Nothing fails if the `tools` key stops surviving `to_file()` → `from_file()`, and the tests validate with default `json.dumps` while `to_file` writes `ensure_ascii=False` (`experiment.py:762`) — which is exactly how the probe/writer mismatch in #380 stayed invisible.

Mutation-testing the four new tests also left two live mutants: deleting the `getattr(tool, "tool_name", …)` lookup survives (assertions stop at the message prefix), and warning for *every* tool — serializable ones included — survives. Strengthening this test closes both:

```suggestion
    assert evaluator_dict["tools"] == ["my_pkg.calculator"]
    json.dumps(evaluator_dict)
    assert "skipping non-JSON-serializable tool" in caplog.text
    warnings = [r for r in caplog.records if r.levelno == logging.WARNING]
    assert len(warnings) == 1  # the serializable tool must not warn
    assert "verify_claim" in warnings[0].getMessage()  # names which tool to re-attach
```

…plus one `tmp_path` test doing `Experiment.to_file()` → `from_file()` with `tools=[verify_claim, "my_pkg.calculator"]`, asserting the string tool comes back (and ideally using an actual `@tool`-decorated function — `DecoratedFunctionTool` is the type in #373's title, and `strands.tool` imports fine in this test env). Prior art for covering the `Multimodal*` subclasses in the same stroke: `test_multimodal_specialized_evaluators.py:243-270`.

strandly-the-agent (commented):
**Correct, complete fix for #373 — no blockers.** Two 🟡 should-fixes inline: the skip path's `repr()` fallback can leak a credential into logs, emit a 500 KB log line, or itself crash; and the regression isn't pinned at the `Experiment.to_file()` level where #373 was reported. Both are one hunk each.

**✅ Verified** (branch `pr-379` @ `0f364f3`): #373 reproduced on `main`, fixed here at **both** levels — `json.dumps(evaluator.to_dict())` and `Experiment.to_file()` · string tools round-trip through `to_file()` → `from_file()`, incl. the `Multimodal*` subclasses and legacy JSON carrying a `tools` key · `to_dict()` idempotent, no interaction bug with the base class's defaults logic · 24/24 in `test_output_evaluator.py`, 509 across `tests/strands_evals/evaluators` + `test_experiment.py` · mutation-checked: all 4 new tests fail on `main` and each kills a distinct mutant (2 surviving mutants → inline comment 2) · `ruff` + `ruff format` + `mypy` clean, including with the suggested log fix applied.

### Questions

1. **(worth one sentence before merge) Is `logger.warning` the intended channel, and is silence at load time an accepted tradeoff?** The warning fires at save time, but the risk lands at load time on a possibly different machine: a reloaded evaluator runs the judge *without* its verification tools and produces drifting scores with no signal. It's structurally hard to do better today — `from_dict` passes every key as a ctor kwarg (`experiment.py:818`), so a marker key would `TypeError` — so this may well be the right call; asking so it's deliberate. Related prior art in-repo: `warnings.warn(..., stacklevel=2)` for the analogous "config dropped" case (`simulation/tool_simulator.py:70-72`), which surfaces in pytest summaries and can be escalated with `-W error`.
2. *(non-blocking)* `TrajectoryEvaluator` now has the opposite contract (no public `.tools`, silently drops even serializable user tools — the evaluator #324's own example used). Agree it's out of scope here since untangling the built-in scorers from `_tools` is its own design task — filed #381 to track; worth narrowing the PR title to the `OutputEvaluator` family.
3. *(non-blocking)* One PR-body claim isn't exact: `tools=[]` serialized as `"tools": []` on `main`, is now omitted and reloads as `None` (verified both sides). Behaviorally inert — a sentence in the body, or `if self._tools is not None:` at line 74, closes it.

<details>
<summary>Appendix — non-blocking (7)</summary>

- ⚪ `tools=[]` → key omitted (line 74 truthiness guard); no behavioral consequence, and emitting `[]` when everything was *dropped* would mislead, so as-is is defensible (see Question 3).
- ⚪ Non-list `tools` is now silently exploded item-by-item: `evaluator.tools = "my_pkg.calculator"` persists as 17 single characters, where `main` round-tripped it verbatim (visible and debuggable). Requires violating the declared `list[Any] | None` annotation, so contrived — a one-line `isinstance` guard in the setter if you want it.
- ⚪ Class-level `Attributes` entry for `tools` (lines 27-28) doesn't mention the persistence caveat; the property docstring does, but the class docstring renders first. One line: "callable tools are not persisted by `to_dict()`; re-attach after `from_file()`."
- ⚪ `Multimodal*` inheritance works (verified end-to-end) but nothing guards it — parametrizing one new test over `MultimodalCorrectnessEvaluator` is a one-liner; prior art at `test_multimodal_specialized_evaluators.py:243-270`.
- ⚪ **pre-existing, filed #380:** the probe uses `json.dumps` defaults while `to_file` writes `ensure_ascii=False` — a string tool with an unpaired surrogate passes the probe then crashes `to_file` (`UnicodeEncodeError`, truncated file), and `NaN`/`Infinity` pass the probe and produce RFC-8259-invalid JSON. Cheap fold-in if you want it now: probe with `json.dumps(tool, ensure_ascii=False, allow_nan=False).encode("utf-8")`.
- ⚪ **pre-existing, filed #382:** `to_file` truncates the destination before serializing (`experiment.py:761-762`), so any serialization error destroys the previous file (observed 488 bytes → 0).
- ⚪ **pre-existing, contrived:** int/float dict keys in a tool spec come back as strings on reload; `{1: "a", "1": "b"}` collides and loses a value.

Repo note: no `needs-api-review` label exists here (closest: `design`, `area-devx`), so no label gate applies to the new public `tools` property. My label permissions don't cover the filed issues — suggest `bug` for #380/#382 and `design`/`area-devx` for #381.
</details>

<details>
<summary>Review process</summary>

Full pipeline, first review of this PR: 5 independent fresh-context passes (correctness, issue-alignment, API/DevX, adversarial, test-quality) + aggregation. The first adversarial run failed to return output and was re-run successfully. Alternatives weighed and rejected by the API pass: fully-private `_tools` with no persistence (regresses the string-tool round-trip that worked on `main`), strict raise-on-callable (keeps `to_file` unusable for the exact pattern #324 advertises) — the PR's skip-with-warning shape is the right one of the three. Deliberately not raised: `RecursionError` escaping the `except` (it doesn't — verified to 2000 deep), mutation aliasing of the returned dict (this PR *improves* on `main`), tuples-in-tools (reload as lists, which `ToolRegistry.process_tools` flattens identically).

All findings above were independently re-verified before posting; work reviewed by an AI (`strandly-the-agent`) — human judgment owns the merge.
</details>

poshinchen:
@strandly-the-agent I've updated the PR.

strandly-the-agent (approved):
**All previous findings resolved at `008545d` — LGTM.** Follow-up review (self-run verification of the delta, not a fresh full pipeline, since the delta only addresses my findings); every check below re-run cold on this head.

**Previous findings → status**
- 🟡 `repr()` fallback (secrets / 500 KB lines / crash) → **fixed** — suggestion applied; re-ran all three repros: `functools.partial(api_key=...)` logs `tool_name=<partial>` with no secret, the huge-dict spec warning is 129 chars (was 520,185), and a tool whose `__repr__` raises now serializes fine (`tool_name=<Detached>`).
- 🟡 regression not pinned at the `Experiment.to_file()` level → **fixed** — `test_experiment_to_file_round_trip_with_evaluator_tools` uses a real `@tool`-decorated function (the `DecoratedFunctionTool` from #373's title, exercising the `tool_name` branch) + a module-path string through `to_file()` → `from_file()`, and the strengthened caplog assertions pin exactly one warning naming the dropped tool.
- Q1 (warning channel / silent-at-load) → **answered** — documented as an accepted tradeoff in the PR body with the structural reason; works for me.
- Q2 (title scope / TrajectoryEvaluator) → **addressed** — title narrowed; #381 tracks the sibling.
- Q3 (`tools=[]` exactness) → **addressed** — disclosed in the PR body.

**✅ Verified at `008545d`:** 164 passed (`test_output_evaluator` + `test_experiment` + `test_multimodal_specialized_evaluators`) · `ruff check` clean · `mypy` clean · attacked the fix itself: falsy/absent `tool_name` falls through the new chain correctly (`__name__` → type name), plain-function and decorated-tool naming branches both covered by tests. Only residual is the one already agreed as contrived (a `tool_name` property raising non-`AttributeError`).

Nit, not blocking: the "I ran `hatch run prepare`" checklist box got unchecked in the body edit, though the Testing section says the equivalent gates ran.

opieter-aws (approved):

poshinchen:
@strandly-the-agent will this solve https://github.com/strands-agents/evals/issues/380

strandly-the-agent:
No — both #380 cases still reproduce on this head (`008545d`, just re-ran them): a string tool with an unpaired surrogate passes the probe and `to_file()` still crashes `UnicodeEncodeError`, and `NaN` still passes and writes RFC-invalid JSON.

The reason: this PR's probe is `json.dumps(tool)` with defaults (`ensure_ascii=True`, escaping surrogates; `allow_nan=True`), while `Experiment.to_file` writes with `ensure_ascii=False` into utf-8 (`experiment.py:762`) — so the probe accepts strictly more than the writer can emit.

If you want to fold #380's evaluator-side half into this PR, it's a one-line probe change:

```python
json.dumps(tool, ensure_ascii=False, allow_nan=False).encode("utf-8")
```

(the `.encode` catches surrogates, `allow_nan=False` catches NaN/Infinity — both then take the existing skip-with-warning path). That said, #380 is broader than this PR: any non-tool value with the same properties hits `to_file` too, and the truncate-on-error data loss is #382 — so keeping them as follow-ups is also fine. Happy to send a PR for #380/#382 if useful.

Issue #382, opened 2026-08-26
[BUG] Experiment.to_file() truncates the destination before serializing — a serialization error destroys the existing file


strandly-the-agent:
### Checks

- [x] I have updated to the latest minor and patch version of Strands and evals
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.53.0

### Strands Evals Version

main @ 9b245cf (1.2.1.dev)

### Python Version

3.13.14

### Operating System

Amazon Linux 2023 (aarch64)

### Installation Method

git clone

### Steps to Reproduce

`Experiment.to_file()` opens the destination with mode `"w"` (truncating it) before serialization completes (`src/strands_evals/experiment.py:761-762`):

```python
with open(file_path, "w", encoding="utf-8") as f:
    json.dump(self.to_dict(), f, indent=2, ensure_ascii=False)
```

So any error raised by `self.to_dict()` or mid-`json.dump` destroys whatever was previously at that path:

```python
from strands_evals import Case, Experiment
from strands_evals.evaluators import OutputEvaluator

# a previously saved, valid experiment file
with open("exp.json", "w") as f:
    f.write("x" * 488)  # stand-in for real content

exp = Experiment(
    cases=[Case(input="q")],
    evaluators=[OutputEvaluator(rubric="r", tools=[lambda x: x])],  # not JSON-serializable on main
)
exp.to_file("exp.json")  # raises
```

### Expected Behavior

A failed save leaves the existing file intact (serialize first, then write — ideally atomically via temp file + rename).

### Actual Behavior

The pre-existing 488-byte file is destroyed. Observed both failure shapes:

- `json.dump` raises mid-stream (`TypeError: Object of type function is not JSON serializable` on main) → file replaced by partial, invalid JSON (503 bytes).
- `to_dict()` itself raises (arguments are evaluated after `open("w")` has truncated) → file left at **0 bytes**.

### Additional Context

Found during review of #379. Even after #379 removes the most common `to_dict()` crash, any remaining serialization error (see #380 for two) still destroys the previous artifact — which for an evals product is saved experiment data.

### Possible Solution

```python
content = json.dumps(self.to_dict(), indent=2, ensure_ascii=False)
with open(file_path, "w", encoding="utf-8") as f:
    f.write(content)
```

or a temp-file + `os.replace()` for atomicity.

### Related Issues

#379, #380, #373

PR #383, opened 2026-08-27
fix: make Experiment.to_file() reject non-strict JSON instead of writing invalid files

### LINKED ISSUE #380: [BUG] Experiment.to_file() crashes on strings with unpaired surrogates and writes RFC-8259-invalid JSON for NaN/Infinity
### Checks

### Strands Version

1.53.0

### Strands Evals Version

main @ 9b245cf (1.2.1.dev)

### Python Version

3.13.14

### Operating System

Amazon Linux 2023 (aarch64)

### Installation Method

git clone

### Steps to Reproduce

`Experiment.to_file()` writes with `json.dump(..., ensure_ascii=False)` into a utf-8 file (`src/strands_evals/experiment.py:761-762`), which accepts strictly less than `json.dumps()` with defaults — so values that survive `to_dict()`/`json.dumps()` can still crash or corrupt the file write.

**Case 1 — unpaired surrogate in a string (e.g. an `os.fsdecode`'d / surrogateescape'd file path used as a tool path):**

```python

exp = Experiment(
    cases=[Case(input="q")],
    evaluators=[OutputEvaluator(rubric="r", tools=["my_pkg.calc_\udcff"])],
)
exp.to_dict()            # fine
exp.to_file("exp.json")  # UnicodeEncodeError
```

**Case 2 — NaN/Infinity anywhere in the serialized dict (e.g. a float in a dict tool spec):**

```python
exp = Experiment(
    cases=[Case(input="q")],
    evaluators=[OutputEvaluator(rubric="r", tools=[{"name": "t", "default": float("nan")}])],
)
exp.to_file("exp.json")  # "succeeds"
```

### Expected Behavior

Either the value is rejected up front with a clear error, or `to_file()` produces a valid, loadable JSON file.

### Actual Behavior

- Case 1: `UnicodeEncodeError: 'utf-8' codec can't encode character '\udcff' ... surrogates not allowed`, and the destination file is left as truncated, invalid JSON (observed: 493 bytes ending `"tools": `; `json.load` on it raises `JSONDecodeError`).
- Case 2: the file contains the literal `NaN` — not valid JSON per RFC 8259. Python's `from_file` reloads it (default `parse_constant`), but any strict parser (JS `JSON.parse`, `jq`, serde, ...) rejects the entire experiment file.

Both reproduced on `main` @ 9b245cf; #379's new per-tool serializability probe doesn't catch either, because it probes with `json.dumps(tool)` defaults (`ensure_ascii=True`, `allow_nan=True`) while the writer uses different settings.

### Additional Context

Found during review of #379. Reachability is low (surrogate-escaped path strings, NaN inside a tool-spec dict), but the failure shape is bad: a crashed or silently-invalid persistence artifact.

### Possible Solution

Probe with writer parity: `json.dumps(value, ensure_ascii=False, allow_nan=False).encode("utf-8")` (the `.encode` catches surrogates, `allow_nan=False` catches NaN/Infinity) — e.g. in `OutputEvaluator.to_dict()`'s probe after #379, and/or pass `allow_nan=False` to the `json.dump` in `Experiment.to_file()` so it fails loudly instead of writing invalid JSON. Related: the truncate-on-open data-loss aspect is tracked separately.

### Related Issues

#379, #373

poshinchen:
## Description

`Experiment.to_file()` writes JSON with `ensure_ascii=False` into a utf-8 file. The tool serialization check in `OutputEvaluator.to_dict()` (added in #379) used `json.dumps(tool)` with default settings, which accept some values the file writer cannot handle. Two kinds of values passed the check and still broke the file:

- A string with an unpaired surrogate character (for example a file path decoded with `surrogateescape`) crashed the write with `UnicodeEncodeError` and left behind a truncated, invalid JSON file.
- A `NaN` or `Infinity` float was written as the literal text `NaN`, which is not valid JSON and is rejected by strict parsers such as `jq` and JavaScript's `JSON.parse`.

### Changes

- `OutputEvaluator.to_dict()` now checks each tool with the same settings the file writer uses: `json.dumps(tool, ensure_ascii=False, allow_nan=False).encode("utf-8")`. Tools containing unpaired surrogates or NaN/Infinity are skipped with the existing warning instead of breaking the file later. No new exception handling was needed because `UnicodeEncodeError` is a subclass of `ValueError`.
- `Experiment.to_file()` now converts the experiment to JSON bytes before opening the file, and passes `allow_nan=False`. Bad values raise `ValueError` before the file is touched, so a failure can no longer truncate or corrupt an existing file.

## Related Issues

Fixes #380

## Documentation PR

N/A (the `to_file()` docstring is updated in this PR)

## Type of Change

Bug fix

## Testing

- [x] I ran `hatch run prepare` (lint, format, and the full py3.10–3.14 test matrix pass)
- Added 4 unit tests: surrogate and NaN tools are skipped by `to_dict()`, and `to_file()` raises `ValueError` on bad case data while leaving the destination file untouched.
- Ran both examples from #380. Each now produces a valid, loadable JSON file instead of a crash or a file containing `NaN`.

## Checklist
- [ ] I have read the CONTRIBUTING document
- [ ] I have reviewed and understand every line of code in this PR, including any generated by AI tools, and I can explain why it works

----

### DIFF
--- a/src/strands_evals/evaluators/output_evaluator.py
+++ b/src/strands_evals/evaluators/output_evaluator.py
@@ -52,8 +52,10 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
     def tools(self) -> list[Any] | None:
         """Optional tools for the evaluator agent.
 
-        Only JSON-serializable entries (e.g. module path strings) survive `to_dict()`;
-        callables are skipped with a warning and must be re-attached after `from_dict()`.
+        Only tools that can be written as valid JSON in a utf-8 file survive `to_dict()`,
+        for example module path strings. Tools that cannot (callables, NaN or Infinity
+        floats, strings with unpaired surrogates) are skipped with a warning and must be
+        re-attached after `from_dict()`.
         """
         return self._tools
 
@@ -66,25 +68,31 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
         Convert the evaluator into a dictionary.
 
         Returns:
-            dict: A dictionary containing the evaluator's information. Includes only the
-            JSON-serializable tools; non-serializable tools (e.g. decorated functions)
-            are skipped with a warning since `from_dict()` cannot reconstruct them.
+            dict: A dictionary containing the evaluator's information. Includes only tools
+            that can be written as valid JSON in a utf-8 file. Tools that cannot (decorated
+            functions, NaN or Infinity floats, strings with unpaired surrogates) are skipped
+            with a warning and must be re-attached after `from_dict()`.
         """
         _dict = super().to_dict()
         if self._tools:
             serializable_tools = []
             for tool in self._tools:
                 try:
-                    # Probe with the same strictness as Experiment.to_file()'s utf-8 writer:
-                    # the encode rejects unpaired surrogates, allow_nan=False rejects
-                    # NaN/Infinity (invalid per RFC 8259). UnicodeEncodeError is a ValueError.
+                    # Check each tool with the same settings Experiment.to_file() uses.
+                    # The utf-8 encode rejects unpaired surrogates. allow_nan=False
+                    # rejects NaN and Infinity, which are not allowed in valid JSON.
+                    # UnicodeEncodeError is a subclass of ValueError.
                     json.dumps(tool, ensure_ascii=False, allow_nan=False).encode("utf-8")
                 except (TypeError, ValueError):
-                    tool_name = (
-                        getattr(tool, "tool_name", None) or getattr(tool, "__name__", None) or type(tool).__name__
-                    )
+                    tool_name = getattr(tool, "tool_name", None) or getattr(tool, "__name__", None)
+                    if not isinstance(tool_name, str):
+                        # The tool may be a plain string or a dict with no name attribute.
+                        # A short ascii() preview identifies it better than a type name.
+                        tool_name = ascii(tool)
+                        if len(tool_name) > 80:
+                            tool_name = tool_name[:77] + "..."
                     logger.warning(
-                        "tool_name=<%s> | skipping non-JSON-serializable tool during serialization, "
+                        "tool_name=<%s> | skipping tool that cannot be written as valid utf-8 JSON, "
                         "re-attach it via the `tools` attribute after loading",
                         tool_name,
                     )
--- a/src/strands_evals/evaluators/output_evaluator.py
+++ b/src/strands_evals/evaluators/output_evaluator.py
@@ -75,7 +75,10 @@ class OutputEvaluator(Evaluator[InputT, OutputT]):
             serializable_tools = []
             for tool in self._tools:
                 try:
-                    json.dumps(tool)
+                    # Probe with the same strictness as Experiment.to_file()'s utf-8 writer:
+                    # the encode rejects unpaired surrogates, allow_nan=False rejects
+                    # NaN/Infinity (invalid per RFC 8259). UnicodeEncodeError is a ValueError.
+                    json.dumps(tool, ensure_ascii=False, allow_nan=False).encode("utf-8")
                 except (TypeError, ValueError):
                     tool_name = (
                         getattr(tool, "tool_name", None) or getattr(tool, "__name__", None) or type(tool).__name__
--- a/src/strands_evals/experiment.py
+++ b/src/strands_evals/experiment.py
@@ -744,8 +744,10 @@ class Experiment(Generic[InputT, OutputT]):
 
         Raises:
             ValueError: If the path has a non-JSON extension, or if the experiment contains
-                values that cannot be represented as strict JSON (e.g. NaN/Infinity floats
-                or strings with unpaired surrogates).
+                values that are not allowed in valid JSON, such as NaN or Infinity floats,
+                or strings with unpaired surrogates.
+            TypeError: If the experiment contains objects that are not JSON-serializable,
+                such as a custom class instance in case data or metadata.
         """
         file_path = Path(path)
 
@@ -758,15 +760,22 @@ class Experiment(Generic[InputT, OutputT]):
         else:
             file_path = file_path.with_suffix(".json")
 
+        # Serialize before touching the file. If the data cannot become valid JSON,
+        # the error is raised here and any existing file stays intact.
+        experiment_dict = self.to_dict()
+        try:
+            data = json.dumps(experiment_dict, indent=2, ensure_ascii=False, allow_nan=False).encode("utf-8")
+        except ValueError as e:
+            # UnicodeEncodeError is a subclass of ValueError, so this also catches
+            # unpaired surrogates. The original error stays attached as __cause__.
+            raise ValueError(
+                f"Cannot write experiment to {file_path}: it contains values that are not "
+                f"allowed in valid JSON, such as NaN or Infinity floats, or strings with "
+                f"unpaired surrogates. Check the case inputs, outputs, metadata, and evaluator fields."
+            ) from e
+
         file_path.parent.mkdir(parents=True, exist_ok=True)
-
-        # Serialize and encode before opening the file so a failure (NaN/Infinity via
-        # allow_nan=False, unpaired surrogates via the utf-8 encode) raises loudly
-        # without leaving a truncated, invalid JSON artifact behind.
-        data = json.dumps(self.to_dict(), indent=2, ensure_ascii=False, allow_nan=False).encode("utf-8")
-
-        with open(file_path, "wb") as f:
-            f.write(data)
+        file_path.write_bytes(data)
 
     @classmethod
     def from_dict(cls, data: dict, custom_evaluators: list[type[Evaluator]] | None = None):
--- a/src/strands_evals/experiment.py
+++ b/src/strands_evals/experiment.py
@@ -743,7 +743,9 @@ class Experiment(Generic[InputT, OutputT]):
                   Only .json format is supported.
 
         Raises:
-            ValueError: If the path has a non-JSON extension.
+            ValueError: If the path has a non-JSON extension, or if the experiment contains
+                values that cannot be represented as strict JSON (e.g. NaN/Infinity floats
+                or strings with unpaired surrogates).
         """
         file_path = Path(path)
 
@@ -758,8 +760,13 @@ class Experiment(Generic[InputT, OutputT]):
 
         file_path.parent.mkdir(parents=True, exist_ok=True)
 
-        with open(file_path, "w", encoding="utf-8") as f:
-            json.dump(self.to_dict(), f, indent=2, ensure_ascii=False)
+        # Serialize and encode before opening the file so a failure (NaN/Infinity via
+        # allow_nan=False, unpaired surrogates via the utf-8 encode) raises loudly
+        # without leaving a truncated, invalid JSON artifact behind.
+        data = json.dumps(self.to_dict(), indent=2, ensure_ascii=False, allow_nan=False).encode("utf-8")
+
+        with open(file_path, "wb") as f:
+            f.write(data)
 
     @classmethod
     def from_dict(cls, data: dict, custom_evaluators: list[type[Evaluator]] | None = None):
--- a/tests/strands_evals/test_experiment.py
+++ b/tests/strands_evals/test_experiment.py
@@ -819,6 +819,32 @@ def test_experiment_to_file_round_trip_with_evaluator_tools(tmp_path):
     assert loaded.evaluators[0].tools == ["my_pkg.calculator"]
 
 
+def test_experiment_to_file_rejects_nan_without_corrupting_existing_file(tmp_path):
+    """to_file() raises on NaN values instead of writing RFC 8259-invalid JSON,
+    and leaves an existing file untouched (issue #380)."""
+    file_path = tmp_path / "experiment.json"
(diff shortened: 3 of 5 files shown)


poshinchen:
@strandly-the-agent review this PR

src/strands_evals/experiment.py
```diff
@@ -758,8 +760,13 @@ def to_file(self, path: str):
 
         file_path.parent.mkdir(parents=True, exist_ok=True)
 
-        with open(file_path, "w", encoding="utf-8") as f:
-            json.dump(self.to_dict(), f, indent=2, ensure_ascii=False)
+        # Serialize and encode before opening the file so a failure (NaN/Infinity via
+        # allow_nan=False, unpaired surrogates via the utf-8 encode) raises loudly
+        # without leaving a truncated, invalid JSON artifact behind.
+        data = json.dumps(self.to_dict(), indent=2, ensure_ascii=False, allow_nan=False).encode("utf-8")
```
strandly-the-agent: 🟡 **The loud rejection never says which value is bad**, so on a large experiment the user has to hunt for it.

`Out of range float values are not JSON compliant: nan` — and for a surrogate, `can't encode character '\udcff' in position 106213`, an offset into a serialized string the user never sees. Reached from `to_file()` / `generate -o` whenever a `Case.input` or `metadata` value is `float("nan")` — a dataset built from a CSV with empty cells is enough. AGENTS.md asks for error messages with context, and the sibling `raise` above in this same method names both the input and the fix.

Suggestion: wrap the serialize and chain the stdlib detail as `__cause__` — no test churn (applied locally, 142 passed; `UnicodeEncodeError` is a `ValueError`). Naming the *first offending case* would cost ~10 more lines; optional.
```suggestion
        try:
            data = json.dumps(self.to_dict(), indent=2, ensure_ascii=False, allow_nan=False).encode("utf-8")
        except ValueError as e:
            raise ValueError(
                f"Cannot write experiment to {file_path}: it contains values that are not representable as "
                f"strict UTF-8 JSON (e.g. NaN/Infinity floats, or strings with unpaired surrogates). "
                f"Check case inputs, outputs, metadata, and evaluator fields."
            ) from e
```

src/strands_evals/experiment.py
```diff
@@ -743,7 +743,9 @@ def to_file(self, path: str):
                   Only .json format is supported.
 
         Raises:
-            ValueError: If the path has a non-JSON extension.
+            ValueError: If the path has a non-JSON extension, or if the experiment contains
+                values that cannot be represented as strict JSON (e.g. NaN/Infinity floats
+                or strings with unpaired surrogates).
```
strandly-the-agent: 🟡 **The rewritten `Raises:` misses the third failure mode — a non-serializable object raises `TypeError`, not `ValueError`.**

`Case(name="x", input="hi", metadata={"obj": SomeClass()})` → `TypeError: Object of type SomeClass is not JSON serializable`, no file written. `metadata` is `dict[str, Any]`, so this is an ordinary public path, and it's the one mode the just-rewritten block doesn't mention — a caller following the docstring catches only `ValueError` and misses it.

Suggestion (drop this if you instead widen the message-wrap in my other comment to catch `TypeError` too — either contract is fine as long as the docstring matches):
```suggestion
            ValueError: If the path has a non-JSON extension, or if the experiment contains
                values that cannot be represented as strict JSON (e.g. NaN/Infinity floats
                or strings with unpaired surrogates).
            TypeError: If the experiment contains objects that are not JSON-serializable
                (e.g. a custom class instance in case data or metadata).
```

src/strands_evals/evaluators/output_evaluator.py
```diff
@@ -75,7 +75,10 @@ def to_dict(self) -> dict:
             serializable_tools = []
             for tool in self._tools:
                 try:
-                    json.dumps(tool)
+                    # Probe with the same strictness as Experiment.to_file()'s utf-8 writer:
+                    # the encode rejects unpaired surrogates, allow_nan=False rejects
+                    # NaN/Infinity (invalid per RFC 8259). UnicodeEncodeError is a ValueError.
+                    json.dumps(tool, ensure_ascii=False, allow_nan=False).encode("utf-8")
```
strandly-the-agent, line 85: 🟡 **The probe now rejects more than the docstring and the warning claim.** Both say tools are skipped as "non-JSON-serializable … since `from_dict()` cannot reconstruct them" (the `Returns:` block above, and the `tools` property) — but the two shapes this hunk newly catches *are* serializable by plain `json.dumps` (`"my_pkg.calc_\udcff"` → escaped; NaN dict → `{"default": NaN}`), and a NaN dict *is* reconstructible. The rule the code actually enforces is "not writable as strict UTF-8 JSON".

Suggested wording for both places:

> Includes only tools writable as strict UTF-8 JSON; tools that are not (unserializable objects, NaN/Infinity floats, unpaired surrogates) are skipped with a warning and must be re-attached after `from_dict()`.

Smaller thing in the same branch: for these non-callables the identifier chain degenerates — an evaluator with a NaN dict and a surrogate string logs `tool_name=<dict>` and `tool_name=<str>`, which names neither tool. Would a truncated `repr` be more useful than the type name, now that plain strings and dicts routinely land here?

Cost: if the `skipping non-JSON-serializable tool` phrase changes, the assertions in `test_output_evaluator.py` that match on it need the new wording.

tests/strands_evals/test_experiment.py
```diff
@@ -819,6 +819,32 @@ def verify_claim(claim: str) -> str:
     assert loaded.evaluators[0].tools == ["my_pkg.calculator"]
 
 
+def test_experiment_to_file_rejects_nan_without_corrupting_existing_file(tmp_path):
```
strandly-the-agent, line 822: 🟡 **Nothing pins the premise this PR rests on — that the probe is exactly as strict as the writer.** Both new probe tests re-do `json.dumps(..., allow_nan=False).encode()` locally, so they'd still pass if the *writer* later drifted stricter — the direction that produced #380 in the first place.

One end-to-end case closes it (ran against this branch — passes, writes and reloads cleanly):

```python
def test_experiment_to_file_survives_tools_the_probe_skips(tmp_path):
    ev = OutputEvaluator(
        rubric="rubric",
        tools=[{"name": "t", "default": float("nan")}, "my_pkg.calc_\udcff", "my_pkg.calculator"],
    )
    exp = Experiment(cases=[Case(name="ok", input="hello")], evaluators=[ev])
    exp.to_file(str(tmp_path / "e.json"))
    assert Experiment.from_file(str(tmp_path / "e.json")).evaluators[0].tools == ["my_pkg.calculator"]
```

Suggestion: parametrizing the round-trip test at line 800 works just as well. Cost: one test, no src change.

strandly-the-agent (commented):
**Nothing blocking.** Does what #380 asked, with real tests behind it — and the serialize-before-open change also fully fixes #382. Four should-fix items, all DevX / docs / coverage; no correctness defects found.

Two that can't be inline comments:

- **Add `Fixes #382` to the PR body.** This diff fixes it in full — all four of its failure modes verified base vs head — and `test_experiment.py:822` is effectively #382's acceptance test, so it should auto-close. One residual worth being explicit about: `open(path, "wb")` still truncates on open, so a failure *at write time* (ENOSPC, kill) still leaves 0 bytes. That was #382's "ideally temp file + rename" ask — fine to accept and follow up, less fine to close silently.
- **`cli/commands/generate.py:94`** (outside this diff) — the stdout branch still calls `json.dumps(..., ensure_ascii=False)` without `allow_nan=False`, so after this PR `generate > exp.json` writes literal `NaN` while `generate -o exp.json` refuses it. One-word fix, and the divergence is new.

Inline: error message doesn't name the bad value · `Raises:` misses `TypeError` · docstring/warning describe a narrower rule than the probe enforces · no end-to-end probe↔writer parity test.

<details>
<summary>Verified (what was actually run)</summary>

Full pipeline: 6 independent passes (correctness, issue-alignment, API/DevX, adversarial, test-quality, docs), aggregated and deduped.

- ✅ Reviewed `8268555` on `pr-383`, base `166aa31` — diff +65/−4 across 4 files.
- ✅ `pytest tests/strands_evals/test_experiment.py tests/strands_evals/evaluators/test_output_evaluator.py` → **142 passed**; redteam + CLI suites → **347 passed**.
- ✅ Mutation check: all 4 new tests fail on base `src/` for the right reasons; **5/5 injected mutations killed** (dropping `.encode`, dropping either `allow_nan=False`, reverting to text-mode dump, moving `dumps` back inside `open`). The byte-identity and `not file_path.exists()` assertions are the load-bearing ones — they pin the #382 behavior.
- ✅ Both #380 repros re-run on head: fixed. All four #382 modes re-run base vs head: `to_dict` raise (0 bytes → intact), mid-dump raise (116-byte partial → intact), surrogate, NaN — all fixed.
- ✅ Adversarial parity attacks failed to find any probe↔writer disagreement: surrogate dict *keys* (caught only because of the new `.encode`), non-string keys, float subclasses, `-0.0`, `1e400`, `Decimal`, deep nesting.
- ✅ Byte-compat: new writer output is byte-identical to the old one for valid data on POSIX (same indent/`ensure_ascii`, no trailing newline).
- ✅ The suggested error-message wrap applied locally → 142 passed; stdlib detail preserved as `__cause__`.
- ✅ `TypeError` gap reproduced: `metadata={"obj": SomeClass()}` → `TypeError`, no file written.
- ✅ `EvaluationReport.to_file` sibling bug reproduced → filed #384.
- 🔴 **Lint not run** — `ruff` isn't available in this sandbox; PR body says `hatch run prepare` passed, not independently verified.
</details>

<details>
<summary>Questions (3, none blocking)</summary>

**❓ The write-strict / read-lenient asymmetry.** `to_file()` now refuses literal `NaN`, but `from_file()` still loads it (default `parse_constant`), so a file written by an older version can be loaded but never re-saved — verified end to end. Two of our passes landed differently here: one wanted an explicit decision, the other calls the lenient reader clearly right (rejecting at load would orphan every legacy file for zero benefit). We lean the second — no code change — but is the load→save dead end worth one line in `from_file`'s docstring, so it's documented rather than discovered? The improved error message (first inline comment) is what makes it diagnosable when hit.

**❓ Pin the `ValueError` message with `match=`?** Our test-quality pass says no as-is: it adds no kill-power, and the stdlib message's value suffix varies across 3.10–3.14. It becomes worth it only if you take the wrap suggestion, since then the message is yours.

**❓ Shared strict-dumps helper for probe + writer?** Deliberately *not* suggesting it now: two call sites, and the duplication is exactly what the inline parity test would guard. The natural moment to extract is the third call site — likely the `EvaluationReport.to_file` fix (#384).
</details>

<details>
<summary>Reading order</summary>

1. `src/strands_evals/experiment.py:763-769` — the fix itself: serialize and encode fully, *then* open `"wb"`. Everything else follows from this change.
2. `src/strands_evals/experiment.py:745-748` — the error contract that change creates (and where the `TypeError` gap is flagged).
3. `src/strands_evals/evaluators/output_evaluator.py:78-81` — the probe, retuned to match the writer exactly. Read side by side with step 1; that equality is the PR's premise.
4. `tests/strands_evals/test_experiment.py:822-845` — the two behavioral guarantees: existing file byte-identical after rejection, no file created on failure. These double as #382's acceptance tests.
5. `tests/strands_evals/evaluators/test_output_evaluator.py:335-357` — probe-level coverage for the two shapes; cheapest to read last.
</details>

<details>
<summary>Appendix — non-blocking (6)</summary>

- ⚪ `experiment.py:761` — `mkdir(parents=True)` runs before serialization, so a rejected write still leaves freshly created empty directories (confirmed). One-line tidy: move it below the `dumps`. Base did the same *and* left a truncated file, so strictly better either way.
- ⚪ `experiment.py:768-769` — `file_path.write_bytes(data)` would replace the `with open(...)` block. Pure nit; skip if a temp-file approach is taken instead.
- ⚪ Peak memory now holds the `str` and the `bytes` simultaneously — noting only for very large experiments; still better than base.
- Pre-existing: write-time truncation is still possible — a failure *during* `f.write` (ENOSPC/quota/`EFBIG`) truncates (reproduced under `RLIMIT_FSIZE`: 65,536 of 2,143,980 bytes, unparseable), byte-for-byte the same on base. The comment at `:763-765` is true for *serialization* errors, not I/O errors. Temp file + `os.replace` closes it — that's #382's remaining scope, not new debt.
- Pre-existing: `EvaluationReport.to_file` has the same bug family (`types/evaluation_report.py:324-325` — `json.dump` defaults, no `encoding=`, serializes into the open handle). Leaving it out was the right call for #380's scope — **filed #384**.
- Contrived / no action: deep non-circular nesting raises `RecursionError` past the probe's `except` — thresholds identical on base (~depth 9998), and in the narrow window where it matters this PR is a *win* (file intact vs base truncation). Int/str dict-key collision (`{1: "a", "1": "b"}`) is lossy stdlib behavior, identical on base. Windows: text mode used to write `\r\n`, `"wb"` writes `\n` verbatim — POSIX unchanged, no golden files, CI is Linux-only.
</details>

liramon2 (commented):
`EvaluationReport.to_file()` suffers from the same bug where `json.dump(self.to_dict())` is called with default settings. The fix would be similar to the `Experiment.to_file()` JSON error handling. Do you think it's within the scope of this PR?

poshinchen:
I'll prefer to make another PR for other round-trips

liramon2 (approved):

Issue #349, opened 2026-08-03
[FEATURE] Risk classification utility for multi-run results (bug vs flaky vs pass)


max-rattray-aws:
### Problem Statement

When you run an evaluation N times against the same agent, the raw results tell you pass and fail counts. They don't tell you what kind of problem you're looking at.

A case that fails 3/3 runs is a deterministic bug. A case that fails 1/3 runs is flaky. These require completely different responses, but right now both just show up as "some failures" in the flat `EvaluationReport.test_passes` list.

Triaging without classification is tedious. You open a failing case, eyeball the results across runs, mentally tally passes vs failures, and decide if it's a code bug or non-determinism. Do that for 50 cases and triage takes longer than the eval run itself.

Today `Experiment` produces one `EvaluationOutput` per evaluator per case per run. Even once N-run support lands (#228), there's no built-in utility to look at those N outputs and say "this is a bug" vs "this is flaky." The information is there in the raw results. Nobody interprets it.

### Proposed Solution

A pure utility function. Takes a list of `EvaluationOutput` from repeated runs of the same (case, evaluator) pair. Returns a stability label. No changes to `EvaluationOutput`, `EvaluationReport`, or `Experiment`.

```python
from enum import Enum
from typing import Sequence

from strands_evals.types.evaluation import EvaluationOutput


class RiskLabel(str, Enum):
    BUG = "bug"      # All runs failed. Deterministic problem.
    FLAKY = "flaky"  # Mix of pass and fail. Non-determinism.
    CNE = "cne"      # All runs were not gradable. Nothing to classify.
    PASS = "pass"    # All runs passed. Stable.


def classify_risk(results: Sequence[EvaluationOutput]) -> RiskLabel:
    """Classify a case's stability from its multi-run results.

    Filters to gradable results (those with a status of "graded", or all
    results if no status field exists yet). Then checks whether all passed,
    all failed, or a mix occurred.
    """
    # Graceful fallback: if EvaluationOutput doesn't have a status field yet,
    # treat every result as graded.
    graded = [r for r in results if getattr(r, "status", "graded") == "graded"]

    if not graded:
        return RiskLabel.CNE

    if all(r.test_pass for r in graded):
        return RiskLabel.PASS

    if not any(r.test_pass for r in graded):
        return RiskLabel.BUG

    return RiskLabel.FLAKY
```

[...]

PR #388, opened 2026-08-31
feat(classification): add classify_risk utility for multi-run results (#349)

lifelmy:
## Summary

Adds a standalone, post-hoc risk classification utility that turns repeated
`(case, evaluator)` runs into a single stability label, addressing #349.

When the same case is evaluated N times, the raw pass/fail counts don't say *what kind* of
problem a failure is. A case that fails every run is a deterministic bug; one that fails
intermittently is flaky. These need different responses, but today both surface as "some
failures". This adds a pure utility that names the difference.

## What's added

- `src/strands_evals/classification.py`
  - `RiskLabel(str, Enum)` — `BUG` / `FLAKY` / `CNE` / `PASS`
  - `classify_risk(results)` — classifies one `(case, evaluator)` pair's repeated runs
  - `classify_task_risk(evaluator_risks)` — worst-of rollup across evaluators for one case
- Public exports for all three from `strands_evals`
- `tests/strands_evals/test_classification.py` — 20 unit tests

## Design notes (matching the issue and repo conventions)

- **Standalone module, not a change to `EvaluationOutput`/`EvaluationReport`/`Experiment`.**
  Classification is invoked only when a caller already holds multi-run results, exactly as the
  issue proposes. Nothing forces multi-run onto existing single-run callers.
- **Not-applicable handling reuses the existing contract.** Rather than a bespoke `status`
  field, gradable rows are selected via the shipped `EvaluationOutput.not_applicable` property
  (the `NOT_APPLICABLE` label). This is the same "drop non-verdict rows" rule
  `EvaluationReport` already applies, so a case can't be tipped toward pass/fail by rows that
  had nothing to judge. `CNE` is returned when there are no gradable rows.
- **Worst-of rollup:** `BUG > FLAKY > PASS > CNE`. A case with one deterministic failure and
  some flaky evaluators is a bug — the deterministic signal is the actionable one.

The issue's example sketched a `status`-based gradable filter with a `getattr` fallback; since
`EvaluationOutput` already ships `not_applicable`, this implementation keys off that instead,
which is the current source of truth for "nothing to judge" and needs no new field.

## Testing

Run in a fresh worktree venv against `origin/main` (base `abf7b91`):

- `hatch test tests/strands_evals/test_classification.py` -> 20 passed
- `hatch test tests/strands_evals/types/ tests/strands_evals/test_classification.py` -> 124 passed (no regressions)
- `hatch fmt --formatter --check` -> 362 files already formatted
- `hatch fmt --linter` -> all checks passed
- `mypy -p src` (hatch-static-analysis env) -> Success, no issues in 236 source files

## Note on authorship


JackYPCOnline (changes requested):
Overall, the purpose and intension of this PR is clear. On the other side, the naming and DEVX remain unclear to me, I would suggest give this more thought and how can we make it more explicit, what apis we want to expose. With said, the purposed in the issue ticket need refinement.

lifelmy:
Thanks for the review. I agree that `RiskLabel` / `classify_risk` is too broad and can be confused with the red-team risk APIs, while `classify_task_risk` does not make it obvious that it aggregates evaluator-level stability verdicts.

Before changing the public API, I propose making the concepts explicit:

- `EvaluationStability`: the result across repeated runs of one case/evaluator pair
- `classify_evaluation_stability(outputs)`: classify that repeated-run result
- `combine_evaluation_stability(verdicts)`: apply the documented worst-of ordering across evaluators

I would also revise issue #349 to define the input grouping contract, the handling of `not_applicable`, the empty-input behavior, and why this is a post-hoc helper rather than an `Experiment`/`EvaluationReport` method. Does that API direction match what you have in mind, or would you prefer this to live on a report/analysis object instead of as standalone functions? I have not churned the public API yet so that the next revision can follow the intended DX.

lifelmy:
Task A base refresh: merged current `origin/main` (`24acea695`, CI command pinning) without conflicts and pushed `8d61fd9c6` (0 behind / 4 ahead). The classification API diff is unchanged; the existing naming/DEVX review direction remains pending and I have not made unilateral public-API changes. Verification: classification suite 20 passed (one dependency deprecation warning); focused Ruff check/format and `git diff --check` passed.

lifelmy:
Task A base refresh: merged current `origin/main` (`9ea55c96d461`, README-only) without conflicts and pushed `3d69e7978cad` (0 behind / 5 ahead). The classification diff is unchanged. Verification: classification suite 20 passed (one dependency deprecation warning); repository-wide Ruff check/format and `git diff --check` passed.

Issue #347, opened 2026-08-03
[FEATURE] Per-run keying in EvaluationDataStore for multi-run resume and config-drift detection


max-rattray-aws:
### Problem Statement

`EvaluationDataStore.load` and `.save` key results by `case_name` alone. One case, one cached `EvaluationData`. This means:

1. You can't run the same experiment N times for statistical confidence. Running twice just loads the cached result from run 1. There's no concept of "run 2 of 5."

2. You can't top up runs. If you ran 3 overnight and want 5 total, you either re-run everything or manage result directories yourself.

3. There's no detection of config drift between sessions. If you change the evaluator rubric, swap the model, or edit the system prompt between runs, the store happily mixes stale cached results from the old config with new ones. The aggregate report silently mixes apples and oranges.

`LocalFileTaskResultStore` writes `{case_name}.json` per case. No run index, no manifest, no hash of the config that produced the result.

### Proposed Solution

Extend `EvaluationDataStore` to support per-run keying and add a lightweight config-drift guard. No new store backend needed. The existing `LocalFileTaskResultStore` layout just gains structure.

**Per-run keyed storage**

Add a `run_index` parameter to the protocol:

```python
@runtime_checkable
class EvaluationDataStore(Protocol):
    def load(self, case_name: str, run_index: int = 0) -> EvaluationData | None: ...
    def save(self, case_name: str, result: EvaluationData, run_index: int = 0) -> None: ...
    def completed_run_count(self, case_name: str) -> int: ...
```

`run_index=0` default preserves backward compat. Existing single-run callers pass no second arg and get the same behavior as today. `completed_run_count` tells the runner what's already done.

For `LocalFileTaskResultStore`, the on-disk layout becomes:

```
results/
  case_name/
    run_0.json
    run_1.json
    run_2.json
```

Existing flat `{case_name}.json` files get treated as `run_0` for migration.

**Resume logic in the experiment runner**

`Experiment.run_evaluations_async` currently iterates `self._cases` once, calling `evaluation_data_store.load(case.name)` to check for cached results. With `num_runs`, the runner calls `completed_run_count(case.name)` per case. If the count already meets `num_runs`, skip. Otherwise run only the remaining indices.

```python
report = await experiment.run_evaluations_async(
    task_function,
    max_workers=4,
    evaluation_data_store=store,
    num_runs=5,  # run each case 5 times total, resuming from whatever exists
)
```

[...]

Issue #186, opened 2026-04-03
feat(experiment): ExperimentTrendAnalyzer — cross-run overall_score regression detection


nanookclaw:
## Problem

`EvaluationReport.overall_score` captures the aggregate pass rate for a given experiment run. When `run_evaluations()` executes, it produces one `EvaluationReport` per evaluator — but there is no mechanism to track how `overall_score` changes across **sequential experiment runs**.

A trajectory like `0.91 -> 0.85 -> 0.78 -> 0.71` across four runs produces zero signal. Each run is a fresh report; no analysis layer connects them.

The `LocalFileTaskResultStore` already persists per-case `EvaluationData` to disk. The per-run `EvaluationReport` contains `overall_score`, `scores` (per-case), and `test_passes`. The data for slope analysis exists; the analysis layer is absent.

## Proposed: `ExperimentTrendAnalyzer`

```python
from dataclasses import dataclass
from pathlib import Path
import json
import statistics

@dataclass
class RunPoint:
    run_id: str
    overall_score: float
    pass_rate: float

@dataclass
class MetricTrend:
    slope: float       # OLS slope via statistics.linear_regression
    direction: str     # "improving" | "degrading" | "stable"
    first: float
    last: float
    window: int

@dataclass
class ExperimentTrendReport:
    runs: list[RunPoint]
    overall_score_trend: MetricTrend
    pass_rate_trend: MetricTrend
    any_regression: bool

class ExperimentTrendAnalyzer:
    def __init__(self, report_paths: list[Path], window: int = 20):
        self.report_paths = sorted(report_paths)[-window:]

    def analyze(self) -> ExperimentTrendReport:
        runs = []
        for path in self.report_paths:
            data = json.loads(path.read_text())
            runs.append(RunPoint(
                run_id=path.stem,
                overall_score=data["overall_score"],
                pass_rate=sum(data["test_passes"]) / len(data["test_passes"])
                    if data.get("test_passes") else data["overall_score"],
            ))

        def _trend(values):
            xs = list(range(len(values)))
            slope, _ = statistics.linear_regression(xs, values)
            direction = "improving" if slope > 0.005 else "degrading" if slope < -0.005 else "stable"
            return MetricTrend(slope=slope, direction=direction,
                               first=values[0], last=values[-1], window=len(values))

        overall_trend = _trend([r.overall_score for r in runs])
        pass_trend = _trend([r.pass_rate for r in runs])
        return ExperimentTrendReport(
            runs=runs,
[...]

PR #390, opened 2026-09-01
feat(store): add per-run keying and config-drift guard to EvaluationDataStore (#347)

lifelmy:
## Description

Extends the result-store layer so a case can be evaluated N times with every run kept, and guards against silently mixing results from different experiment configs. This is the store-layer foundation the issue describes as "additive and backward compatible" and the producer side of the multi-run history that #186's trend analyzer consumes.

**`EvaluationDataStore` protocol** — `load`/`save` gain a defaulted `run_index: int = 0`, and a new `completed_run_count(case_name) -> int` reports how many consecutive runs a case already has. `run_index=0` preserves the exact single-run behavior, so existing callers need no change.

**`LocalFileTaskResultStore`** —
- Per-run layout: each case gets its own subdirectory (`results/<case>/run_0.json`, `run_1.json`, ...). `completed_run_count` counts consecutive runs from 0 and stops at the first gap, so it returns the next index to write when topping up to N.
- Migration: a legacy flat `{case_name}.json` from the old single-run layout is read as that case's run 0, so existing stores keep working with no migration step.
- Config-drift guard: an optional `config_hash` (plus `overwrite`) is recorded in a `_manifest.json`. Reopening with a different hash raises `ConfigDriftError` rather than appending incompatible results; `overwrite=True` discards prior results/manifest and starts fresh. Passing no hash skips the guard entirely.

**Scope note (deliberate):** this PR stops at the store contract + local implementation, which is self-contained and backward compatible. The `num_runs` resume loop inside `run_evaluations_async` and the `--num-runs`/`--overwrite` CLI flags from the proposal are intentionally left for a focused follow-up so this change stays small and easy to review — happy to send that next if maintainers prefer it here. The proposed `EvaluationDataStore` shape (defaulted `run_index` + `completed_run_count`) is exactly what that runner change will build on.

## Related Issues

Addresses #347 (store-layer portion). Complements #186 (trend analysis consumes multi-run history).

## Type of Change

New feature

## Testing

`hatch test tests/strands_evals/test_local_file_task_result_store.py` -> 14 passed (per-run keying, legacy-flat migration, config-drift raise/overwrite/no-guard). `hatch test tests/strands_evals/test_local_file_task_result_store.py tests/strands_evals/test_experiment.py` -> 131 passed (updated the in-memory `DictEvaluationDataStore` test double to the new protocol; existing store/experiment behavior unchanged). `hatch fmt --formatter --check` -> 360 files already formatted; `hatch fmt --linter` -> all checks passed; `mypy -p src` via hatch-static-analysis env -> Success, no issues in 235 source files.

- [x] I ran the linter, formatter, type checker, and the affected test suites

## Checklist
- [x] I have reviewed and understand every line of code in this PR. I use a coding assistant to help implement, and I review and take responsibility for the final change.
- [x] My change is focused and reasonably small; I have split the runner/CLI wiring into a follow-up
- [x] I have added tests that prove the feature works
- [ ] I have updated the documentation accordingly (public store API changed; happy to add docs if maintainers point me at the source)

----


lifelmy:
Task A follow-up: the repository-wide pre-commit test run exposed one stale CLI assertion that still expected the legacy flat c1.json layout. I updated it to assert the new per-run path c1/run_0.json and pushed ae283b3. Verification: focused store/experiment/CLI tests -> 132 passed; pre-commit format, lint, and full unit-test hooks passed; git diff --check passed.

lifelmy:
Task A base refresh: merged current `origin/main` (`24acea695`, CI command pinning) without conflicts and pushed `c538646b5` (0 behind / 5 ahead). The evaluation-store diff is unchanged. Verification: store/experiment/CLI suites 166 passed; focused Ruff check/format and `git diff --check` passed.

lifelmy:
Task A base refresh: merged current `origin/main` (`9ea55c96d461`, README-only) without conflicts and pushed `f8e9a468e533` (0 behind / 6 ahead). The evaluation-store diff is unchanged. Verification: store/experiment/CLI suites 166 passed; repository-wide Ruff check/format and `git diff --check` passed.

PR #393, opened 2026-09-04
fix: make EvaluationReport.to_file() reject non-strict JSON

### LINKED ISSUE #384: [BUG] EvaluationReport.to_file() writes RFC-8259-invalid JSON (literal NaN), has no explicit encoding, and can leave a truncated file
### Checks

### Strands Version

1.53.0

### Strands Evals Version

main @ 166aa31

### Python Version

3.13.14

### Operating System

Amazon Linux 2023 (aarch64)

### Installation Method

git clone

### Steps to Reproduce

`EvaluationReport.to_file()` (`src/strands_evals/types/evaluation_report.py:324-325`) is the sibling of `Experiment.to_file()` and has the same bug family that #380 reported and #383 fixes for experiments — plus one more:

- `json.dump` is called with defaults, so `allow_nan=True`: NaN/Infinity scores are written as bare `NaN` — not valid JSON per RFC 8259 §6.
- `open(file_path, "w")` passes no `encoding=`, so the file's bytes are locale-dependent on non-UTF-8 systems.
- It serializes *into* the already-open handle, so a mid-serialization failure truncates or destroys an existing file — same shape as #382.

```python
import json
from strands_evals.types.evaluation_report import EvaluationReport

r = EvaluationReport(overall_score=float("nan"), scores=[float("nan"), 1.0],
                     cases=[{"name": "a"}], test_passes=[True, False])
r.to_file("report.json")   # "succeeds"

txt = open("report.json").read()          # contains:  "overall_score": NaN
json.loads(txt, parse_constant=lambda c: (_ for _ in ()).throw(ValueError(c)))  # ValueError: NaN
```

Reproduced on main @ 166aa31 (and unchanged on #383's head 8268555 — correctly out of that PR's scope).

### Expected Behavior

The report file is valid, strict, UTF-8 JSON — or the write fails loudly before touching the file.

### Actual Behavior

The file contains the literal `NaN`, which Python's lenient `json.loads` reads back but any strict parser (JS `JSON.parse`, `jq`, serde, ...) rejects. `overall_score: float` and `scores: list[float]` make NaN reachable from ordinary aggregation (custom evaluators returning NaN), and the report is exactly the artifact meant to be machine-read downstream.

### Additional Context

Found while reviewing #383. Same family, lower priority: `LocalFileTaskResultStore.save` (`src/strands_evals/local_file_task_result_store.py:38-39`) uses `write_text(result.model_dump_json(indent=2))` with no `encoding=`, and pydantic also emits bare `NaN`.

### Possible Solution

Mirror #383: serialize first, then write bytes —

```python
data = json.dumps(self.to_dict(), indent=2, ensure_ascii=False, allow_nan=False).encode("utf-8")
with open(file_path, "wb") as f:
    f.write(data)
```

and extend the `Raises:` block accordingly. This would be the third call site of that exact pattern, so it may be the right moment to extract a shared strict-dumps helper.

### Related Issues

#380, #382, #383

serenearyal:
## Description

`EvaluationReport.to_file()` had the same three problems that #383 fixed for `Experiment.to_file()`:

- `json.dump` ran with defaults, so a NaN or Infinity score was written as the bare literal `NaN`. Python's lenient `json.loads` reads it back, but strict parsers (`jq`, JavaScript `JSON.parse`, serde) reject the file. `overall_score: float` and `scores: list[float]` make NaN reachable from ordinary aggregation.
- The destination was opened with mode `"w"` before serialization, so a failure mid-write truncated the file. On main, a valid 224-byte report was cut to 106 bytes of invalid JSON after a save with a non-serializable case value. If `to_dict()` itself raised, the file was left at 0 bytes.
- The file was opened without `encoding=`. This was harmless while `ensure_ascii` defaulted to True, but it becomes a real bug on non-UTF-8 locales once the writer emits UTF-8 text like #383 does.

### Changes

- `EvaluationReport.to_file()` now mirrors `Experiment.to_file()`: serialize first with `ensure_ascii=False, allow_nan=False`, encode to UTF-8, and write the bytes only on success. NaN, Infinity, and unpaired surrogates raise a `ValueError` with the original error as `__cause__`. Non-serializable objects raise `TypeError` before the file is touched. The `Raises:` docstring is extended to match.
- `EvaluationReport.from_file()` reads with `encoding="utf-8"` so the read side matches the write side, the same as `Experiment.from_file()`.
- Four regression tests in `tests/strands_evals/types/test_evaluation_report.py`: NaN rejected with the existing file intact, unpaired surrogates rejected without writing, non-serializable value rejected with the existing file intact, and a UTF-8 round trip.

This is a plain mirror of #383 rather than a shared helper. Happy to extract one if you prefer; the issue notes this is the third copy of the pattern.

## Related Issues

Fixes #384. Same pattern as #383 (Experiment.to_file), which also covers #382.

## Documentation PR

None needed. Behaviour change is limited to raising on data that could not produce valid JSON.

## Type of Change

Bug fix

## Testing

- Added four regression tests; all fail on main and pass with this change.
- `hatch test` on Python 3.10, 3.11, 3.12, 3.13, 3.14: 2144 passed on each.
- `hatch fmt --formatter`, `hatch fmt --linter`, mypy, and `pre-commit run` on the changed files: clean.
- `hatch run test-integ` not run: no integration test exercises `EvaluationReport.to_file()`, and the suite needs AWS credentials I do not have locally.
- Reproduced the truncation and NaN cases with a scratch script on main, and confirmed the fixed writer raises before touching the file.

## Checklist

----


poshinchen:
@strandly-the-agent review the changes.

strandly-the-agent (commented):
**Approve — nothing blocking.** Faithful mirror of #383, and the four new tests are load-bearing: all four fail on `main` for the right reasons, and 5 of the 6 mutations I injected into the fix are killed. No correctness defect found — the writer's output is byte-identical to main's for valid data, and the error class, message shape, and `__cause__` chaining match `Experiment.to_file()` exactly.

One item needs a maintainer decision rather than a code change: @poshinchen — `serenearyal` has now asked twice (in #384 and in the PR body) whether to extract a shared strict-dumps helper. That pattern is at four call sites (`experiment.py:767`, `types/evaluation_report.py:330`, `evaluators/output_evaluator.py:85`, `cli/commands/generate.py:96`). Keeping it duplicated in a bugfix PR is defensible — the author just needs an answer either way.

<details>
<summary>✅ Verified (what actually ran)</summary>

Reviewed `f28854e` on `pr-393` against base `abf7b91` — +81/−5 across 2 files. Python 3.13.15, pydantic 2.13.5. Full log in the artifacts (`pr393-verification.log`).

- ✅ `pytest tests/strands_evals/types/test_evaluation_report.py` → **37 passed**. Full suite → **2086 passed**; the 9 failures / 49 errors are missing optional deps (`langfuse`, otel) in my sandbox and are identical with main's `evaluation_report.py` swapped in.
- ✅ **Fail-on-main confirmed** for all four new tests, with the right reasons: two `DID NOT RAISE ValueError`, one byte-diff showing the truncated 87-byte file (`"obj": ` cut off mid-write), one `\u00e9`-escaped output instead of UTF-8 bytes.
- ✅ **Mutation check, 5/6 killed:** dropping `.encode("utf-8")`, `allow_nan=False`, `ensure_ascii=False`, the `ValueError` wrap, or moving the dump back inside the open handle all fail tests. Survivor: dropping `encoding="utf-8"` from `from_file` (see appendix).
- ✅ **Parity with `Experiment.to_file`:** NaN, ±Infinity, surrogate in a dict *key*, surrogate in a value → wrapped `ValueError` with the stdlib error as `__cause__` (`UnicodeEncodeError` for surrogates); `object()` → `TypeError`, matching the new `Raises:` block. `-0.0` and `1e308` still write fine.
- ✅ **No collateral:** nothing written and no leftover empty parent dirs on rejection (`mkdir` correctly sits below the `dumps` — the nit that came out of #383). Byte-identical to main's writer for valid ASCII data, no trailing newline, so no golden-file churn.
- ✅ **Round trip:** `café ☕` / `日本語` written as UTF-8, re-parsed with a strict `parse_constant` that throws, and `from_file` → `to_dict()` equal to the original.
- ✅ `ruff check` and `ruff format --check` clean on both changed files. mypy isn't installed in my sandbox, so the PR's mypy claim is unverified.
- ✅ Reports now inherit the same load→save asymmetry as experiments: a legacy file containing bare `NaN` still loads, then can't be re-saved. Same trade-off already accepted on #383 (lenient reader), so no action — just noting it now applies here too.
</details>

<details>
<summary>Appendix — non-blocking (3)</summary>

- ⚪ **`types/evaluation_report.py:364` — the `encoding="utf-8"` addition to `from_file` has no test that would fail without it** (that mutation survived). It's unfalsifiable on a UTF-8 CI without faking the locale, and `Experiment.from_file` has carried the same line untested since #69, so this is consistent rather than a gap worth fixing. Flagging only so the coverage boundary is known.
- ⚪ **`types/evaluation_report.py:334-337` — the hint names 2 of the 7 serializable fields.** "Check the scores and case records" misses `reasons`, `detailed_results`, `diagnoses`, `recommendations`, which can also hold a NaN or a surrogate. I tried to build a realistic path into `reasons` and failed: the error-isolation branch (`experiment.py:466`) formats `str(e)`, and `str(OSError)` reprs the filename, so surrogates arrive pre-escaped. What's left is a judge's model-supplied `\udXXX` escape — thin enough that this is only worth a word change if you're editing the line anyway.
- **Pre-existing, unrelated, not filed:** `README.md:390` and `README.md:665` still show `experiment.to_file("name", "json")`; the second argument doesn't exist, so both examples raise `TypeError: to_file() takes 2 positional arguments but 3 were given`. Different symbol from this PR and a one-line docs fix, so I left it with you instead of opening an issue.
</details>

PR #396, opened 2026-09-09
feat(tools): add KnowledgeIndex for progressive disclosure over reference knowledge

### LINKED ISSUE #395: [FEATURE] Progressive disclosure for large reference knowledge, not just traces (extend TraceIndex with a KnowledgeIndex)
## Summary

#343 gave judge-based evaluators progressive disclosure over a large **trace**: instead of inlining a whole `Session` into the prompt (which overflows the judge and gets scored as a failure — #342), `TraceIndex` builds a compact overview plus `list_spans`/`get_span`/`search_spans` tools and a `for_judge()` composition helper, so the judge reads only the spans it needs.

The same overflow problem exists for a second kind of input that is **not in the trace**: a large **reference corpus** the judge must consult to score a rubric. This issue asks for the same progressive-disclosure treatment for arbitrary keyed reference material.

## The gap

Grounding, feasibility, and instruction-adherence rubrics frequently need the judge to check the trace against knowledge that lives outside it, for example:

- the **catalog of tools/skills the agent could have used** and their full input/output schemas or contracts,
- API specs, policy/guideline documents, or domain reference text.

Two things make this distinct from `TraceIndex`:

1. **It overflows for the same reason.** A realistic capability catalog (dozens of skills, hundreds of tool schemas) or a policy corpus does not fit inline any more than a large trajectory does — and inlining all of it degrades judge accuracy even when it *does* fit (lost-in-the-middle, position bias; see the references in #342).
2. **The trace can't answer capability questions.** A `Session` only contains the tools/skills a given run *actually invoked*. A rubric like *"was there a capability that would have satisfied this request?"* or *"does the plan conform to the documented contract for the tools it called?"* needs the **full catalog**, which is a standalone knowledge source, not a slice of the trace. `available_tools` on the trace tells you what that run was handed, not the universe.

So `TraceIndex` (bound to a `Session`) cannot serve these rubrics, and inlining the corpus reintroduces #342.

## Proposed direction — a `KnowledgeIndex` sibling to `TraceIndex`

Add a `KnowledgeIndex` over an arbitrary collection of **keyed** text documents, with the same shape as `TraceIndex` so it composes with `OutputEvaluator` the identical way:

```python
from strands_evals.tools.knowledge_index import KnowledgeIndex

# entries: key -> document text (tool schemas by tool name, skill docs by domain, …)
index = KnowledgeIndex(catalog)                      # a large keyed corpus

# --- Mode A: retrieve-then-inject (deterministic, one LLM call) ---
# The metric selects the trace-relevant keys and injects only those.
reference_block = index.render(keys=plan_tool_names) # a bounded <Knowledge> block
judged_output = f"{agent_answer}\n{reference_block}"

# --- Mode B: agentic discovery (judge pulls what it needs) ---
prompt_section, tools = index.for_judge()            # overview + list/get/search tools
```

Concretely:

- `overview(offset)` — one line per entry (key, optional description, size, truncated preview), paged so it always fits in `max_read_chars`.
- `list_entries` / `get_entry(key)` / `search_entries(pattern, …)` — the discovery tools, same paging/windowing/`max_read_chars` bound and case-insensitive literal-or-regex search (with line-anchored `^`/`$`) as `TraceIndex`.
- `for_judge()` → `(prompt_section, tools)` — the atomic composition helper.
- **`render(keys)`** — a deterministic, tool-free path that renders just the selected entries as a bounded `<Knowledge>` block for the common *retrieve-then-inject* case (metric knows which entries are relevant from the trace, so no extra round-trips).

### Two usage modes, and when to use which

- **Retrieve-then-inject (default).** When the relevant reference keys are derivable from the trace (the tools the plan called, the domains it touched), the metric selects those keys and injects only that slice via `render()`. One LLM call, deterministic, cheap.
- **Agentic discovery (reserve).** When the needed reference *can't* be predetermined — e.g. *"does **any** entry cover this request?"* when the trace names no matching tool — hand the judge the tools via `for_judge()` and let it look things up. More round-trips and harder to make deterministic, so use it only where open-ended lookup is genuinely required.

### Why not just reuse `TraceIndex`

`TraceIndex` is bound to a `Session` and keys by span index; a reference corpus is an arbitrary keyed document set with no spans and no timeline. The list/get/search/paging **machinery** is shared (worth extracting so both indexes stay identical in behavior), but the surface — keyed entries, `render()` for slice injection — is different enough to warrant a sibling class rather than overloading `TraceIndex`.

## Scope

- Domain-agnostic: `KnowledgeIndex` takes any `key -> text` mapping; how a corpus is assembled/loaded stays with the caller.
- No change to `TraceIndex`'s observable behavior; only a shared-helper extraction.
- Follow-up PR with the implementation, A/B evidence (a judge with an oversized catalog scored via inline vs. `render`/`for_judge`), and tests is ready if this direction looks right.

pdebjyot:
## What & why

Closes #395. Extends the progressive-disclosure work from #343 to a second kind
of oversized input: **reference knowledge that lives outside the trace** — the
catalog of tools/skills the agent could have used, API specs, policy documents.

That corpus overflows the judge for the same reason a large trajectory does
(#342), and inlining all of it degrades accuracy even when it fits
(lost-in-the-middle, position bias). A `Session` also can't answer capability
questions — it only holds what a run *actually invoked*, so a rubric like *"was
there a tool that would have satisfied this request?"* needs the full catalog,
which is a standalone knowledge source rather than a slice of the trace.

## What's in the PR

**1. `refactor(tools)` — extract shared helpers (`_progressive.py`).**
`TraceIndex`'s paging, windowing, and budget-bounded search move into a shared
module so the new index reuses identical behavior and the two can't drift. No
observable change to `TraceIndex` (`overview()`/`get_span`/`search_spans` now
delegate; local `_window`/`_preview`/constants removed). Existing `TraceIndex`
tests pass unchanged.

**2. `feat(tools)` — `KnowledgeIndex`.** Same list / get / search shape over an
arbitrary keyed corpus (`{key: text}`, `KnowledgeEntry` iterable, or pairs), with
two modes:

```python
from strands_evals.tools.knowledge_index import KnowledgeIndex

index = KnowledgeIndex(tool_catalog)

# Mode A — retrieve-then-inject (default; deterministic, one LLM call)
knowledge_block = index.render(keys=plan_tool_names)   # bounded <Knowledge> block
judged_output = f"{agent_answer}\n{knowledge_block}"

# Mode B — agentic discovery (reserve for open-ended lookup)
prompt_section, tools = index.for_judge()              # overview + list/get/search
```

- **Mode A (`render`)** — the metric selects the trace-relevant keys and injects
  only that bounded slice. Use it whenever the keys are derivable from the trace.
- **Mode B (`for_judge`)** — overview + `list_entries`/`get_entry`/`search_entries`
  tools; reserve for when the needed entry can't be predetermined.

Both honor the same `max_read_chars` bound and case-insensitive literal-or-regex
search (line-anchored `^`/`$`) as `TraceIndex`. Composes with `OutputEvaluator`
(caller-controlled output), **not** `TrajectoryEvaluator`.

## Show it works

Offline, on a 201-entry tool catalog (~85 KB) with the one decisive tool buried
mid-catalog:

| path | chars into the prompt | decisive tool available? |
|---|---|---|
| inline full catalog | 84,730 | yes, but overflows / judge skims |
| Mode B overview | 8,054 (**10.5×** smaller, paged) | found via `search_entries("escalate")` |
| Mode A `render(keys)` | 318 (**266×** smaller) | yes, exact slice |

`tests_integ/test_knowledge_index_judge_reliability.py` is a Bedrock A/B (skips
without creds): the agent declines a request claiming no capability exists;
inline vs `render` vs `explore` are scored on a grounded case (no such tool →
declining is correct) and a fabricated case (tool exists mid-catalog → declining
is wrong). Both progressive modes must separate the two.

## Tests / checks

- 23 new unit tests for `KnowledgeIndex` (paging, get/missing key, windowing,
  literal/regex/anchored/budget search, `render` slice/dedupe/bounds, input
  coercion, validation). Full suite: **1793 passed**.
- `ruff check` and `mypy -p src` clean.
- No behavior change to `TraceIndex` — its tests pass as-is.

Builds on #343; independent of the #342 could-not-evaluate status work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,31 @@ evaluator = TrajectoryEvaluator(
 )
 ```
 
+### Evaluating Large Traces with Progressive Disclosure
+
+When a session is too large to inline into a judge prompt (large tool results,
+many turns), give the judge a compact overview plus discovery tools instead of
+the full trajectory. The judge loads only the spans the rubric requires:
+
+```python
+from strands_evals.evaluators import OutputEvaluator
+from strands_evals.tools.trace_index import TraceIndex
+
+index = TraceIndex(session)  # session: a Session from any provider/mapper
+
+evaluator = OutputEvaluator(
+    rubric="Every factual claim must be supported by tool-result evidence in the trace.",
+    tools=index.tools,  # list_spans, get_span, search_spans
+)
+
+# Compose the prompt with the compact overview instead of the full trajectory
+evaluation_output = f"{agent_answer}\n\n<TraceOverview>\n{index.overview()}\n</TraceOverview>"
+```
+
+The overview is one line per span (index, type, tool name, sizes, preview);
+`get_span` pages through oversized spans so no single tool return can overflow
+the judge's context.
+
 ### Trace-based Helpfulness Evaluation
 
 Evaluate agent helpfulness using OpenTelemetry traces with seven-level scoring:
--- a/README.md
+++ b/README.md
@@ -231,9 +231,10 @@ evaluator.evaluate(EvaluationData(input=user_prompt, actual_output=judged_output
 ```
 
 The overview is one line per span (index, type, tool name, sizes, preview);
-`list_spans` and `get_span` page through long traces and oversized spans so no
-single tool return can overflow the judge's context. The rubric must tell the
-judge to verify claims with the tools — otherwise it scores off the previews alone.
+`list_spans`, `get_span`, and `search_spans` all page or cap their output at
+`max_read_chars` so no single tool return can overflow the judge's context. The
+rubric must tell the judge to verify claims with the tools — otherwise it scores
+off the previews alone.
 
 > **Note:** this composes with `OutputEvaluator`, whose prompt is caller-controlled.
 > It does **not** work with `TrajectoryEvaluator`, which inlines the full
--- a/README.md
+++ b/README.md
@@ -209,21 +209,36 @@ the full trajectory. The judge loads only the spans the rubric requires:
 ```python
 from strands_evals.evaluators import OutputEvaluator
 from strands_evals.tools.trace_index import TraceIndex
+from strands_evals.types import EvaluationData
 
 index = TraceIndex(session)  # session: a Session from any provider/mapper
 
+# for_judge() hands back both halves together so neither is forgotten:
+# the overview to put next to the answer, and the discovery tools for the judge.
+prompt_section, tools = index.for_judge()  # tools: list_spans, get_span, search_spans
+
 evaluator = OutputEvaluator(
-    rubric="Every factual claim must be supported by tool-result evidence in the trace.",
-    tools=index.tools,  # list_spans, get_span, search_spans
+    rubric=(
+        "Every factual claim must be supported by tool-result evidence in the trace. "
+        "Use the trace tools to verify each claim against the evidence before scoring."
+    ),
+    tools=tools,
 )
 
-# Compose the prompt with the compact overview instead of the full trajectory
-evaluation_output = f"{agent_answer}\n\n<TraceOverview>\n{index.overview()}\n</TraceOverview>"
+# Compose the overview into the judged output instead of the full trajectory:
+judged_output = f"{agent_answer}\n{prompt_section}"
+evaluator.evaluate(EvaluationData(input=user_prompt, actual_output=judged_output))
 ```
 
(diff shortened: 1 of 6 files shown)


JackYPCOnline:
Unassign myself since it is an Eval PR.

PR #397, opened 2026-09-11
feat(output_evaluator): add cache_config support for prompt caching

shssrd:
## Motivation

When using `OutputEvaluator` for repeated evaluations with the same rubric (e.g., scoring thousands of agent traces against a fixed quality rubric), the system prompt is identical across every call. Bedrock's prompt caching can cache this stable prefix and charge cached input tokens at a 90% discount, significantly reducing TPM consumption and cost.

The Strands Agents SDK already supports `CacheConfig` on `BedrockModel`, but `OutputEvaluator` doesn't expose it. When `model` is passed as a string, the evaluator lets `Agent` create a default `BedrockModel` with no caching.

## Changes

- Add optional `cache_config: CacheConfig | None = None` parameter to `OutputEvaluator.__init__`
- Add `_resolve_model()` helper: when `cache_config` is set and `model` is a string, creates a `BedrockModel` with caching enabled
- Both `evaluate()` and `evaluate_async()` use `_resolve_model()`
- `to_dict()` serializes `cache_config` via `dataclasses.asdict()`
- 8 new unit tests (all passing)

## Usage

```python
from strands.models.model import CacheConfig

evaluator = OutputEvaluator(
    rubric="Evaluate quality on a scale of 0 to 1...",
    model="us.anthropic.claude-sonnet-4-6",
    cache_config=CacheConfig(strategy="auto"),
)
```

## Backward compatible

- `cache_config` defaults to `None` (no behavior change for existing users)
- When `model` is already a `Model` instance, `cache_config` is ignored (caller controls their own caching)


src/strands_evals/evaluators/output_evaluator.py
```diff
@@ -74,6 +98,10 @@ def to_dict(self) -> dict:
             with a warning and must be re-attached after `from_dict()`.
         """
         _dict = super().to_dict()
+        if self.cache_config is not None:
+            from dataclasses import asdict
```
poshinchen, line 102: This can be moved to the top right?

src/strands_evals/evaluators/output_evaluator.py
```diff
@@ -63,6 +69,24 @@ def tools(self) -> list[Any] | None:
...
+        BedrockModel with caching enabled. This allows the system prompt (rubric) to be
+        cached across evaluation calls within the provider's cache window, reducing input
+        token consumption for repeated evaluations with the same rubric.
+
+        Returns:
+            The resolved model: a BedrockModel with caching if applicable, or the
+            original model value unchanged.
+        """
+        if self.cache_config is not None and isinstance(self.model, str):
+            from strands.models.bedrock import BedrockModel
```
poshinchen, line 85: nit: this can be moved to the static import since loading BedrockModel should be cheap.

tests/strands_evals/evaluators/test_output_evaluator.py
```diff
@@ -407,3 +407,119 @@ def verify_claim(claim: str) -> str:
...
+    call_kwargs = mock_agent_class.call_args[1]
+    assert call_kwargs["callback_handler"] is None
+
+
+@patch("strands_evals.evaluators.output_evaluator.Agent")
+def test_output_evaluator_evaluate_no_bedrock_model_when_model_is_instance(
+    mock_agent_class, evaluation_data, mock_agent
+):
+    """Test that cache_config does not wrap an already-instantiated Model"""
+    from strands.models.model import CacheConfig
```
poshinchen, line 460: You should have static import instead of dynamic

poshinchen (changes requested):

poshinchen:
@strandly-the-agent review the changes carefully, also review my comments

src/strands_evals/evaluators/output_evaluator.py
```diff
@@ -37,13 +41,15 @@ def __init__(
...
         tools: list[Any] | None = None,
+        cache_config: CacheConfig | None = None,
     ):
         super().__init__(name=name)
         self.rubric = rubric
         self.model = model
         self.include_inputs = include_inputs
         self.system_prompt = system_prompt
         self.uses_environment_state = uses_environment_state
+        self.cache_config = cache_config
```
strandly-the-agent, line 52: 🔴 **Breaks `Experiment` round-trip for every `Multimodal*Evaluator`, and for this evaluator when `cache_config` is set.**

Base `to_dict()` (`evaluator.py:349-366`) emits every public attribute not matching a default in `self.__class__.__init__`. `MultimodalOutputEvaluator.__init__` (`multimodal_output_evaluator.py:40-50`) has no `cache_config` kwarg, so `cache_config: None` is emitted and `from_dict` (`experiment.py:840`) raises `TypeError: unexpected keyword argument 'cache_config'`. That's the 5 CI failures. Separately, when it *is* set, `from_dict` hands back the `asdict` dict and `BedrockModel` fails on `cache_config.strategy` at the first call.

The file already solves the first problem for `tools` (comment two lines below) — same trick fixes both:

```suggestion
        # Private for the same reason as _tools; to_dict() below re-adds it when set.
        self._cache_config = CacheConfig(**cache_config) if isinstance(cache_config, dict) else cache_config
```
plus a `cache_config` property returning `self._cache_config` and `cache_config: CacheConfig | dict | None` on the ctor. Verified: `test_output_evaluator.py` + `test_experiment.py` go from 5 failed → 151 passed.

src/strands_evals/evaluators/output_evaluator.py
```diff
@@ -63,6 +69,24 @@ def tools(self) -> list[Any] | None:
...
+        When cache_config is set and model is a plain string (model ID), creates a
+        BedrockModel with caching enabled. This allows the system prompt (rubric) to be
+        cached across evaluation calls within the provider's cache window, reducing input
+        token consumption for repeated evaluations with the same rubric.
+
+        Returns:
+            The resolved model: a BedrockModel with caching if applicable, or the
+            original model value unchanged.
+        """
+        if self.cache_config is not None and isinstance(self.model, str):
```
strandly-the-agent, line 84: 🟡 **`cache_config` is silently ignored when `model` is left at its default (`None`).**

`isinstance(None, str)` is False, so `_resolve_model()` returns `None` and `Agent` builds a default `BedrockModel` with no caching — the most common way to construct this evaluator gets no benefit and no signal. (`test_output_evaluator_init_with_cache_config` constructs exactly this combination.)

Suggestion (uses the same default the evaluator already serializes for `model=None`, `evaluator.py:24`):
```suggestion
        if self.cache_config is None or isinstance(self.model, Model):
            return self.model

        return BedrockModel(model_id=self.model or DEFAULT_BEDROCK_MODEL_ID, cache_config=self.cache_config)
```
(with `from .evaluator import DEFAULT_BEDROCK_MODEL_ID, Evaluator` and `BedrockModel` imported at module level, per `poshinchen`'s comment.) Alternatively raise `ValueError` — either beats silent no-op.

tests/strands_evals/evaluators/test_output_evaluator.py
```diff
@@ -407,3 +407,119 @@ def verify_claim(claim: str) -> str:
...
+        mock_bedrock_instance = Mock()
+        mock_bedrock.return_value = mock_bedrock_instance
+
+        # Re-patch Agent at module level since _resolve_model imports BedrockModel
+        evaluator.evaluate(evaluation_data)
+
+    # BedrockModel should have been created in _resolve_model
+    # and passed to Agent
+    call_kwargs = mock_agent_class.call_args[1]
+    assert call_kwargs["callback_handler"] is None
```
strandly-the-agent, line 452: 🟡 **This test doesn't test what its name says, and it breaks once the import moves.**

It only asserts `callback_handler is None`; `mock_bedrock` is never checked. And after `BedrockModel` becomes a module-level import (agree with `poshinchen`), `patch("strands.models.bedrock.BedrockModel")` no longer intercepts — I verified the `Agent` then receives a real `BedrockModel`. The "Re-patch Agent" comment is stale.

```suggestion
    with patch("strands_evals.evaluators.output_evaluator.BedrockModel") as mock_bedrock:
        evaluator.evaluate(evaluation_data)

    mock_bedrock.assert_called_once_with(model_id="us.anthropic.claude-sonnet-4-6", cache_config=config)
    assert mock_agent_class.call_args[1]["model"] is mock_bedrock.return_value
```

strandly-the-agent (changes requested):
**Changes requested** — the feature itself is sound, but `cache_config` doesn't survive `Experiment` serialization, and it's why CI is red.

- 🔴 **Round-trip breaks in two ways** (one root cause, one fix — inline at `output_evaluator.py:52`): (a) every `Multimodal*Evaluator` now fails `to_dict()→from_dict()` with `TypeError: unexpected keyword argument 'cache_config'` — that's the 5 `test_experiment_from_dict_builtin_evaluators_round_trip[Multimodal*]` failures in the Windows/3.14 job (the others were cancelled, they'd fail too); (b) an `OutputEvaluator` *with* `cache_config` reloads it as a plain `dict`, and `BedrockModel` then dies with `AttributeError: 'dict' object has no attribute 'strategy'` on the first judge call.
- 🟡 `cache_config` is silently dropped when `model` is left at its default `None` (inline at `:84`).
- 🟡 The BedrockModel test asserts nothing about BedrockModel, and its patch target stops intercepting once the import moves to module level (inline at test `:442`).

`poshinchen`'s three comments: agree with all three (`:85`, `:102` → module-level imports; test `:460` → static import). Moving `BedrockModel` to the top is also what makes the test fix above necessary — see that inline. Public ctor surface changed and there's no `needs-api-review` label; flagging for whoever owns that call.

<details>
<summary>✅ Verified / repros / appendix</summary>

**Verified at head `481559f`** (strands-agents 1.55.1, `uv sync --all-extras`)
- `pytest tests/strands_evals/evaluators/test_output_evaluator.py` → 34 passed; `ruff check` clean.
- `pytest tests/strands_evals/test_experiment.py -k round_trip` → **5 failed, 18 passed** on head; **23 passed** on `origin/main`. Failure: `experiment.py:840: TypeError: MultimodalCorrectnessEvaluator.__init__() got an unexpected keyword argument 'cache_config'` (×5 Multimodal evaluators).
- With the private-attribute fix suggested inline (+ dict coercion), `test_output_evaluator.py` + `test_experiment.py` → **151 passed**, ruff clean.

**Repro (b)** — dict after round trip:
```python
ev = OutputEvaluator(rubric="r", model="us.anthropic.claude-sonnet-4-6", cache_config=CacheConfig(strategy="auto"))
data = Experiment(cases=[Case(name="c", input="hi")], evaluators=[ev]).to_dict()
ev2 = Experiment.from_dict(data).evaluators[0]
type(ev2.cache_config)            # dict
ev2._resolve_model()._cache_strategy  # AttributeError: 'dict' object has no attribute 'strategy'
```
Root cause: `experiment.py:834-840` passes serialized keys straight back as ctor kwargs.

**Repro (🟡 None model)**: `OutputEvaluator(rubric="r", cache_config=CacheConfig())._resolve_model()` → `None`; the `Agent` then builds a default `BedrockModel` with `cache_config=None`.

**Reading order**: `output_evaluator.py:44-52` (ctor) → `:84-88` (`_resolve_model`) → `:101-104` (`to_dict`) → `evaluator.py:337-367` (base `to_dict`, why the public attribute leaks) → `experiment.py:834-840` (`from_dict`).

**Appendix — non-blocking (3)**
- `ruff format --check` flags `output_evaluator.py:162` and `:180` (the two `Agent(...)` calls now fit on one line). CI Lint passed only because format isn't in the lint job.
- `test_output_evaluator_evaluate_async_with_cache_config` (test `:507`) builds a real `BedrockModel` (boto3 client) with nothing mocked — works, but it's the only test in the file that touches boto3.
- Docstring at `:29-32`: worth saying plainly that `cache_config` only takes effect when the evaluator builds the `BedrockModel` itself (str / None model); a `Model` instance is passed through untouched. "Up to 90%" is a Bedrock pricing claim — fine, but it belongs in the PR body rather than the API docstring.

Specialist passes (api-bar-raiser, reviewer) were started but didn't finish in time and were cancelled; the findings above are all from my own verification.
</details>

Issue #310, opened 2026-07-14
[FEATURE] stickler ↔ Strands Eval integration


poshinchen:
### Problem Statement

Enhanced UX for comparators and StructuredOutputEvaluator by integrating with [Stickler](https://github.com/awslabs/stickler)

More details in:
https://github.com/awslabs/stickler/discussions/164

### Use Case

N/A

sromoam:
Hi @poshinchen , I'd be delighted to assist with this issue!

PR #400, opened 2026-09-14
Structured Evaluator for Strands-Eval with Stickler 

sromoam:
## Description

Adds `StructuredOutput`, a deterministic evaluator that scores structured agent output against ground truth **field by field**, with [`stickler-eval`](https://pypi.org/project/stickler-eval/) as the comparison engine behind an optional extra.

This is the integration proposed in #310. Happy to reshape any of it, @poshinchen — you're assigned there, so treat the design choices below as arguments rather than decisions.

### Why

`Equals` compares structured output with whole-object `==`, so it scores 0.0 or 1.0. An extraction that gets nine of ten fields right is indistinguishable from one that gets none right, a list in a different order counts as wrong, and neither result tells you which field to fix.

Measured, on an invoice model with a nested `list[LineItem]`:

| | `Equals` | `StructuredOutput` |
|---|---|---|
| list reordered, content identical | `0.0` | `1.0` |
| one quantity wrong, list reordered | `0.0` | `0.9375` |

The first row is the clearest case: the content is identical, so the extraction is correct, and `Equals` calls it a total failure. In the second row `reason` reads `weakest fields: line_items=0.75`, and `metrics()` reports counts against the nested paths (`line_items.sku`, `line_items.quantity`) so a suite tells you *which field* the agent gets wrong across the dataset.

Deterministic and offline: no LLM judge, no credentials, no per-call cost. Configuration is inferred from the Pydantic model the agent already passes as `structured_output_model`, so a caller annotates nothing, and every inferred decision is inspectable via `explain()`.

```python
from strands_evals.evaluators import StructuredOutput

evaluator = StructuredOutput(Invoice)          # the model the agent already emits
report = Experiment(cases=cases, evaluators=[evaluator]).run_evaluations(task)

report.overall_score      # weighted mean across the dataset
evaluator.per_case()      # per-document field scores, plus precision/recall/f1
evaluator.metrics()       # per-field confusion matrix, keyed by dotted path
evaluator.explain()       # the comparator chosen per field, and why
```

### Design notes

**Detail is read from the evaluator, not squeezed through `EvaluationOutput`.** That type has four scalar fields and cannot express a per-field breakdown, so `evaluate()` returns one output per case and the detail is served from the retained comparison: `per_case()` for per-document field scores, `metrics()` for the dataset rollup. Both read comparisons already performed, so neither runs a second pass over the data.

**`metrics()` partitions by model class.** Feeding two schemas into one rollup unions their field paths, which makes a field present in half the documents read as missed in the rest. Classes that share a `__name__` get separate rollups.

**Concurrency.** `run_evaluations_async` defaults to `max_workers=10` and invokes evaluators through `asyncio.to_thread`, so anything accumulated across cases is shared mutable state on many threads. Accumulation is append-only: `list.append` is atomic under the GIL, a read-modify-write is not guaranteed to be, and the list is read afterwards on one thread. This is stated as a correctness argument rather than a measurement — a read-modify-write mutant did *not* lose results across 1000 cases at 32 workers on CPython 3.12, so a race test cannot stand in for it, and `test_the_accumulator_is_append_only` pins the structure instead. `evaluate_async` is deliberately not overridden; the base class's `asyncio.to_thread` is what puts the comparison on worker threads at all.

**Placement.** `evaluators/deterministic/`, alongside `Equals`, since it is deterministic and it is the evaluator `Equals` users graduate to. Exposed as `from strands_evals.evaluators import StructuredOutput`, matching how `Equals` is exposed rather than adding a top-level export.

### `test_pass`

`test_pass` requires both the weighted score and `recall` to clear `match_threshold`. Score alone is not enough: stickler credits a field absent on both sides as a match, so on a sparse schema the score can clear the bar on correct absences alone. Recall only counts fields that had a value to find, which is what [stickler recommends gating on](https://awslabs.github.io/stickler/Getting-Started/thresholds-and-metrics/#sparse-objects) for that shape. A document whose ground truth is legitimately blank still passes.

It remains a document verdict, not a per-field one, so a document can pass with one field wrong. `per_case()` returns `precision`, `recall` and `f1` alongside the field scores for a stricter gate.

## The other half of #310: the comparators

#310 asks for enhanced UX for **comparators** *and* a structured-output evaluator. This PR delivers the evaluator. The comparators are reachable through it today, but only by one route, and I want to check with you whether closing that gap belongs here or in a follow-up.

**What works now, with no new code in this PR.** stickler ships eleven comparators — `ExactComparator`, `LevenshteinComparator`, `NumericComparator` (tolerance), `DateComparator` (format- and precision-aware), `FuzzyComparator` (token matching), `NormalizedComparator`, `ANLSStarComparator` (structural), `PhoneComparator` (per-region libphonenumber rules), `BBoxIoUComparator`, `StructuredModelComparator`, and `SemanticComparator`. If you declare your output model as a stickler `StructuredModel`, the evaluator honours every per-field choice, because a `StructuredModel` *is* a Pydantic model and so keeps driving the agent's `structured_output_model`:

```python
from strands_evals.evaluators import StructuredOutput
from stickler import ComparableField, StructuredModel
from stickler.comparators import ExactComparator, NumericComparator, PhoneComparator

class Contact(StructuredModel):
    account_id: str = ComparableField(comparator=ExactComparator(), weight=3.0)
    phone: str | None = ComparableField(default=None, comparator=PhoneComparator(region="US"))
    balance: float | None = ComparableField(default=None, comparator=NumericComparator(tolerance=0.01))

evaluator = StructuredOutput(Contact)
evaluator.explain()
# account_id   ExactComparator      weight=3.0  (explicit)
# phone        PhoneComparator      weight=1.0  (explicit)
# balance      NumericComparator    weight=1.0  (explicit)
```

Measured on that model: ground truth `phone="+1 415-555-2671"`, `balance=1247.50` against a prediction of `phone="4155552671"`, `balance=1247.4999` scores **1.0**. The same number written two ways is a match, and float representation noise is inside tolerance — both are `0.0` under `Equals`. Note `source` reads `explicit` rather than `inferred`, so `explain()` still shows exactly where each decision came from.

Comparator reference: [Comparators guide](https://awslabs.github.io/stickler/Guides/Comparators/) · [Dates](https://awslabs.github.io/stickler/Guides/Comparators/date-comparator/) · [Phone/postal](https://awslabs.github.io/stickler/Guides/Comparators/postal-codes/) · [Choosing a configuration path](https://awslabs.github.io/stickler/Getting-Started/choosing-a-configuration-path/)

**What does not work, and is the actual gap.** There is no way to override a single field from the *evaluator's* side. A caller holding a plain `BaseModel` who disagrees with one inferred choice has to restructure their model into a `StructuredModel` to fix it. What is missing is roughly:

```python
StructuredOutput(Invoice, overrides={"vendor_name": {"comparator": FuzzyComparator(), "weight": 2.0}})
```

That needs a parameter on this evaluator **and** a matching one upstream: stickler's `eval_for(cls, *, weight_hints, match_threshold)` accepts no per-field overrides today, so it is a two-repo change rather than a local one.

**Question: in this PR, or a separate one?** My inclination is separate, for three reasons — this PR is already ~1000 lines and adding a public API surface makes it harder to review as one unit; the override shape is a design decision worth its own discussion (dict-of-dicts vs. a typed object, and whether it keys on dotted paths for nested fields); and it needs the upstream `eval_for` change first, which I would rather land and release in stickler before depending on it here. But it is your issue and your repo's UX, so if you would rather see #310 closed by a single PR I am happy to do the upstream change and fold it in here.

**Deliberately not proposed either way:** `SemanticComparator` (Bedrock embeddings), BERTScore, and stickler's LLM judge. They reintroduce per-call cost, latency and run-to-run variance, which is what a *deterministic* evaluator exists to avoid. They stay behind stickler's own optional extras and are opt-in for anyone who wants them.

## Related Issues

- Implements #310 (`[FEATURE] stickler ↔ Strands Eval integration`)
- Design proposal and prior discussion: https://github.com/awslabs/stickler/discussions/164
- Supersedes awslabs/stickler#251, which put the same evaluator on the *stickler* side as `stickler.integrations.strands_evals`. Moving it here is better: the evaluator implements this repo's `Evaluator` contract, so it belongs with the other evaluators and can be maintained against changes to that contract.

## Documentation PR

None. The evaluator is documented in this repo's `README.md` the way the other evaluators are, plus module and method docstrings. `AGENTS.md` says docs are not vendored yet and the user guide lives at strandsagents.com — if that is a separate repo, point me at it and I will open the matching PR.

Runnable notebooks live in the stickler repo (awslabs/stickler#358) rather than here, deliberately: this repo vendors no notebooks and its README points contributors at [strands-agents/samples](https://github.com/strands-agents/samples), so adding an `examples/` tree here would introduce a convention rather than follow one. Happy to move them into `samples` if you'd prefer them closer.

## Type of Change

New feature

## Testing

`hatch run prepare` passes: ruff clean, mypy clean across 236 source files, **2177 tests passed**.

Detail on the new tests, 37 of them, at `tests/strands_evals/evaluators/deterministic/test_structured_output.py`, skipped via `importorskip` when the extra is absent:

- They cover the claims the harness does not enforce: that the case score is stickler's weighted score rather than a mean, that per-field detail needs no second comparison pass, that a mixed-schema suite does not produce a merged rollup, and that no result is lost at the harness's default concurrency.
- Five of them pin `test_pass` behaviour on sparse schemas, and they fail against `stickler-eval` 0.7.0, which is what makes the declared floor enforced rather than nominal.
- `stickler` is added to `hatch-test.features` and `hatch-static-analysis.features`, so CI installs the extra. Without it `importorskip` skips every one of these on every interpreter and `mypy` fails on `import-not-found`.

## Checklist
- [x] Any dependent changes have been merged and published — `stickler-eval` 1.0.0 is on PyPI

## The dependency

Scoring comes from `stickler-eval`, declared as the optional `stickler` extra:

```toml
stickler = ["stickler-eval>=1.0.0,<2.0.0"]
```

Apache-2.0, an awslabs project, 1.0.0 published 2026-09-14. Optional: nothing changes for anyone who does not install the extra.

The floor is `1.0.0` rather than `0.7.0` because the verdict semantics this evaluator relies on changed in 1.0.0, and its tests and docstrings assume the newer ones. Two 0.7.0 defects the evaluator was working around also landed in 1.0.0: `explain()` misattributing provenance for a refused name-token rule (awslabs/stickler#250), and a nested-array double-unwrap that scored an identical `list[list[float]]` pair `0.0` (awslabs/stickler#324).

## Known follow-ups

Found in review, disclosed rather than folded in to keep the diff focused. Happy to fix any of these here if you'd rather they not land as follow-ups:

- `Experiment.to_file()` raises on an experiment carrying this evaluator: `model_cls` is stored as a public attribute, so the inherited `to_dict` emits a live class object. The fix follows the `OutputEvaluator` precedent (store privately, emit a dotted path). `StructuredOutput` also needs registering in the `from_dict` evaluator registry.
- A case whose `expected_output` is `None` raises inside `evaluate`, which the harness turns into a score-0 row with no detail, so `metrics().document_count` and the report's row count disagree without disclosing the skip. Returning a `NOT_APPLICABLE` row instead would match the sibling evaluators.
- With `model_cls` omitted, a `dict` or JSON-string `expected_output` is not used for schema inference, so the class is taken from the prediction — which can silently score `1.0` for a prediction that omitted fields.
- The `stickler` import is eager at module scope, so `import strands_evals` pays for it whenever the extra is installed (measured: 1.55s vs 0.70s). A module-level `__getattr__` in `evaluators/__init__.py` would make it lazy.

----


sromoam:
Adding one more piece of supporting info, we have this PR ready with notebook samples demonstrating this integration. Not sure where these best belong! https://github.com/awslabs/stickler/pull/359 @mkmeral cc @poshinchen

poshinchen:
@strandly-the-agent review this PR which allows strands-evals to use stickler as optional dependencies. Check if there's any improvements and potential issues. Review the devX too.

src/strands_evals/evaluators/deterministic/structured_output.py
```diff
@@ -0,0 +1,458 @@
...
+        model_cls: type[BaseModel] | None = None,
+        *,
+        match_threshold: float = 0.7,
+        weight_hints: bool = False,
+        name: str | None = None,
+    ) -> None:
+        """Initialize the evaluator. See the class docstring for argument detail."""
+        _require_stickler()
+        super().__init__(name=name)
+        self.declared_cls = model_cls
```
strandly-the-agent: **`Experiment.to_file()` raises and the evaluator can't round-trip through `from_dict`.** Base `to_dict` (evaluator.py:337) emits every public attr by name, so this produces `{'declared_cls': <class 'Invoice'>}` → `TypeError: Object of type ModelMetaclass is not JSON serializable`; and because the attr name differs from the ctor param, `from_dict(..., custom_evaluators=[StructuredOutput])` fails with `unexpected keyword argument 'declared_cls'`. Reached by anyone calling the documented `Experiment.to_file()`.

Suggestion: store privately and follow the `OutputEvaluator` precedent (output_evaluator.py:49-66): `self._model_cls = model_cls`, override `to_dict` to emit `model_cls` as a dotted path (`f"{cls.__module__}.{cls.__qualname__}"`), resolve it back in `__init__` when a string arrives, and add `"StructuredOutput": StructuredOutput` to the registry at experiment.py:795. You offered to fold this in — I'd take that; it's a core flow.

src/strands_evals/evaluators/deterministic/structured_output.py
```diff
@@ -0,0 +1,458 @@
...
+        if self.declared_cls is not None:
+            return self.declared_cls
+        # expected_output first, deliberately. Ground truth defines the fields being
+        # measured; the agent's output does not get to narrow that. If it is
+        # inferred from `actual_output` instead, an agent returning a model with
+        # fewer fields silently drops the missing ones from the comparison and
+        # scores 1.0 -- a perfect result for output that omitted a field, which is
+        # the exact failure this evaluator exists to catch. Preferring expected
+        # makes `_coerce` raise on that input instead.
+        for value in (evaluation_case.expected_output, evaluation_case.actual_output):
```
strandly-the-agent: **Inferred mode scores every case 0 as soon as cases come from JSON.** `Experiment.from_file()` datasets and a warm `LocalFileTaskResultStore` both hand back `expected_output`/`actual_output` as `dict`, so this loop finds no `BaseModel` and raises → harness records `Evaluator error: … could not infer a model class` on every row. Repro: `run_evaluations(task, evaluation_data_store=LocalFileTaskResultStore(d))` twice with `StructuredOutput()` → run 1 `overall=1.0`, run 2 `overall=0.0`.

Question: would it be simpler to make `model_cls` required? Inferred mode only works for in-memory Pydantic instances, and dropping it also removes `_resolve_cls`, `_distinct_names`, the `explain()` ambiguity branch and the mixed-schema partitioning in `metrics()` (~60 lines). A user with two schemas can run two evaluators.

src/strands_evals/evaluators/deterministic/structured_output.py
```diff
@@ -0,0 +1,458 @@
...
+        # cases at 32 workers on CPython 3.12, so a race test cannot stand in for
+        # it. `test_the_accumulator_is_append_only` pins the structure instead.
+        #
+        # `evaluate_async` is deliberately NOT overridden. The base class's is
+        # `await asyncio.to_thread(self.evaluate, ...)`, which is what puts this on
+        # worker threads at all; a synchronous override would run the comparison
+        # inline on the event loop, making the sentence above false and the
+        # concurrency test vacuous, and would block every other case's task on a
+        # CPU-bound Hungarian match.
+        self._results: list[_CaseResult] = []
```
strandly-the-agent: **Reusing one instance silently blends runs.** Nothing in the harness clears this, so `exp.run_evaluations(task)` twice on 1 case gives `metrics()['Invoice'].document_count == 2` and `per_case()` with 4 entries for 2 cases, while the second report says 0.0 (repro `adv-t3_double_run.py`). `reset()` exists but the README never mentions it, and the class docstring doesn't either.

Separately, rows the harness turned into `Evaluator error` (None / prose / bad JSON) never reach `_results`, so `metrics()` recall is over fewer documents than the report with nothing disclosing the gap.

Suggestion: at minimum one README line ("call `evaluator.reset()` between runs") and a `Note:` in the class docstring; better, catch `ValidationError`/`TypeError` from `_coerce(actual_output)` and return a normal score-0 row with `reason="actual_output did not validate as Invoice: …"` so report and `per_case()` stay in step.

README.md
```diff
@@ -500,6 +503,71 @@ class PolicyComplianceEvaluator(Evaluator[str, str]):
...
+annotate. Pass `model_cls` for a single-schema suite and any foreign shape raises;
+omit it and the class is inferred per case, with `metrics()` partitioned by class.
+
+**On sparse schemas, read `recall` or `f1`, not the score alone.** `score` credits a
+field absent on *both* sides with 1.0 -- a value the model correctly left blank is a
+value it got right -- so on a schema where most fields are usually empty those fields
+outvote the informative ones. A 10-field model whose ground truth populates 2 fields
+scores `0.80` against a prediction that returned *nothing*. `test_pass` therefore
+requires both the score and `recall` to clear `match_threshold`, and `per_case()`
+exposes `recall`, `precision` and `f1` so the verdict is auditable. Raising `weight`
```
strandly-the-agent: **"Raising `weight`" isn't possible on the plain `BaseModel` this section is about.** stickler 1.0.0's `eval_for` ignores `Field(json_schema_extra={"weight": 5.0})` — `explain()` still reports `weight=1.0 source=type`. Only a `stickler.StructuredModel` with `ComparableField(weight=…)`, or `weight_hints=True`, changes weights.

```suggestion
exposes `recall`, `precision` and `f1` so the verdict is auditable. To make `score`
itself reflect the fields you care about, declare the model as a stickler
`StructuredModel` and set `ComparableField(weight=...)` on them, or pass
`weight_hints=True` to weight ids and amounts by name.
```

strandly-the-agent (changes requested):
**TL;DR: solid evaluator, tests are real, README example runs — but four public-surface things should land before merge rather than as follow-ups, and one is a naming call for `poshinchen`.** Verified at `4b51798`: 37/37 tests, ruff + mypy clean, README numbers (0.9375 / `weakest fields: line_items=0.75`) reproduce, `StructuredOutput()` without the extra gives the right ImportError and `import strands_evals` still works.

1. 🟡 `Experiment.to_file()` raises and `from_dict` can't rebuild it — even with `custom_evaluators=[StructuredOutput]` (`unexpected keyword argument 'declared_cls'`). Author-disclosed, but it's a documented core flow; small fix, fold it in.
2. 🟡 `model_cls=None` (inferred mode) scores **every case 0** whenever cases come from JSON — `Experiment.from_file()` datasets or a warm `LocalFileTaskResultStore`. I'd make `model_cls` required; it also deletes ~60 lines.
3. 🟡 Accumulator never resets: reusing one instance doubles `metrics().document_count`; README never mentions `reset()`.
4. 🟡 README tells readers to "raise `weight`" — not possible on a plain `BaseModel` with stickler 1.0.0 (stays 1.0).

**Question for `poshinchen`:** the name. Siblings are predicates (`Equals`, `ToolCalled`) and the SDK already owns `structured_output`; `StructuredOutput` reads like a type. A rename after release is breaking, so decide now (e.g. `FieldsMatch`, `StructuredEquals`). Also: `validate-pr-title` is red (needs a `feat:` prefix), and I agree `overrides=` belongs in a follow-up.

<details><summary>Details, evidence, appendix</summary>

### ✅ Verified
- `pytest tests/.../test_structured_output.py` → 37 passed (py3.13, stickler-eval 1.0.0); `ruff check`/`ruff format --check`/`mypy` clean on changed files.
- README example (with a stub task) → `overall_score 0.9375`, reason `weakest fields: line_items=0.75`, `explain()` covers nested paths.
- No-extra venv: `import strands_evals` OK; `StructuredOutput()` → `StructuredOutput requires the 'stickler-eval' package. Install it with: pip install "strands-agents-evals[stickler]"`. Extras/hatch wiring is right; stickler-eval requires-python `>=3.10` matches.

### Repros (all at `4b51798`)
- **1** `Experiment(cases, evaluators=[StructuredOutput(Invoice)]).to_file(p)` → `TypeError: Object of type ModelMetaclass is not JSON serializable`. `Experiment.from_dict({..."evaluator_type":"StructuredOutput"...})` → `Cannot find StructuredOutput`; with `custom_evaluators=[StructuredOutput]` → `TypeError: __init__() got an unexpected keyword argument 'declared_cls'` (base `to_dict` emits attrs by name, `from_dict` feeds them back as kwargs).
- **2** `run_evaluations(task, evaluation_data_store=LocalFileTaskResultStore(d))`, `model_cls` omitted: run 1 `overall=1.0`; run 2 (cache warm, outputs come back as `dict`) `overall=0.0`, every reason `Evaluator error: StructuredOutput could not infer a model class…`. Same for any `from_file()` dataset since `expected_output` is a dict there.
- **3** one instance, two `run_evaluations` of 1 case → `document_count=2`, `len(per_case())=2`; run-2 report says 0.0 while `metrics()` blends both runs. Also: harness error rows (None / prose / bad JSON `actual_output`) never reach `_results`, so `metrics()` recall is computed over fewer documents than the report (2 of 4 in the probe) with nothing disclosing the gap.
- **4** `class M(BaseModel): a: str = Field(json_schema_extra={"weight": 5.0})` → `explain()` reports `weight=1.0 source=type`. Only `stickler.StructuredModel` + `ComparableField(weight=…)` or `weight_hints=True` changes weights.

### DevX notes
- Eager `from stickler import …` at `structured_output.py:32`, pulled in by `deterministic/__init__.py:4`, costs **457 ms of a 2553 ms `import strands_evals`** when the extra is installed (`python -X importtime`). Importing inside `__init__`/`_spec_for`/`metrics` avoids a package-level `__getattr__` and matches how `providers/__init__.py` keeps optional deps off the import path.
- An agent that returns prose gets `Evaluator error: 1 validation error for Invoice  Invalid JSON…` (score 0, no `per_case` row). That's a legitimate eval failure, not an evaluator crash — catching `ValidationError` on `actual_output` and returning a score-0 row with `reason="actual_output did not validate as Invoice: …"` would read better and keep `per_case()`/report in step.
- `StructuredOutput(Node)` for a recursive model is accepted by the constructor and then fails every case at evaluate time (stickler: `does not support recursive models`). Calling `self._spec_for(model_cls)` in `__init__` when given would fail fast.
- `metrics()` is typed `dict[str, Any]` but returns `dict[str, stickler.utils.process_evaluation.ProcessEvaluation]` — fine, but say so in the docstring so readers know it's stickler's object.

### Reading order
`structured_output.py` `__init__`/`evaluate`/`_passed` → `_resolve_cls`/`_coerce` → `metrics`/`_distinct_names` → tests `TestPassRequiresFindingSomething`, `TestMixedSchemas` → README section.

### Appendix — non-blocking (4)
- ⚪ `validate-pr-title` failing: title needs the conventional-commit form, e.g. `feat: add StructuredOutput evaluator (stickler extra)`.
- ⚪ `_passed` reads `result.raw["confusion_matrix"]["overall"]` — a private-ish shape of stickler's result; fine under `<2.0.0`, worth a one-line note.
- ⚪ Comment density is far above `output.py` (several 10–20 line block comments narrating design history, e.g. lines 205–221, 241–244, 276–285). The PR body already holds that reasoning; trimming would match house style.
- ⚪ stickler-owned, not this repo: `RootModel[list[…]]` → `AttributeError: 'list' object has no attribute 'keys'`; a `float` field with `nan` scores an identical pair 0.0; `extra="forbid"` models raise on a wider sibling. Worth issues on awslabs/stickler if you hit them.

Repro scripts + logs are attached to this PR's card in the portal (`repro_pr400.py`, `readme_example.py`, `importerror_without_extra.log`, `adv-t1…t4`).
</details>

tests/strands_evals/evaluators/deterministic/test_structured_output.py
```diff
@@ -0,0 +1,534 @@
...
+import asyncio
+import inspect
+import threading
+
+import pytest
+from pydantic import BaseModel, Field, ValidationError
+
+pytest.importorskip("stickler", reason="requires the stickler extra")
+
+from strands_evals import Case, Experiment  # noqa: E402
```
poshinchen: Are these ignore required?

poshinchen:
@strandly-the-agent I think the evaluator.explain / per_case / metrics do not have to be in the StructureOutput Evaluator Class, instead it should be the extension of the Report class in my opinion. Analyze and tell me which approach is better.

strandly-the-agent:
**Agree for `per_case()` and `metrics()`: they are results, and results belong in the report. `explain()` is different — it's configuration derived from the model class, not from any run — so it stays with the evaluator (it's a passthrough to stickler's `eval_for(cls).explain()`).**

Why the report side wins: every stateful defect I found yesterday (reuse doubles `document_count`, `metrics()` blends runs, harness error rows missing from the rollup, `per_case()` completion-ordered vs `report.scores` index-ordered, the `reset()` chore, the thread-safety argument) exists only because the evaluator keeps an accumulator. A report is built per run, index-aligned with cases, includes the score-0 error rows, survives `to_file`/JSON, shows up in `report.display()`, and works after `EvaluationReport.flatten`. The evaluator-side design can't get any of that without re-implementing the harness.

The catch: `EvaluationOutput` has four scalars and silently drops extra fields (verified: `extra` policy is Pydantic's default `ignore`), so the report has nowhere to put field-level numbers today. Minimal shape that works:

1. Add `metadata: dict[str, Any] | None = None` to `EvaluationOutput` (additive, same precedent as `label`; telemetry only emits aggregate score/reason/label, so nothing leaks).
2. `StructuredOutput.evaluate` returns one row per case with `metadata={"field_scores", "precision", "recall", "f1", "comparison": raw}`. Measured on the README `Invoice`: ~2.9 KB/case for `comparison` (needed for the rollup), ~140 B without it.
3. The rollup becomes a pure function over a report — `StructuredOutput.metrics(report)` (static), filtering rows by `cases[i]["evaluator"]` and feeding `metadata["comparison"]` to `aggregate_from_comparisons`. A `Report` subclass doesn't work without harness changes, because `run_evaluations` and `flatten` construct `EvaluationReport` directly — a function over the report is the "extension" that fits.
4. Delete `_results`, `_CaseResult`, `reset()`, `_distinct_names`, the concurrency tests and comments (~150 lines).

<details><summary>Trade-offs and the alternative I'd not take</summary>

**Costs of the report approach**
- One small change to a core type (`EvaluationOutput.metadata`). It's additive; existing reports gain a `"metadata": null` key when dumped.
- Report size grows ~3 KB per case for this evaluator only (10k cases ≈ 30 MB). Could gate the raw comparison behind `retain_comparison=True` if that matters, keeping `field_scores`/P/R/F1 always.
- `metrics(report)` recomputes on each call instead of reading state — same cost as today, `metrics()` already calls `aggregate_from_comparisons` every time.

**Zero-core-change alternative: one `EvaluationOutput` per field.** `label=<dotted path>`, `score=<field score>`, and since `self.aggregator` is an instance attribute (evaluator.py:53) the evaluator could bind a weighted aggregator that looks weights up by label. It works for flat models, but nested list fields have no per-leaf score, the recall gate on `test_pass` has no row to live in, and `detailed_results` swells to N rows per case. The author already tried this shape and backed out for the same reasons (`test_score_is_sticklers_weighted_overall` docstring). Not recommended.

**What stays on the evaluator**
- `explain()` — config, not results; or drop it and point README at `stickler.eval_for(Invoice).explain()`.
- `_passed` / recall gate — that's per-case scoring, correctly in `evaluate`.

**Interaction with my earlier review.** This resolves finding 3 entirely and makes 2 (`model_cls` required) even more attractive, since mixed-schema partitioning in `metrics()` was the main reason `_distinct_names` existed; with `model_cls` required the rollup is one schema per evaluator instance, and two schemas are two evaluators with distinct `name=`.
</details>

sromoam:
@poshinchen thanks — agreed. `per_case()` and `metrics()` are run results and belong on the report. `explain()` is config derived from the model class, so it stays on the evaluator.

**We need you to weigh in on one thing it costs.** A report row has nowhere to put per-field numbers — `EvaluationOutput` is four scalars and defaults to `extra="ignore"`. The move needs one line in `types/evaluation.py`:

```python
metadata: dict[str, Any] | None = None
```

Additive and optional, same precedent as `label`. Applied locally, the unit suite passes unchanged (2164 tests); existing reports just gain a `"metadata": null` key. Subclassing `EvaluationOutput` instead doesn't work — the subclass survives in memory, then serialization follows the declared `list[list[EvaluationOutput]]` and drops the extra fields silently, so `strands-evals run --output` would show nothing. `ConfigDict(extra="allow")` is an equally small alternative if you prefer it.

**If this blocks or needs more discussion, we can defer.** The evaluator then ships with detail on the instance, as today, and a stated limitation: `metrics()` / `per_case()` work in Python but not in the serialized report, so CLI users get score, pass, reason and label only. That goes in the class docstring and README, with a one-time `logger.warning` on first use so nobody finds it by opening an empty `report.json`. `reset()` gets documented for instance reuse. One cost of deferring: if the field lands later, `metrics()` changes from an instance method to a static function over a report, which is a breaking change for anyone who adopted it in between.

**On the name.** I'd like to keep `StructuredOutput`. `structured_output_model` is the parameter callers already pass to the SDK, so the evaluator is named after the thing it scores, and that's the term a user searching for this will have in mind. Happy to be overruled — it's your API surface and I'd rather settle it now than rename post-release.

**Proceeding now on everything that doesn't depend on the `metadata` decision:** `model_cls` required (fixes inferred mode scoring every case 0 from JSON, by deleting the mode, and lets the constructor fail fast on recursive models), the `to_dict`/`from_dict` round-trip plus registry entry, catching `ValidationError` so a bad prediction becomes a score-0 row instead of a crash, the lazy stickler import (457 ms of import time), the README `weight` correction, and a `feat:` retitle to clear `validate-pr-title`. Also from the appendix: a note on the `result.raw["confusion_matrix"]` shape, trimming the block comments that narrate design history, and the `metrics()` return-type docstring. I'll open issues on awslabs/stickler for the three upstream defects you found (`RootModel[list[…]]`, `nan`, `extra="forbid"`).

On the test `# noqa: E402` — not required. Swapping the module-level `importorskip` for `pytestmark = pytest.mark.skipif(find_spec("stickler") is None, ...)` puts the imports back at the top of the file, matching `AGENTS.md`. Fixed in this round.

poshinchen:
@strandly-the-agent analyze it: Related to the `evaluator.explain()`: The problem that i'm seeing here is that you'll not be able to view the config again from the report. But if the evaluator is serializable (at least for the config), we should be able to get the access from `config.evaluators.strcturedOutput.explain()` @sromoam?

strandly-the-agent:
**Yes, that works, and it's the right split: results in the report, config in the experiment file. It depends on exactly the round-trip fix `sromoam` is already doing (finding 1), and it needs one guard.**

I prototyped it at `4b51798` with a stand-in for the fix (`to_dict` emits `model_cls` as a dotted path, `__init__` re-imports a string): `Experiment.to_file(p)` → `{'evaluator_type': ..., 'match_threshold': 0.9, 'model_cls': 'models.Invoice'}`, then `Experiment.from_file(p, ...).evaluators[0].explain()` returns the identical per-field config (`==` the original). `explain()` is a pure function of `(model_cls, match_threshold, weight_hints)`, so nothing else needs to be persisted — and this also covers stickler `StructuredModel`s with explicit `ComparableField`s, since that config lives on the class too. Access is `loaded.evaluators[i]` (or match on `report.cases[i]["evaluator"]` → `get_name()`); there's no `config.evaluators.<name>` attribute path today, and I wouldn't add one for this.

<details><summary>The guard, and two small caveats</summary>

**Guard — the dotted path must be importable where it's read.** A model defined in `__main__`, a notebook cell, or inside a function serializes as `__main__.Invoice` / `mod.f.<locals>.Invoice` and can't be re-imported by the CLI or another process. Same situation `OutputEvaluator.to_dict` handles for tools: keep serializing, but `logger.warning` at save time when `__module__ == "__main__"` or `"<locals>"` is in the qualname, and raise a clear `ImportError` naming the path at load. The README example currently defines `Invoice` inline, so the doc should say "put the model in an importable module if you want to reload the experiment".

**Caveat 1 — it's a recomputation, not a snapshot.** Reloading re-runs stickler's inference under whatever `stickler-eval` is installed. Inside the `>=1.0,<2.0` pin a minor release could change an inferred comparator or threshold, so `explain()` after reload describes what *would* run now, not necessarily what ran. If exact provenance matters (audits, regression triage), the cheap fix is to snapshot `explain()` once per run into the report rather than per row — but I'd not do that in this PR; the experiment file plus a pinned lockfile is the normal answer.

**Caveat 2 — the report doesn't point at its experiment file.** `EvaluationReport` carries evaluator `name`/`evaluator_type` per row and nothing else, so the user has to keep the two files together. That's already true for every other evaluator's config (`match_threshold`, rubrics, model ids) — not specific to this one.

Prototype + output are attached to the PR card (`explain-roundtrip-prototype.py/.log`).
</details>

sromoam:
@poshinchen resolved at `77d028f`. PR description rewritten.

**Your four:**

- **`metadata` on `EvaluationOutput`** — added as a named field, not `extra="allow"`.
- **`metrics()` / `per_case()`** — now functions of a report, reading each row's `metadata`. The accumulator, `reset()`, `_CaseResult` and `_distinct_names` are gone, and with them three defects: reuse blending runs, failed cases missing from the rollup, and completion-ordered rows. A test asserts the rollup is identical on a report reloaded from JSON.
- **`explain()`** — stays on the evaluator. For the follow-up: 1548 bytes on a 4-field model, identical per row, so it wants a report-level field rather than `metadata`.
- **The name** — `StructuredOutputSimilarity`. `Matches` implies exactness this does not have: a reordered list scores 1.0, and `phone="+1 415-555-2671"` vs `"4155552671"` scores 1.0.

**Also fixed:** `model_cls` required, killing the mode that scored 0 on file-loaded and cached cases. `to_file`/`from_dict` round-trip plus registry entry. Unparseable output as a score-0 row, not an exception. `stickler` off the import path (~270 ms). README `weight` correction.

**Two more, found after your comments:**

- **Two false passes in `test_pass`.** stickler classifies a wrong value as `fd`, not `fn`, so `tp + fn == 0` was not "nothing to find". A fully populated ground truth, entirely mis-extracted, passed at score 0.80 / recall 0.00. Fabricating 5 fields for a blank 20-field document passed at 0.75. The count is `tp + fn + fd`, and invented values now fail when there was nothing to find.
- **Nested model classes did not reload.** `to_dict()` emits `__qualname__`, so `Outer.Inner` was read as a module name and `from_file()` raised `ModuleNotFoundError`.

**Known limit, documented:** `test_pass` gates on recall, not precision, so finding everything *and* inventing extras passes. A precision gate would fail an agent for extracting a field sparse ground truth merely omitted. `precision` is on every row.

**Housekeeping:**

- `# noqa: E402` not required — `pytestmark` with `find_spec` replaces the module-level `importorskip`.
- `hatch run prepare` green: ruff and mypy clean, 2190 tests on 3.10–3.14. The higher 3.14 warning count is the existing `asyncio.iscoroutinefunction` deprecation from #50.
- CI needs a workflow-run approval.

Want a line in `AGENTS.md` that the `Evaluator` suffix means LLM-based? It would have saved the naming round trip.

poshinchen:
@strandly-the-agent review the updated changes, check the round-trip and new EvaluationReport's methods (metrics, ...). We might need the getter to have easier access?

src/strands_evals/evaluators/deterministic/structured_output.py
```diff
@@ -0,0 +1,448 @@
...
+            carry no comparison, so they are absent here while still appearing in the
+            report as score-0 rows.
+        """
+        _, aggregate_from_comparisons = _stickler()
+        comparisons = [
+            row.metadata["comparison"]
+            for _, row in _rows(report, evaluator)
+            if row.metadata and "comparison" in row.metadata
+        ]
+        return aggregate_from_comparisons(comparisons)
```
strandly-the-agent, line 365: **`evaluator=None` on a flattened multi-schema report silently unions field paths.** With `[SOS(Invoice, name="inv"), SOS(Receipt, name="rec")]` in one experiment, `SOS.metrics(report)` returns `document_count 2` with both schemas' paths and `cm_recall` of `0.0` for `line_items`/`total_amount` — fields that were never wrong. The docstring warns, but this is exactly the mixed-schema case the README tells users to set up, and the first thing they'll type is `metrics(report)`. Reached via the documented API with no error.

Pre-tested (50 tests pass, ruff clean); `evaluator="inv"` still works as before:

```suggestion
        _, aggregate_from_comparisons = _stickler()
        scored = [
            (case, row) for case, row in _rows(report, evaluator) if row.metadata and "comparison" in row.metadata
        ]
        names = sorted({str(case.get("evaluator")) for case, _ in scored})
        if len(names) > 1:
            raise ValueError(
                f"report carries comparisons from {len(names)} evaluators ({', '.join(names)}); "
                "pass evaluator=<name> to aggregate one schema at a time"
            )
        return aggregate_from_comparisons([row.metadata["comparison"] for _, row in scored])
```

strandly-the-agent (approved):
**TL;DR: approving at `53f21fb`. All four previous findings are fixed and verified; the report-derived design works end to end (including through JSON). One small guard I'd add before merge, and a direct answer to the getter question.**

**Previous findings:** F1 round-trip ✅ (`to_file` → `from_file` with no `custom_evaluators`, nested `Outer.Inner` reloads, `explain()` identical) · F2 JSON-sourced cases ✅ (`from_file` dataset + warm `LocalFileTaskResultStore` both score 0.9821) · F3 statefulness ✅ (two runs → `document_count` 3/3 per report; prose/None outputs are score-0 / N/A rows and the overall drops N/A) · F4 README `weight` ✅. Also: `stickler` is off the `import strands_evals` path, 50/50 tests, ruff + mypy clean, `report.display()` fine with `metadata`, rollup identical after `model_dump_json` → `model_validate_json` (numpy floats included), ~3.5 KB/case.

🟡 **New:** `metrics(report)` with `evaluator=None` on a flattened multi-schema report silently unions the field paths — the exact misreport the docstring warns about, but nothing enforces it. Inline comment has a pre-tested one-hunk guard (50 tests still pass).

**On the getter (`poshinchen`):** I'd *not* put `metrics()` on `EvaluationReport` — it needs stickler, and the core report type should stay extra-free. Keep the static `StructuredOutputSimilarity.metrics(report, evaluator=...)` and make it refuse a mixed report (the guard). If you want an easier handle, the generic one is `EvaluationReport.for_evaluator(name) -> EvaluationReport` — a sub-report that composes with `.overall_score`, `.display()` and any evaluator's helpers; the CLI already regroups by that tag by hand (`run.py:230-239`). That's core API, so I'd make it a follow-up, not this PR.

Note: CI hasn't run on this head yet (needs workflow approval).

<details><summary>Evidence and appendix</summary>

### ✅ Verified (py3.13, stickler-eval 1.0.0)
- `pytest tests/.../test_structured_output.py` → 50 passed; `ruff check`/`ruff format --check` clean on changed files; `mypy` clean on `structured_output.py`, `types/evaluation.py`.
- Round-trip: `to_dict()` → `{'evaluator_type': 'StructuredOutputSimilarity', 'name': 'inv', 'match_threshold': 0.9, 'weight_hints': True, 'model_cls': 'models.Invoice'}`; `Experiment.from_file(p).evaluators[0]` has the same `model_cls`/threshold/hints/name and `explain()` compares `==`. `SOS(Outer.Inner).to_dict()['model_cls'] == 'models.Outer.Inner'` and reloads to the class.
- Reload of report: `SOS.metrics(r).field_metrics == SOS.metrics(EvaluationReport.model_validate_json(r.model_dump_json())).field_metrics` → True, also for a `list[Enum]` field whose `field_scores` are `np.float64`.
- Zero comparison rows (all N/A or all validation failures) → `ProcessEvaluation` with `document_count 0`, `field_metrics {}` — no crash.
- Harness already rejects two unnamed instances (`experiment.py:209` `Evaluator names must be unique`), so the unnamed-collision case is moot.
- `python -X importtime -c 'import strands_evals'` no longer lists `stickler`.

### Repro for the new finding (`getter.py`)
`Experiment(cases=[inv, rec], evaluators=[SOS(Invoice, name="inv"), SOS(Receipt, name="rec"), Equals()]).run_evaluations(task)` then `SOS.metrics(r)` → `document_count 2`, `field_metrics` keys `['invoice_id','line_items','merchant','tax','total_amount','vendor_name']`, `cm_recall` `line_items=0.0`, `total_amount=0.0` — fields that were never wrong read as missed because the Receipt rows lack them. No warning. With the guard: `ValueError: report carries comparisons from 2 evaluators (inv, rec); pass evaluator=<name>…`; `SOS.metrics(r, evaluator="inv")` → the four Invoice paths only.

`per_case(report)` is unaffected: it lists rows with their `model` label and aggregates nothing.

### Appendix — non-blocking
- ⚪ `_import_dotted_path` (structured_output.py:96-98) swallows only `ImportError`; a module that raises something else at import propagates the raw error rather than the directed `TypeError`. Fine, arguably better.
- ⚪ `per_case()` returns `dict(row.metadata["field_scores"])` which may contain `np.float64` in-process (JSON-clean after a round-trip). Harmless for printing/dataframes; mention if anyone `json.dumps` it directly.
- ⚪ `AGENTS.md` line about the `Evaluator` suffix meaning LLM-based, as `sromoam` suggested — cheap and would have saved the naming round.

### Process note
Two fresh-context specialist passes were dispatched for this delta and had not returned when this was posted; everything above was verified directly by me at `53f21fb`. Repro scripts and logs (`verify.py/.log`, `getter.py/.log`) are attached to the PR card.
</details>

src/strands_evals/evaluators/deterministic/structured_output.py
strandly-the-agent: Fixed at 53f21fb: `to_file`/`from_file` round-trip verified, nested `Outer.Inner` reloads, `explain()` identical.

src/strands_evals/evaluators/deterministic/structured_output.py
strandly-the-agent: Fixed at 53f21fb: `model_cls` required; `from_file` dataset and warm `LocalFileTaskResultStore` both score (0.9821/0.9821).

src/strands_evals/evaluators/deterministic/structured_output.py
strandly-the-agent: Fixed at 53f21fb: no accumulator; `metrics`/`per_case` are functions of the report, error rows are score-0/N/A rows.

README.md
strandly-the-agent: Fixed at 53f21fb: README now points at `StructuredModel` + `ComparableField(weight=...)` or `weight_hints=True`.

src/strands_evals/evaluators/deterministic/structured_output.py
```diff
@@ -0,0 +1,448 @@
...
+                EvaluationOutput(
+                    score=0.0,
+                    test_pass=True,
+                    reason="expected_output is None, so there was no ground truth to compare against",
+                    label=NOT_APPLICABLE,
+                )
+            ]
+
+        try:
+            expected = self._coerce(evaluation_case.expected_output, "expected_output")
```
strandly-the-agent, line 281: **A wrong-type `expected_output` is coerced to a blank model and scores 1.0 / pass on an all-optional schema.** `_coerce` runs `cls.model_validate(value.model_dump())` for a foreign `BaseModel` (`:427-429`), pydantic ignores the unknown keys, both sides come out blank, and `_passed` (`:159-160`) treats "nothing findable, nothing invented" as a match. Reached by the README's mixed-schema recipe, since every evaluator scores every case.

Pre-tested: 49/50 pass (`test_a_foreign_shape_does_not_score` asserts the old score-0 behaviour and should expect `NOT_APPLICABLE`), ruff clean; the mixed run goes to `overall 1.0`, `document_count 1`:

```suggestion
        if isinstance(evaluation_case.expected_output, BaseModel) and not isinstance(
            evaluation_case.expected_output, self._model_cls
        ):
            return [
                EvaluationOutput(
                    score=0.0,
                    test_pass=True,
                    reason=f"expected_output is a {type(evaluation_case.expected_output).__name__}, not a {name}",
                    label=NOT_APPLICABLE,
                )
            ]

        try:
            expected = self._coerce(evaluation_case.expected_output, "expected_output")
```

Dicts from JSON datasets can't be told apart this way, so README:566 should drop "any foreign shape scores 0" and say one `Experiment` per schema for file-loaded suites.

src/strands_evals/evaluators/deterministic/structured_output.py
```diff
@@ -0,0 +1,448 @@
...
+
+    def to_dict(self) -> dict:
+        """Convert the evaluator into a dictionary.
+
+        Returns:
+            dict: The evaluator's information, with `model_cls` as a dotted path so the
+            result is JSON-serializable and `from_dict` can resolve it back.
+        """
+        _dict = super().to_dict()
+        _dict["model_cls"] = f"{self._model_cls.__module__}.{self._model_cls.__qualname__}"
```
strandly-the-agent, line 254: **Emits paths that can never reload.** A model defined in the eval script serializes as `__main__.Invoice`, one defined inside a function as `pkg.fn.<locals>.Invoice`; both pass `to_file` and fail (or bind a different same-named class) at `from_file` in another process. Same situation `OutputEvaluator.to_dict` handles for tools with a save-time `logger.warning`.

Suggestion: `if cls.__module__ == "__main__" or "<locals>" in cls.__qualname__: logger.warning("model_cls=<%s> | not importable by dotted path; from_file() will not reload this evaluator", path)` — or raise, if you'd rather fail early.

strandly-the-agent (changes requested):
**Addendum — withdrawing the approve above on one point.** The two specialist passes I'd dispatched came back after I posted, and one finding holds up and is worse than anything in my review: **a case of the wrong type scores `1.0` and passes** when the target model's fields are all optional — the sparse schema this PR is written around — and the README's own mixed-schema recipe triggers it.

🔴 `_coerce` (`structured_output.py:427-429`, `:430-431`) validates a foreign `BaseModel`/dict into `model_cls`; pydantic's default `extra="ignore"` drops every key, so both sides become blank instances and `_passed` (`:159-160`, nothing findable + no `fa`) calls it a match. Repro: `[SOS(Sparse, name="invoice"), SOS(Person, name="person")]` over one Sparse case and one Person case → `invoice/person: score 1.0 pass True "all fields matched"`, `metrics(evaluator="invoice").document_count == 2` for a one-invoice suite. README:566 ("any foreign shape scores 0 with a reason rather than silently coercing") is false for exactly this shape. Since the harness runs every evaluator on every case, the README recipe cross-scores half the suite either way (`person/inv` scores 0 for a required-field model, dragging `overall` to 0.75).

Inline has a pre-tested fix: a foreign `expected_output` model becomes a `NOT_APPLICABLE` row. With it the same run reports `overall 1.0`, `document_count 1`, and the mixed-schema recipe actually works for in-memory cases (49/50 tests pass; the one failing test encodes the old score-0 behaviour and needs its assertion flipped to N/A). JSON-loaded dicts can't be told apart this way, so README should still say one `Experiment` per schema for file-based suites.

🟡 `to_dict()` (`:254`) still emits `__main__.X` / `pkg.fn.<locals>.X`, which only fails at `from_file` time in another process — the save-time warning from my 09-22 comment isn't in. Small.

<details><summary>Other pass output, and the getter question again</summary>

- The static `metrics` ignores the instance: `ev.metrics(report)` — the natural port of the old call — type-checks and silently aggregates every evaluator's rows. With the guard from my previous inline comment this raises instead of misleading, which is enough; if you'd rather it *work*, make `metrics`/`per_case` instance methods defaulting `evaluator=self.get_name()`, at the cost of needing an instance to read a saved report. Your call, `poshinchen` — both are fine, silent is not.
- `_import_dotted_path` (`:96-98`) catches the target module's *own* `ImportError` (e.g. a missing dep inside `myapp.models`) and reports "could not resolve", hiding the real `ModuleNotFoundError`. Nit: catch only when the prefix module itself is missing (`exc.name` check).
- `metrics(evaluator=)` docstring (`:343-345`) states the wrong mechanism: an absent path contributes nothing, not a false miss — what actually misleads is `document_count` and `tn`. Doc nit.
- Design question from the reviewer pass: `match_threshold` doubles as the recall bar, so with two populated fields only 100 % recall passes (0.5 < 0.7). Intended?
- Held: `aggregate_from_comparisons([])` fine on every zero-row shape; JSON round-trip numerically exact; `display()`, `_print_summary`, `strands-evals run --output` all survive `metadata` (~4.8 KB/case, ~76 % of the report — maintainer call).

Scripts: `/scratch/v2/adv/p1c.py`, `p8.py` (attached to the PR card as `v2-adv-*`).
</details>

sromoam:
@poshinchen this round's findings are fixed at `ddb7897`, and the two open design questions have answers below for your confirmation.

**Fixed:**

- **The wrong-type false pass.** A foreign `expected_output` was coerced through `model_validate(other.model_dump())`; Pydantic drops unknown keys, so an all-optional schema compared blank against blank and scored 1.0. It now returns a `NOT_APPLICABLE` row, and a foreign `actual_output` is a score-0 failure. Your reviewer's patch, plus the `actual_output` side.
- **`metrics(report)` on a report carrying two evaluators** now raises with a fix-it message instead of merging schemas. This also covers `ev.metrics(report)` — it can no longer mislead.
- **`to_dict()` warns at save time** for `__main__` and `<locals>` model paths that cannot reload, following the `OutputEvaluator` precedent.
- **`_import_dotted_path` re-raises a module's own `ImportError`** instead of reporting a missing dependency inside the target module as "could not resolve".
- **The README mixed-schema recipe is removed.** Guidance is now one `Experiment` per output type. The recipe produced 9 rows for 3 documents (6 of them cross-pair noise) and only worked in memory, since a serialized case is a `dict` with no class. An unbound mode (`model_cls` omitted, class inferred per case) would serve that layout properly and is noted in the PR as a follow-up — it can never work for file-loaded suites, so it needs its own design.

**On the reviewer's `match_threshold` question — no change, documented instead.** The premise was that recall punishes wrong values on small documents; measured, it does not: a wrong value is `fd`, which recall ignores, so recall only drops for *missing* values. The two gates are complements — score catches wrong values, recall catches omissions. One knob bounding both is now stated explicitly in the docstring and README, with the escape hatch (`metadata["recall"]`) for anyone wanting independent policies. A separate `recall_threshold=` stays additive if demand appears.

**On your getter question — recommending they stay static.** The main consumer of a rollup is a saved report read in another process (`report.json` from the CLI); an instance method would force importing the model class just to aggregate comparisons the report already carries. With one `Experiment` per schema as the documented layout, `metrics(report)` needs no argument in the common case, and the new raise means the one wrong call fails loudly. Happy to flip to instance methods defaulting `evaluator=self.get_name()` if you prefer the ergonomics.

`hatch run prepare` green: ruff and mypy clean, 2197 tests on 3.10–3.14 (57 in the evaluator's file). CI needs a workflow-run approval when you get a chance.

sromoam:
@poshinchen two more commits, from testing end to end ahead of your review, and one question for you.

**Tested:** a clean install of the wheel from GitHub; a real `strands.Agent` on Bedrock (`global.anthropic.claude-sonnet-4-6`); `strands-evals validate` and `run` with both `--task` and `--agent`; `strands-evals report` on the saved output; and `per_case` / `metrics` on that report in a separate process. The README example, run as written, reproduces its numbers.

**Fixed in `10ee56c`**, both found by that test:

- **Re-running a notebook cell scored every case 0.** Re-running a class definition creates a new class object, and the foreign-model check compared class identity, so every case was not-applicable. It now compares fields: a model of another class is scored when nothing would be lost converting it, and is not-applicable when it has fields `model_cls` lacks.
- **A task returning the `AgentResult` scored 0.** It is now unwrapped to its `structured_output`.

**`a339b32`** trims the evaluator's docstrings from 537 lines to 401. Docstrings and comments only: the syntax tree with docstrings removed is identical before and after.

**Question: the raw comparison's share of the report.** Each row's `metadata["comparison"]` is what `metrics(report)` aggregates. In a 4-case report it is 72% of the JSON, about 2.7 KB per case on a 4-field model, or roughly 27 MB for 10,000 cases. I measured whether `metrics()` could work from less: it cannot. The confusion matrix it needs is 96% of the comparison. So the choice is:

1. **Keep it as is.** `metrics()` works on any report, including one reloaded from disk.
2. **Add `retain_comparison: bool = True`.** Setting it to `False` stores only `field_scores`, `precision`, `recall` and `f1` (about 140 bytes per case), and `metrics()` raises on such a report.

I'd keep it as is. Option 2 is purely additive, so it can come later if someone reaches a size limit. But the report format is yours, so it's your call.

Separately, not in this PR's scope: if any task returns an `AgentResult`, `strands-evals run --output` cannot save the report (`Unable to serialize unknown type: strands.telemetry.metrics.Trace`). The built-in `Contains` evaluator fails the same way, so the harness is storing the raw result. Happy to file an issue if you'd like one.

PR #401, opened 2026-09-15
fix: improve root cause detector parsing returned json

ybdarrenwang:
## Description
The root cause analysis pipeline runs in two phases: a) Failure detection — a plain text-mode LLM call whose response we parse into FailureDetectionStructuredOutput. b) Root cause analysis — a Strands structured-output call (structured_output_model=RCAStructuredOutput).

Leaving (a) as a text-mode call was intentional. Per the _call_model docstring, putting the failure taxonomy in the system role makes the model treat it with higher authority and over-apply categories (more false positives), and text mode also avoids tool-use structured-output overhead that was found to degrade detection quality. So (a) keeps a free-text response and parses the JSON out of it.

However, we found that the model often wraps that JSON in prose or markdown, and the parser couldn't recover it:
```
Failed to parse LLM response for failure detection: 1 validation error for FailureDetectionStructuredOutput
  Invalid JSON: expected value at line 1 column 1 [type=json_invalid, input_value='Looking at this session,...["high"]\n    }\n  ]\n}', input_type=str]
```
On this failure the detector logs a warning and returns [], so RCA has nothing to analyze and the diagnosis silently comes back empty.

This PR adds tolerant extraction helpers that handle the three shapes the model emits in practice, by scanning for the first balanced {...}/[...] span that actually parses (skipping prose brackets like a mentioned regex). This keeps the intentional text-mode design while making detection resilient to wrapped output.

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

## Checklist

----


poshinchen:
@strandly-the-agent review the changes, it seems like it's the strands-agents structureOutput issue.

src/strands_evals/detectors/failure_detector.py
```diff
@@ -176,13 +177,110 @@ async def _stream() -> str:
...
+            candidate = span.strip()
+            try:
+                parsed = json.loads(candidate)
+            except json.JSONDecodeError:
+                continue
+            if first_parseable is None:
+                first_parseable = candidate
+            if isinstance(parsed, dict) and "errors" in parsed:
+                return candidate
+    return first_parseable
```
strandly-the-agent, line 281: **First `"errors"`-bearing span wins, so an echoed prompt example shadows the real answer — and if the echo validates, it's reported as a real failure.**

The prompt contains three `{"errors": [...]}` blocks (`failure_detection_v0.py:199`, `:223+`). A response that restates one unfenced before answering hits `return candidate` on the echo. Verified with the first `EXAMPLES` block verbatim + a real payload: main → `[]` + warning; this branch → `[('span_id_where_repetition_occurs', [0.9])]`, no warning, and RCA then analyses a span that doesn't exist. Full-schema validation wouldn't help — a well-formed example validates.

Suggestion: keep the *last* errors-bearing span (a preamble is the documented failure mode; a trailing recap isn't). 107 tests pass with this.

```suggestion
            if isinstance(parsed, dict) and "errors" in parsed:
                payload = candidate
    return payload or first_parseable
```

(plus `payload: str | None = None` next to `first_parseable` on line 269.)

tests/strands_evals/detectors/test_failure_detector.py
```diff
@@ -103,6 +103,45 @@ def test_extract_json_with_surrounding_text():
...
+
+    The detector schema is well-known, so the span that validates against
+    it wins over an earlier span that merely parses as JSON. Otherwise the
+    decoy fails schema validation downstream and reintroduces the
+    silent-empty diagnosis this logic exists to prevent.
+    """
+    real_payload = (
+        '{"errors": [{"location": "s1", "category": ["err"], '
+        '"confidence": ["high"], "evidence": ["ev"]}]}'
+    )
```
strandly-the-agent, line 123: `ruff format --check` (0.16.8) rewrites this — fits in 120 cols. Won't fail CI (Lint runs the linter only), but `hatch run prepare` would have applied it.

```suggestion
    real_payload = '{"errors": [{"location": "s1", "category": ["err"], "confidence": ["high"], "evidence": ["ev"]}]}'
```

strandly-the-agent (changes requested):
**Changes requested — one regression, two-line fix.** The extraction fixes the reported shape (verified: prose + pretty-printed JSON goes `[]` → 1 item vs main). But the new `"errors"`-key preference returns the *first* match, so a model that restates one of the prompt's own `EXAMPLES` blocks before answering now yields a **fabricated failure at 0.9 confidence with no warning** — main returned `[]` + a warning. Inline suggestion below; 107 tests still pass with it.

On the structured_output question: agree with the earlier answer — not a strands bug. In strands 1.56.0 `structured_output_model` is forced tool use (`event_loop.py:369-378` sets `tool_choice` on `end_turn`) and the result is `toolUse.input`, so it can't produce prose-wrapped text. This path (`_call_model`, `failure_detector.py:151-176`) is raw `model.stream()` by design and never enters that machinery — the prose wrapping is plain text-mode model behaviour.

<details>
<summary>✅ Verified (head 692ee13)</summary>

- `pytest tests/strands_evals/detectors -q` → 107 passed; `ruff check` clean; `mypy` clean on `failure_detector.py`.
- `ruff format --check` (0.16.8) flags `test_failure_detector.py:120-123`. CI's Lint job runs `hatch fmt --linter --check` only, so it stays green — but `hatch run prepare` would have rewritten it; suggestion inline.
- Repros on pr-401 vs main: prose+pretty JSON `0→1`, prose+JSON+trailer `0→1`, fenced/clean unchanged, no-JSON-at-all still `[]` + warning. Echoed-example case: main `[]`, pr-401 `[('span_id_where_repetition_occurs', [0.9])]`, pr-401+fix `[('real-span', [0.5])]`.
- Held under attack: braces/quotes inside JSON strings, stray `"` in prose before the payload, unterminated fence (falls through correctly), wrapper objects `{"result": {"errors": …}}`, 180 KB of prose → 0.05 s, 3000 unbalanced `{` → 0.65 s.
- Adversarial pass timed out before reporting; the above is from the reviewer pass plus my own runs, so coverage of exotic inputs is partial.

</details>

<details>
<summary>Questions (non-blocking)</summary>

- Is the `[...]` scan at `failure_detector.py:270` ever useful? `_iter_balanced_spans` already yields nested objects inside an array, and `model_validate_json` rejects a top-level list anyway — `_extract_json('The failures are: [{"location": "s1"}]')` returns the inner object, never the array. Dropping it removes a loop level and makes `test_extract_json_skips_regex_in_prose` vacuous (`[a-z0-9-]` is never a candidate).
- Title says "root cause detector" but the only caller of `_extract_json` is `failure_detector.py:291`; RCA uses `structured_output_model` and never parses text. `fix(detectors): recover prose-wrapped JSON in failure detection` would match the change.

</details>

<details>
<summary>Appendix — non-blocking (4)</summary>

- ⚪ Docstrings on `_extract_json` / `_extract_balanced_json` are ~20 lines each on private helpers; median private-fn docstring in `detectors/` is 4 lines. Most of it narrates the code — the one non-obvious thing (why a decoy-resistant discriminator) fits in 2–3 lines.
- ⚪ `test_extract_json_prefers_schema_over_decoy` (`:107`) says "the span that validates against it wins" but the code only checks key presence — that's exactly the gap above. Rename to `_prefers_errors_key_over_decoy`.
- ⚪ `test_extract_json_bare_object_after_prose` (`:106`) is compact single-line JSON; the reported shape is pretty-printed. `json.dumps(..., indent=2)` + assert through `_parse_text_result` makes the test *be* the bug report. `test_extract_json_falls_back_to_first_parseable` (`:139`) pins an internal choice with no observable difference (`{"foo": 1}` and raw text both end in `[]` + warning).
- ⚪ Pre-existing: the fence regex at `:200` takes the *first* fenced block, so a fenced echoed example followed by the fenced real answer returns the echo (`{"errors": []}` validates → silent `[]`). Unchanged by this PR; noting only because the same "last wins" idea would cover it if you touch that line.

</details>

poshinchen:
@strandly-the-agent review the updated changes

tests/strands_evals/detectors/test_failure_detector.py
```diff
@@ -103,6 +103,42 @@ def test_extract_json_with_surrounding_text():
...
+def test_extract_json_skips_regex_in_prose():
+    """Bracketed prose that isn't JSON (e.g. a regex) is skipped, not returned."""
+    text = 'The ids match [a-z0-9-]. Here is the result: {"errors": []}'
+    assert _extract_json(text) == '{"errors": []}'
+
+
+def test_extract_json_falls_back_to_first_parseable():
+    """When nothing validates against the schema, the first parseable span is returned."""
+    text = 'Here is some data: {"foo": 1} and more.'
+    assert _extract_json(text) == '{"foo": 1}'
```
strandly-the-agent, line 139: Non-blocking: nothing pins the case that motivated last-wins. This passes on b31b1c2 and fails on 692ee13 (returns `example-span`).

```suggestion
    assert _extract_json(text) == '{"foo": 1}'


def test_extract_json_echoed_example_does_not_shadow_real_payload():
    """A restated prompt example (which itself validates) must not be reported as the result."""
    example = '{"errors": [{"location": "example-span", "category": ["c"], "evidence": ["e"], "confidence": ["high"]}]}'
    real = '{"errors": [{"location": "real-span", "category": ["c"], "evidence": ["e"], "confidence": ["low"]}]}'
    result = _parse_text_result(f"Following the example format:\n{example}\nMy annotation:\n{real}")
    assert [f.span_id for f in result] == ["real-span"]
```

strandly-the-agent (approved):
**Approve.** Both previous findings are fixed at b31b1c2: last-wins selection (`payload` now initialised, so the `UnboundLocalError` from the intermediate commit is gone) and the formatting nit. The echoed-example case that previously produced a fabricated 0.9-confidence failure now returns the real payload; the `{"foo": 1}` fallback and pure-prose inputs return `[]` + warning without raising.

One non-blocking ask inline: there's no test pinning the echoed-example behaviour — the one below passes on this head and fails on 692ee13, so it's a real regression guard.

<details>
<summary>✅ Verified (b31b1c2) / attacked the fix</summary>

- Delta since 692ee13 on PR files is exactly the two accepted suggestions + `payload: str | None = None`; the rest is the `main` merge.
- `pytest tests/strands_evals/detectors -q` → 107 passed; `ruff check` clean; `ruff format --check` → 19 files already formatted; `mypy` clean on `failure_detector.py`.
- Repros: echoed `EXAMPLES` block + real → `[('real-span', [0.5])]`; echoed `OUTPUT_FORMAT` + real → same; `{"foo": 1}` fallback → `[]` + warning (no raise); pure prose → `[]` + warning; prose + pretty JSON → 1 item.
- Known limit of last-wins (I flagged it when suggesting it): a *trailing* `{"errors": []}` recap after the real payload validates and yields a silent `[]`. A leading preamble is the documented failure mode and a trailing recap isn't, so I'd leave it; noting so nobody rediscovers it as a surprise.

</details>

poshinchen (approved):

